PR #4, opened 2026-04-09
[Sprint 3] MCP Server

uzyn:
## Summary
- Add MCP server (`aimx mcp`) using `rmcp` crate with stdio transport for AI agent email access
- Implement all 9 MCP tools: `mailbox_create`, `mailbox_list`, `mailbox_delete`, `email_list` (with unread/from/since/subject filters), `email_read`, `email_mark_read`, `email_mark_unread`, `email_send` (with DKIM signing), `email_reply` (with In-Reply-To/References threading)
- Add 21 unit tests for email listing, filtering, frontmatter parsing, read/unread status, and tool registration
- Add 11 MCP integration tests that spawn `aimx mcp` as a child process and communicate via JSON-RPC over stdio

### S3.1 — MCP Transport + Mailbox Tools
- [x] `aimx mcp` starts an MCP server in stdio mode
- [x] Server responds to MCP `initialize` handshake correctly
- [x] `mailbox_create(name)` creates mailbox and returns confirmation
- [x] `mailbox_list()` returns all mailboxes with message counts (total and unread)
- [x] `mailbox_delete(name)` deletes mailbox (with appropriate safeguards)
- [x] Server exits cleanly when stdin closes
- [x] Integration tests: spawn `aimx mcp`, send JSON-RPC requests, assert responses

### S3.2 — Email Read + List Tools
- [x] `email_list(mailbox)` returns frontmatter of all emails in the mailbox
- [x] `email_list` supports optional filters: `unread`, `from`, `since`, `subject`
- [x] `email_read(mailbox, id)` returns full Markdown content of the email
- [x] `email_mark_read(mailbox, id)` updates frontmatter `read: true`
- [x] `email_mark_unread(mailbox, id)` updates frontmatter `read: false`
- [x] Non-existent mailbox or email ID returns clear MCP error
- [x] Unit tests for email listing with each filter type and combinations
- [x] Unit tests for mark read/unread
- [x] Integration tests via MCP JSON-RPC

### S3.3 — Email Send + Reply Tools
- [x] `email_send(from_mailbox, to, subject, body, attachments?)` composes, DKIM-signs, and sends
- [x] `email_reply(mailbox, id, body)` replies with correct In-Reply-To/References headers
- [x] Send/reply return confirmation with the sent Message-ID
- [x] Errors (missing mailbox, invalid recipient, missing DKIM key) return clear MCP errors
[...]

PR #3, opened 2026-04-09
[Sprint 2.5] Non-blocking Cleanup

uzyn:
## Summary

- **S2.5-1 Ingest Hardening:** Added `--data-dir` / `AIMX_DATA_DIR` CLI option to override default data path. Added mailbox name validation to reject `..`, `/`, `\`, empty, and null byte names. Replaced hand-rolled `yaml_escape` with `serde_yaml` struct serialization for frontmatter generation, eliminating edge cases. Added full-pipeline integration tests exercising `ingest_email()` with all fixture `.eml` files via CLI subprocess.
- **S2.5-2 Send Hardening:** Added `escape_filename()` to sanitize attachment filenames in MIME headers (quotes, newlines, carriage returns). Refactored duplicated header construction into `write_common_headers()`. Added end-to-end integration tests for `aimx dkim-keygen` CLI. Added test verifying custom `dkim_selector` config value is used in DKIM signing.

## Changes

| File | Change |
|------|--------|
| `Cargo.toml` | Added clap `env` feature for `AIMX_DATA_DIR` env var support |
| `src/cli.rs` | Added `--data-dir` global option with `AIMX_DATA_DIR` env |
| `src/main.rs` | Thread `data_dir` option through to all subcommand handlers |
| `src/ingest.rs` | Replace `yaml_escape` + manual frontmatter with `serde_yaml` struct serialization; accept `data_dir` override |
| `src/mailbox.rs` | Add `validate_mailbox_name()` with path traversal checks; accept `data_dir` override |
| `src/send.rs` | Add `escape_filename()`, extract `write_common_headers()`, accept `data_dir` override |
| `tests/integration.rs` | 12 tests total: full-pipeline ingest for all 4 fixtures, routing, env var, dkim-keygen e2e |

## Test results

- **64 unit tests** passing (8 new)
- **12 integration tests** passing (7 new)
- `cargo clippy -- -D warnings` clean
- `cargo fmt -- --check` clean

## Test plan

- [x] All existing tests pass (no regressions)
- [x] `--data-dir` flag overrides default data directory
- [x] `AIMX_DATA_DIR` env var overrides default data directory
- [x] Mailbox name validation rejects `..`, `/`, `\`, empty, null bytes
- [x] Frontmatter generated by serde_yaml is valid YAML with correct field values
- [x] Special characters in email fields (newlines, colons, hashes, @) are handled correctly by serde_yaml
- [x] Attachment filenames with quotes/newlines/CR are escaped in MIME headers
[...]

PR #2, opened 2026-04-09
[Sprint 2] DKIM + Production-Quality Outbound

uzyn:
## Summary

- **S2.1 DKIM Key Generation**: `aimx dkim-keygen` generates 2048-bit RSA keypair, stores in `<data_dir>/dkim/`, outputs DNS TXT record for DKIM public key. Existing keys protected with `--force` flag.
- **S2.2 DKIM Signing on Outbound**: All outbound email signed with DKIM-Signature header using RSA-SHA256 via `mail-auth` crate. Signs From, To, Subject, Date, Message-ID, In-Reply-To, References headers. Configurable selector (default: `dkim`). Missing key produces clear warning, does not crash.
- **S2.3 Email Threading**: `--reply-to <message-id>` flag on `aimx send` sets In-Reply-To and References headers. Bare message IDs auto-wrapped in angle brackets. `build_references()` utility for constructing References chains (used by Sprint 3 MCP reply).
- **S2.4 File Attachments on Send**: `--attachment /path/to/file` flag (repeatable) creates multipart/mixed MIME with base64-encoded attachments. MIME type inferred from file extension via `mime_guess`. Missing files produce clear errors.

## Changes

- New `src/dkim.rs` module: key generation, DNS record formatting, private key loading, DKIM signing
- Updated `src/cli.rs`: added `DkimKeygen` command, `--reply-to` and `--attachment` flags to `Send`
- Updated `src/send.rs`: MIME multipart composition, threading headers, DKIM signing integration, attachment handling
- Updated `src/config.rs`: added `dkim_selector` field with `"dkim"` default
- Updated `src/main.rs`: wired `DkimKeygen` command
- Dependencies: `rsa`, `base64`, `mail-auth`, `mime_guess`, `rand`

## Test plan

- [x] 60 tests total (55 unit + 5 integration), up from 37 in Sprint 1
- [x] DKIM: keypair generation (valid 2048-bit), no-overwrite guard, force overwrite, DNS record format, load/missing key, sign-and-verify roundtrip
- [x] Threading: In-Reply-To set, References set, bare ID normalization, omitted when no reply-to
- [x] Attachments: single file, multiple files, MIME type inference (pdf/png/txt), missing file error, combined with reply-to
- [x] DKIM signing integration: send via mock transport verifies DKIM-Signature header present
- [x] All existing Sprint 1 tests still pass
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt -- --check` clean

PR #1, opened 2026-04-09
[Sprint 1] Core Pipeline + Idea Validation

uzyn:
## Summary

- **S1.1 — Project Scaffolding + CI**: Cargo project with `clap` CLI showing all planned subcommands, `config.yaml` schema with serde, GitHub Actions CI (test + clippy + fmt), test fixtures for `.eml` files
- **S1.2 — Email Ingest Pipeline**: `aimx ingest <rcpt>` reads raw EML from stdin, parses with `mail-parser`, produces Markdown with YAML frontmatter (all required fields, `read: false`), routes to correct mailbox or catchall, names files `YYYY-MM-DD-NNN.md`
- **S1.3 — Attachment Extraction**: Extracts attachments to `<mailbox>/attachments/`, deduplicates filenames, lists attachments in frontmatter with filename/content_type/size/path
- **S1.4 — Basic Email Sending**: `aimx send --from --to --subject --body` composes RFC 5322 message and hands to sendmail, with trait-abstracted transport for testing
- **S1.5 — Mailbox Management**: `aimx mailbox create/list/delete` with directory creation, config persistence, message counts, confirmation prompt, catchall protection

## Test plan

- [x] 30 unit tests covering config parsing, ingest (plain/HTML/multipart/routing/frontmatter), attachments (single/multiple/duplicate/none), send (composition/headers/mock transport), mailbox (create/list/delete/errors)
- [x] 5 integration tests covering CLI help output and fixture file parsing
- [x] `cargo clippy -- -D warnings` passes clean
- [x] `cargo fmt -- --check` passes clean
- [x] CI workflow configured for push/PR on main and sprint branches

PR #9, opened 2026-04-10
[Sprint 7] Security Hardening + Critical Fixes

uzyn:
## Summary
- **S7.1 - DKIM Enforcement**: `send::run()` and MCP `email_send`/`email_reply` now error when DKIM key is missing instead of silently sending unsigned email
- **S7.2 - CRLF Header Injection**: Added `sanitize_header_value()` to strip `\r\n` from From/To/Subject headers, preventing header injection attacks
- **S7.3 - Atomic File ID**: Replaced TOCTOU-vulnerable `generate_file_id()` + `fs::write()` with `create_file_atomic()` using `O_CREAT|O_EXCL`
- **S7.4 - Verify Race Condition**: Moved "before" file snapshot before `send::run()` so fast replies aren't missed; added missing catchall dir error
- **S7.5 - Verify in Setup**: Added `VerifyRunner` trait and interactive prompt after DNS verification to optionally run end-to-end email verification

## Test plan
- [x] 235 unit tests pass (15 new tests added)
- [x] 35 integration tests pass (no regressions)
- [x] S7.1: `run_with_missing_dkim_key_returns_error`, `run_with_missing_config_returns_error`, `load_dkim_key_missing_returns_error`, `load_dkim_key_present_returns_ok`
- [x] S7.2: `subject_crlf_injection_produces_no_extra_headers`, `from_crlf_injection_produces_no_extra_headers`, `to_crlf_injection_produces_no_extra_headers`, `bare_newline_injection_produces_no_extra_headers`, `normal_headers_pass_unchanged`
- [x] S7.3: `atomic_file_creation_retries_on_collision`, `atomic_file_creation_two_rapid_calls_produce_different_ids`
- [x] S7.4: `run_errors_on_missing_catchall_dir`
- [x] S7.5: `verify_runner_trait_mock_pass`, `verify_runner_trait_mock_fail`, `setup_verify_step_available`

### DIFF
--- a/docs/sprint.md
+++ b/docs/sprint.md
@@ -631,7 +631,7 @@ aimx verify
 
 ---
 
-## Sprint 7 — Security Hardening + Critical Fixes (Days 16–18.5) [NOT STARTED]
+## Sprint 7 — Security Hardening + Critical Fixes (Days 16–18.5) [IN PROGRESS]
 
 **Goal:** Fix all critical and high-severity issues found in the post-v1 code review audit. These address security vulnerabilities, data loss risks, and PRD compliance gaps.
 
@@ -951,7 +951,7 @@ One of these is wrong. Research suggests DigitalOcean's current policy is closer
 | 5 | 10.5–12.5 | Setup Wizard | `aimx setup` — one-command setup with preflight + DNS | Done |
 | 5.5 | 12.5–13 | Non-blocking Cleanup | Serialization, resolver dedup, SPF fix, setup backup | Done |
 | 6 | 13–15.5 | Verify Service + Polish | Hosted probe, status/verify CLI, README | Done |
-| 7 | 16–18.5 | Security Hardening + Critical Fixes | DKIM enforcement, header injection fix, atomic ingest, verify race fix, setup e2e verify | Not Started |
+| 7 | 16–18.5 | Security Hardening + Critical Fixes | DKIM enforcement, header injection fix, atomic ingest, verify race fix, setup e2e verify | In Progress |
 | 8 | 19–21.5 | Setup Robustness, CI & Documentation | DNS verification accuracy, data-dir propagation, SPF fix, configurable verify URLs, CI coverage, doc fixes | Not Started |
 
 ## Deferred to v2
--- a/src/ingest.rs
+++ b/src/ingest.rs
@@ -116,14 +116,10 @@ pub fn ingest_email(
 
     let attachments = extract_attachments(&message, &mailbox_dir)?;
 
-    let id = generate_file_id(&mailbox_dir);
-    let filename = format!("{id}.md");
-    let filepath = mailbox_dir.join(&filename);
-
     let (dkim_result, spf_result) = verify_auth(raw, rcpt);
 
-    let meta = EmailMetadata {
-        id: id.clone(),
+    let meta_template = EmailMetadata {
+        id: String::new(), // placeholder, set after atomic creation
         message_id,
         from,
         to,
@@ -138,8 +134,7 @@ pub fn ingest_email(
         spf: spf_result,
     };
 
-    let content = format_markdown(&meta, &body);
-    std::fs::write(&filepath, content)?;
+    let (_id, filepath, meta) = create_file_atomic(&mailbox_dir, &meta_template, &body)?;
 
     if let Some(mailbox_config) = config.mailboxes.get(&mailbox) {
         let ctx = TriggerContext {
@@ -404,17 +399,35 @@ fn deduplicate_filename(dir: &Path, filename: &str) -> String {
     unreachable!()
 }
 
-fn generate_file_id(mailbox_dir: &Path) -> String {
+fn create_file_atomic(
+    mailbox_dir: &Path,
+    meta_template: &EmailMetadata,
+    body: &str,
+) -> Result<(String, std::path::PathBuf, EmailMetadata), Box<dyn std::error::Error>> {
+    use std::fs::OpenOptions;
+    use std::io::Write;
+
     let today = chrono::Utc::now().format("%Y-%m-%d").to_string();
     let mut counter = 1u32;
 
     loop {
         let candidate = format!("{today}-{counter:03}");
         let path = mailbox_dir.join(format!("{candidate}.md"));
-        if !path.exists() {
-            return candidate;
+
+        match OpenOptions::new().write(true).create_new(true).open(&path) {
+            Ok(mut file) => {
+                let mut meta = meta_template.clone();
+                meta.id = candidate.clone();
+                let content = format_markdown(&meta, body);
+                file.write_all(content.as_bytes())?;
+                return Ok((candidate, path, meta));
+            }
+            Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
+                counter += 1;
+                continue;
+            }
+            Err(e) => return Err(e.into()),
         }
-        counter += 1;
     }
 }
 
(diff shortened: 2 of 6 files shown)


uzyn (commented):
# Sprint 7 Code Review: Security Hardening + Critical Fixes

## Sprint Goal Assessment

Sprint 7 aims to fix all critical and high-severity issues from the post-v1 code review audit, covering security vulnerabilities (DKIM enforcement, header injection), data loss risks (atomic file IDs), race conditions (verify timing), and PRD compliance (verify in setup). All five stories are implemented and the overall approach is solid.

### S7.1 -- DKIM Enforcement

| Criterion | Status |
|-----------|--------|
| `send::run()` returns error when DKIM config/key missing | MET |
| MCP `email_send` returns clear MCP error when DKIM key unavailable | MET |
| MCP `email_reply` returns clear MCP error when DKIM key unavailable | MET |
| Unit test: `send::run()` with missing DKIM key returns error | MET |
| Unit test: MCP send/reply with missing DKIM key returns error response | MET (load_dkim_key_missing_returns_error covers the shared helper; no separate MCP-level send/reply integration test, but acceptable since both paths call the same `load_dkim_key`) |

### S7.2 -- CRLF Header Injection

| Criterion | Status |
|-----------|--------|
| From/To/Subject sanitized to strip CRLF | MET |
| Sanitization covers both `\r\n` and bare `\n` | MET |
| `compose_message()` returns error if header contains CRLF after sanitization (defense in depth) | NOT MET -- see Bug #1 below |
| Unit test: subject with `\r\n` no injected headers | MET |
| Unit test: from/to with `\r\n` no injected headers | MET |
| Unit test: normal headers pass unchanged | MET |

### S7.3 -- Atomic File ID Generation

| Criterion | Status |
|-----------|--------|
| File creation uses `OpenOptions::create_new(true)` | MET |
| On collision, ID counter increments and retries | MET |
| Unit test: two rapid calls produce different IDs | MET |
| Unit test: pre-existing file triggers retry, not overwrite | MET |

### S7.4 -- Verify Race Condition

| Criterion | Status |
|-----------|--------|
| "Before" snapshot taken before sending test email | MET |
| Existing unit tests still pass | MET |
| Missing catchall directory produces clear error | MET |

### S7.5 -- Verify in Setup

| Criterion | Status |
|-----------|--------|
| After DNS passes, setup offers end-to-end verify | MET |
| Verify failure is non-blocking | MET |
| Verify success reports full success | MET |
| User can decline verify | MET |
| Unit test: setup flow includes verify step (mockable) | PARTIAL -- mock trait exists and is tested, but no test exercises `run_setup_with_verify` with the mock (stdin blocks). The `setup_verify_step_available` test only calls `runner.run_verify(None)` directly, which is identical to `verify_runner_trait_mock_pass`. |

## Bugs Found

### Bug #1 (Non-blocker): `validate_header_value` is dead code after sanitization

In `src/send.rs` lines 77-83, `write_common_headers` first calls `sanitize_header_value()` which strips all `\r` and `\n` characters, then calls `validate_header_value()` on the already-sanitized values. Since sanitization removes all CR/LF, the validation check can **never** trigger -- it is dead code.

The S7.2 AC explicitly states: "compose_message() returns an error if a header value contains CRLF after sanitization (defense in depth)". The current implementation satisfies the "defense" idea (injection is prevented by sanitization), but the "in depth" layer (validation returning an error) is unreachable.

**Recommendation:** Either (a) validate first, then sanitize as a fallback, or (b) remove `validate_header_value` since it creates a false sense of defense-in-depth. Option (a) would be:
```rust
validate_header_value("From", &args.from)?;
let from = sanitize_header_value(&args.from);
```
This way, direct callers get an error, and the sanitization still protects as a safety net. Alternatively, just remove the dead validation since sanitization alone is sufficient.

### Bug #2 (Non-blocker): In-Reply-To and References headers are not sanitized

In `src/send.rs` lines 91-98, the `In-Reply-To` and `References` headers are interpolated without CRLF sanitization:
```rust
msg.push_str(&format!("In-Reply-To: {reply_id}\r\n"));
...
msg.push_str(&format!("References: {refs}\r\n"));
```

These values come from `args.reply_to` and `args.references`, which in the MCP `email_reply` path originate from stored email frontmatter (`meta.message_id`, `meta.references`). A malicious inbound email could contain a `Message-ID` header with CRLF sequences, which would be stored verbatim in frontmatter and then injected into outbound headers when an agent replies.

**Recommendation:** Apply `sanitize_header_value()` to `reply_id` and `refs` before interpolation.

## Test Coverage Gaps

1. **S7.5 integration test gap:** The `setup_verify_step_available` test is essentially a duplicate of `verify_runner_trait_mock_pass` -- it calls the mock directly rather than testing the verify step within `run_setup_with_verify`. The comment acknowledges this ("We can't fully test run_setup_with_verify here because it reads stdin"). Consider providing a stdin mock or extracting the verify prompt logic into a testable function.

2. **No counter overflow guard in `create_file_atomic`:** The `counter` (`u32`) increments without an upper bound. While extremely unlikely in practice (would need ~4 billion files in one day), a safety check like `if counter > 999_999 { return Err(...) }` would be prudent.

3. **No test for CRLF in reply_to/references:** Given the injection risk in Bug #2, tests for CRLF in `reply_to` and `references` fields would be valuable.

## Security Issues

1. **CRLF injection in In-Reply-To/References** (Bug #2 above) -- Medium severity. A crafted inbound email could cause header injection in outbound replies.

2. **`send_with_transport` still accepts `Option<dkim_key>`** -- Low severity. The function signature still allows `None` for the DKIM key, meaning future callers could accidentally bypass DKIM. All current callers enforce it upstream, so this is a design note rather than a bug.

## Alignment with PRD

- **FR-18 (DKIM signing):** Now properly enforced -- unsigned sends are rejected. Aligned.
- **FR-8 (End-to-end verify in setup):** Now integrated as an optional step. Aligned.
- **NFR-6 (Secure defaults):** DKIM enforcement improves this. CRLF sanitization is a good addition. The In-Reply-To/References gap slightly weakens it.

## Summary

The sprint delivers all five stories with solid implementations. The atomic file creation (S7.3), verify race fix (S7.4), and DKIM enforcement (S7.1) are clean and correct. The CRLF sanitization (S7.2) effectively prevents the primary attack vector but has two gaps: dead validation code and unsanitized In-Reply-To/References headers.

**Verdict: Needs minor fixes**

The two bugs are non-blocking for the security goals (the primary injection vectors in From/To/Subject are properly handled), but Bug #2 (unsanitized In-Reply-To/References) is a genuine injection vector that should be addressed before merge given this is a security hardening sprint. Bug #1 (dead validation) is a code quality issue.

uzyn:
## Review Feedback — Changes and Responses

### Fixed

1. **Bug #1: Dead `validate_header_value` code** — Reordered to validate *before* sanitize in `write_common_headers`. `compose_message()` now returns an error when From/To/Subject contain CRLF characters (satisfying the S7.2 AC for defense-in-depth). Sanitization remains as a safety net after validation. Updated all four CRLF injection tests to assert error returns instead of silent sanitization.

2. **Bug #2: Unsanitized In-Reply-To/References headers** — Applied `sanitize_header_value()` to both `reply_to` and `references` values before interpolation into headers. Added two new tests (`reply_to_crlf_injection_sanitized`, `references_crlf_injection_sanitized`) that verify CRLF in these fields does not produce injected header lines.

3. **Suggestion #3: S7.5 test duplication** — Replaced `setup_verify_step_available` with a compile-time function pointer check that verifies `run_setup_with_verify` accepts a `VerifyRunner` trait object. This is now distinct from `verify_runner_trait_mock_pass` (which tests the mock behavior).

4. **Suggestion #4: No upper-bound guard on counter** — Added a `MAX_COUNTER` of 999,999 to `create_file_atomic`. If exhausted, returns a descriptive error instead of silently wrapping.

### Intentionally Left As-Is

5. **Suggestion #5: `send_with_transport` still accepts `Option` for DKIM key** — This is a deliberate design choice. The function is the low-level transport layer; DKIM enforcement is the caller's responsibility (and `run()` enforces it). Keeping `Option` allows the function to be used in tests without DKIM setup and maintains separation of concerns. All production code paths already enforce DKIM via `run()` or the MCP handlers.

### Test Results

All 237 unit tests and 35 integration tests pass. Formatter is clean.

uzyn (commented):
## Re-review: Sprint 7 — Security Hardening + Critical Fixes

All five previously flagged issues have been satisfactorily addressed. **APPROVED** (posted as comment since GitHub prevents self-approval).

### Issue-by-Issue Verification

1. **Bug #1 (Dead `validate_header_value` code)** — RESOLVED. Validation now runs *before* sanitization in `write_common_headers` (lines 77-79 validate raw values, lines 81-83 sanitize as safety net). All four CRLF injection tests now assert error returns instead of silent sanitization. The defense-in-depth layering is correct: validate → error on bad input, sanitize → safety net.

2. **Bug #2 (Unsanitized In-Reply-To/References)** — RESOLVED. `sanitize_header_value()` is applied to both `reply_to` (line 92) and `references` (line 95) before interpolation into headers. Two new tests (`reply_to_crlf_injection_sanitized`, `references_crlf_injection_sanitized`) confirm no header injection occurs through these fields.

3. **Suggestion #3 (S7.5 test duplication)** — RESOLVED. `setup_verify_step_available` is now a compile-time function pointer check that verifies `run_setup_with_verify` accepts the correct signature including `&dyn VerifyRunner`. This is clearly distinct from `verify_runner_trait_mock_pass` which tests mock behavior.

4. **Suggestion #4 (No upper-bound guard on counter)** — RESOLVED. `MAX_COUNTER` of 999,999 added to `create_file_atomic` with a descriptive error message when exhausted.

5. **Suggestion #5 (`send_with_transport` accepts Option for DKIM key)** — ACCEPTED AS-IS. The implementer's reasoning is sound: `send_with_transport` is the low-level transport layer, and DKIM enforcement is the caller's responsibility. Both `run()` and the MCP handlers now properly enforce DKIM (returning errors on missing keys), so all production paths are covered. Keeping `Option` at the transport layer maintains separation of concerns and enables clean testing.

```
[Sprint 7] Security hardening and critical fixes

- S7.1: Enforce DKIM signing on all outbound email paths (send::run,
  MCP email_send, MCP email_reply) — missing DKIM key now returns an
  error instead of silently sending unsigned messages
- S7.2: Prevent CRLF header injection in From/To/Subject (validate
  then sanitize) and In-Reply-To/References (sanitize)
- S7.3: Atomic file ID generation using OpenOptions::create_new to
  eliminate TOCTOU race in concurrent mail delivery
- S7.4: Fix verify race condition by snapshotting mailbox before
  sending test email; add missing-catchall-dir guard
- S7.5: Integrate optional end-to-end verify into setup wizard via
  VerifyRunner trait
```

LGTM — approving.

PR #19, opened 2026-04-11
Rename probe_url to verify_host, add --verify-host flag

uzyn:
## Summary
- Rename config key `probe_url` → `verify_host`. The default is now a base URL (`https://check.aimx.email`) with no `/probe` suffix — aimx appends `/probe` internally when making the HTTP check.
- Add `--verify-host <URL>` flag to `aimx verify`, `aimx setup`, and `aimx preflight`, with precedence **CLI flag > config > default**.
- No legacy `probe_url` support: the project is pre-launch, so the old key is dropped entirely (any stale `probe_url` in a user config is silently ignored by serde and falls back to the default).

## Motivation
`probe_url` leaked an implementation detail (the `/probe` endpoint path) and forced users overriding it to know the exact route. `verify_host` describes *where* the verify service lives, not which endpoint.

## Test plan
- [x] `cargo fmt` clean
- [x] `cargo test` — 273 unit + 35 integration tests pass, including four new precedence tests for `resolve_verify_host` and a new trailing-slash-strip test (`real_network_ops_from_verify_host_strips_trailing_slash`)
- [x] `aimx verify --help`, `aimx setup --help`, `aimx preflight --help` all show the new `--verify-host` flag
- [x] Smoke: `aimx verify --verify-host https://localhost` produces a different outbound-check error than the default (proves the flag flows through to `RealNetworkOps`)
- [x] Smoke: trailing-slash input (`https://check.aimx.email/`) produces identical behavior to the no-slash form

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -30,7 +30,7 @@ The `--data-dir` flag takes precedence over the environment variable.
 | `domain` | string | *(required)* | The email domain (e.g. `agent.yourdomain.com`) |
 | `data_dir` | string | `/var/lib/aimx` | Directory for storing config, keys, and mailboxes |
 | `dkim_selector` | string | `dkim` | DKIM selector name used in DNS records |
-| `probe_url` | string | *(unset)* | Custom verify service URL (defaults to `check.aimx.email`) |
+| `verify_host` | string | `https://check.aimx.email` | Base URL of the verify service used by `aimx verify`, `setup`, and `preflight`. Can be overridden per-invocation with the `--verify-host` flag. |
 
 ### Mailbox settings
 
@@ -87,8 +87,8 @@ data_dir = "/var/lib/aimx"
 # DKIM selector name (default: dkim)
 dkim_selector = "dkim"
 
-# Custom verify service URL (optional)
-# probe_url = "https://verify.yourdomain.com/probe"
+# Custom verify service host (optional)
+# verify_host = "https://verify.yourdomain.com"
 
 # ----------------------------
 # Mailboxes
--- a/docs/guide/setup.md
+++ b/docs/guide/setup.md
@@ -258,7 +258,12 @@ If you prefer not to use the public instance:
 
 4. Point aimx to your instance in `config.toml`:
    ```toml
-   probe_url = "https://verify.yourdomain.com/probe"
+   verify_host = "https://verify.yourdomain.com"
+   ```
+
+   Or override it per-invocation with `--verify-host`:
+   ```
+   aimx verify --verify-host https://verify.yourdomain.com
    ```
 
 The verify service provides:
--- a/docs/guide/troubleshooting.md
+++ b/docs/guide/troubleshooting.md
@@ -13,7 +13,12 @@ aimx status
 
 # Check port 25 connectivity (outbound and inbound)
 aimx verify
+
+# Test against a self-hosted verify service instead of the default
+aimx verify --verify-host https://verify.yourdomain.com
 ```
+
+The `--verify-host` flag is also accepted by `aimx setup` and `aimx preflight`, and overrides the `verify_host` value from `config.toml` for the current invocation.
 
 ## Common issues
 
--- a/services/verify/README.md
+++ b/services/verify/README.md
@@ -46,9 +46,15 @@ To self-host (replacing `check.aimx.email`):
 2. Point your domain's DNS to the server
 3. Configure a reverse proxy (Caddy/nginx) for HTTPS on the HTTP port
 4. Run with `BIND_ADDR=127.0.0.1:3025`
-5. In your aimx `config.toml`, set `probe_url`:
+5. In your aimx `config.toml`, set `verify_host` to the base URL of your instance (no path):
    ```toml
-   probe_url = "https://verify.yourdomain.com/probe"
+   verify_host = "https://verify.yourdomain.com"
+   ```
+   aimx appends `/probe` to this base URL when making the HTTP check.
+
+   You can also override it per-invocation:
+   ```
+   aimx verify --verify-host https://verify.yourdomain.com
    ```
 
 No MTA, no email sending, no DNS records needed on the verify server -- it only needs:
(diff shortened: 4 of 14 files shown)


uzyn:
@codex review this for soundness, completeness and for any issues.

uzyn:
@claude review this for soundness, completeness and for any issues.

uzyn (commented):
## Changes Overview

Renames the config key `probe_url` → `verify_host` (now a base URL with no `/probe` suffix; aimx appends `/probe` internally), adds a `--verify-host` flag to `verify`, `setup`, and `preflight` with precedence CLI > config > default, and trims trailing slashes on assignment. No legacy `probe_url` shim — the old key is silently ignored by serde.

I ran `cargo check --all-targets`, `cargo clippy --all-targets -- -D warnings`, and the targeted `verify_host`/`resolve_host` tests locally — all clean.

## Scope Alignment

The implementation matches the PR description. `/probe` is consistently moved from the config value to `check_inbound_port25` at `src/setup.rs:247`. Trailing-slash handling lives in `RealNetworkOps::from_verify_host` (`src/setup.rs:203-210`) and is exercised by `real_network_ops_from_verify_host_strips_trailing_slash`. All eight call sites that construct a `Config` in tests were updated. No scope creep.

## Potential Bugs

None that affect correctness. A few robustness gaps worth noting (all non-blocking):

**1. `--verify-host` accepts any string, no URL validation** — Non-blocker
`--verify-host \"\"`, `--verify-host check.aimx.email` (no scheme), `--verify-host ftp://foo` all flow through. Tracing an empty string: `from_verify_host(\"\")` → `verify_host = \"\"`, `probe_url = \"/probe\"`, `curl -s -m 60 /probe` fails silently, and the user sees `FAIL` with no hint that the flag was garbage. Similarly, a bare hostname like `check.aimx.email` gives `curl check.aimx.email/probe` (interpreted as a relative path → curl error). Cheap fix: reject values that don't start with `http://` or `https://` in `from_verify_host`, or add a clap `value_parser`.

**2. IPv6 literal hosts mangle the SMTP address** — Non-blocker, preexisting
`derive_smtp_addr_from_verify_host(\"https://[::1]:3025\")` → `without_scheme = \"[::1]:3025\"` → contains `:` → split on `:`, take first → `\"[\"` → returns `\"[:25\"`. The port-strip at `src/setup.rs:221-225` is naive. This is preexisting (same shape as the old `derive_smtp_addr_from_probe_url`), but the rename is a natural moment to harden it. Feel free to defer.

## Security

Nothing exploitable. The verify-host URL flows to `curl` via separate argv (no shell injection), `-s -m 60` without `-L` (no redirect following), and TLS verification is not disabled. Untrusted input here is the user's own CLI flag, not a remote attacker's.

## Test Coverage

Good coverage on the new logic: four precedence tests for `resolve_verify_host`, plus `real_network_ops_from_verify_host_strips_trailing_slash`. Two gaps:

**1. No test pins legacy `probe_url` as silently ignored.** The PR description promises this, and it's the actual migration path users will hit, but there's no regression guard. `config_with_legacy_verify_address_parses` covers a different removed field. Add something like:

```rust
#[test]
fn config_with_legacy_probe_url_is_silently_ignored() {
    let toml_str = \"domain = \\\"test.com\\\"\nprobe_url = \\\"https://old.example.com/probe\\\"\n[mailboxes]\n\";
    let config: Config = toml::from_str(toml_str).unwrap();
    assert!(config.verify_host.is_none());
}
```

This also guards against a future `#[serde(deny_unknown_fields)]` quietly breaking migration.

**2. `main::resolve_verify_host` has no tests.** `Setup` and `Preflight` go through the main.rs copy (`src/main.rs:71-79`); only `Verify` goes through the tested `verify::resolve_verify_host`. See Code Quality #1.

## Code Quality

**1. Duplicated `resolve_verify_host`** — Non-blocker
Two implementations with different signatures:
- `src/main.rs:71-79` — private, takes `data_dir: Option<&Path>`, loads config internally, hard-codes `DEFAULT_VERIFY_HOST`.
- `src/verify.rs:20-31` — `pub`, takes `Option<&Config>` and `default: &str`.

They do the same thing. `Setup --verify-host` and `Preflight --verify-host` rely on the untested copy; `Verify --verify-host` uses the tested one. Recommend consolidating on the `verify.rs` version and having `main.rs` load the config once then call it — that way all three commands share the same precedence logic and the same tests.

**2. `verify::resolve_verify_host` is unnecessarily `pub`** — Non-blocker
Only called from its own module. Dropping `pub` is free cleanup (or keep it `pub` if #1 consolidates main.rs onto it).

**3. Silent migration UX trap** — Non-blocker, informational
A user with an existing self-hosted config (`probe_url = \"https://verify.theirdomain.com/probe\"`) will silently fall back to the public `check.aimx.email` endpoint instead of their private one. That's a privacy-adjacent surprise. The PR body acknowledges this explicitly under \"pre-launch, old key is dropped entirely\", so it's an intentional choice — flagging in case you want to soften it with a one-time warning when `Config::load_from_data_dir` sees an unknown `probe_url` key.

**4. Minor naming inconsistency** — Non-blocker
`verify::run(data_dir, host_override)` but the CLI enum uses `verify_host` and main.rs does `verify::run(..., verify_host.as_deref())`. Renaming the `verify::run` parameter to `verify_host` would match the rest of the codebase end-to-end.

**Overall verdict**: Needs minor fixes — the core rename is correct, tested, and clippy/fmt clean. No blockers; the items below are polish.

**Non-blockers** (should fix in this PR or a follow-up):
- Consolidate the duplicate `resolve_verify_host` in `src/main.rs:71` with the one in `src/verify.rs:20` so Setup/Preflight share Verify's tested precedence logic
- Add a test pinning that a legacy `probe_url` TOML field is silently ignored
- Validate that `--verify-host` starts with `http://` or `https://` (reject empty/bare-host/non-http schemes with a clear error)
- Drop `pub` on `verify::resolve_verify_host` if it stays private

**Nice-to-haves**:
- Harden `derive_smtp_addr_from_verify_host` against IPv6 literals (preexisting)
- Rename `verify::run`'s second parameter from `host_override` to `verify_host` for consistency
- Consider a one-time warning when an old `probe_url` key is detected in config to avoid silent fallback to the public service

PR #21, opened 2026-04-11
[Sprint 12] aimx-verify Security Hardening + /reach Endpoint

uzyn:
## Summary

Sprint 12 fixes three real bugs in the aimx-verify service discovered during post-Sprint-11 debugging, and adds a new plain-TCP \`/reach\` endpoint for Sprint 13's preflight fix.

### S12.1 - 4-Layer Caddy self-probe fix + /reach endpoint
- **Layer 1 (network):** default \`BIND_ADDR\` moved from \`0.0.0.0:3025\` to \`127.0.0.1:3025\`
- **Layer 2 (proxy):** canonical \`services/verify/Caddyfile\` committed with \`header_up -X-Forwarded-For\` and \`header_up X-AIMX-Client-IP {remote_host}\`, plus \`{\$DOMAIN}\` env var interpolation
- **Layer 3 (app):** \`resolve_client_ip(peer, headers)\` trust-boundary helper — non-loopback peer uses \`peer.ip()\`; loopback peer requires a valid \`X-AIMX-Client-IP\` header (rejecting loopback/private/link-local/unspecified values). Never reads \`X-Forwarded-For\`.
- **Layer 4 (target guard):** \`is_blocked_target()\` rejects loopback, unspecified, link-local, broadcast, multicast, RFC 1918, RFC 6598 (CGNAT), and RFC 4193 ranges. Applied in both \`/probe\` and \`/reach\` before any outbound connection.
- **New \`GET /reach\`:** plain TCP connect to \`caller_ip:25\` with 10s timeout, same response shape as \`/probe\`, no SMTP handshake. Any listening socket satisfies reachability.
- Both handlers return **HTTP 400** when the peer is loopback and \`X-AIMX-Client-IP\` is missing/unparseable/rejected.

### S12.2 - Fix self-EHLO trap in built-in SMTP listener
\`handle_smtp_connection\` rewritten around a line-oriented \`BufReader\` loop with per-line 10s timeouts and an overall 30s wall-clock timeout. Implements a minimal but correct SMTP state machine: banner → (\`EHLO\`/\`HELO\` → \`250 hostname\` | \`QUIT\` → \`221 Bye\` | other → \`500 Command not recognized\`). Closes cleanly on peer close or idle timeout. The self-loop round-trip test proves \`check_port25_ehlo\` now completes successfully against the listener.

### S12.3 - Docs
- \`services/verify/README.md\` rewritten for the two-endpoint model, Caddy deployment contract, loopback-bind default, and security rationale for not exposing \`BIND_ADDR=0.0.0.0:3025\` directly
- \`docs/manual-setup.md\` Part A updated to reference the canonical Caddyfile, the \`/reach\` vs \`/probe\` split, and the minimal-SMTP listener semantics
- \`docs/prd.md\` FR-38 expanded to describe the two-endpoint design and target-guard security model
- Repo-root \`README.md\` NOT modified (per sprint spec)

## Test plan
- [x] \`cargo fmt -- --check\` clean in \`services/verify/\`
- [x] \`cargo clippy --all-targets -- -D warnings\` clean in \`services/verify/\`
- [x] \`cargo test\` — all 35 tests pass
- [x] New tests cover every acceptance criterion in S12.1 and S12.2:
  - \`resolve_client_ip\`: direct public IPv4/IPv6, loopback-with-valid-header, loopback-missing, loopback-with-loopback/private/unspecified/link-local/unparseable header, IPv6 loopback-with-header
  - Layer 4 target guard: loopback, IPv6 loopback, 0.0.0.0, ::, RFC 1918 boundaries (including 172.16/12 edge cases), link-local, fe80::, fc00::, fd00::
  - Handler 400 responses for \`/probe\` and \`/reach\` on loopback peer without/with private header
  - \`/reach\` unreachable host returns \`reachable: false\`
  - Plain TCP connect to a listening (non-SMTP) socket succeeds — the semantic difference from \`/probe\`
  - SMTP session: \`EHLO\` → \`250\`, \`HELO\` → \`250\`, unknown command → \`500\` (session continues), peer close mid-session, round-trip \`check_port25_ehlo\` against \`handle_smtp_connection\`
- [x] \`grep -r X-Forwarded-For services/verify/src/\` — only the doc comment explaining why we don't read it; no runtime code references

## Breaking changes
Default HTTP bind is now \`127.0.0.1:3025\` instead of \`0.0.0.0:3025\`. Operators of the currently-deployed service must either (a) install Caddy with the canonical Caddyfile in front, or (b) set \`BIND_ADDR=0.0.0.0:3025\` explicitly (not recommended — no trust boundary for \`X-AIMX-Client-IP\`). Documented in the verify README.


uzyn (commented):
# Sprint 12 Code Review — PR #21

## Sprint Goal Assessment

Sprint 12 set out to fix three post-Sprint-11 bugs (Caddy self-probe loop, XFF SSRF/port-scan risk, self-EHLO trap) and add a plain-TCP `/reach` endpoint for Sprint 13's preflight. The PR covers all three stories with real code, thoughtful layering, thorough unit tests, and good docs. The 4-layer defense model (network/proxy/app/target guard) is well-implemented and the SMTP state machine rewrite is clean.

However, **this is a security hardening sprint, and I found one real SSRF bypass in the Layer 4 target guard**. The Layer 4 guard's stated purpose is defense-in-depth — "Each layer fails closed without the others" (sprint spec). The current implementation does not fail closed for IPv4-mapped IPv6 addresses, which is exactly the kind of fallback layer the sprint is supposed to harden.

---

## S12.1 — 4-Layer Caddy Self-Probe Fix + /reach Endpoint

| AC | Status | Notes |
|---|---|---|
| Default `BIND_ADDR` → `127.0.0.1:3025` | Met | `main.rs:314` |
| Canonical `Caddyfile` committed | Met | Correct directives; `{$DOMAIN:check.aimx.email}` default present |
| `resolve_client_ip` trust-boundary helper | Met (with bug) | See Blocker 1 below |
| `/probe` returns 400 on loopback peer w/o valid header | Met | |
| `GET /reach` route with plain TCP + 10s timeout | Met | |
| Layer 4 target guard on `/probe` and `/reach` | **Partial** | Blocker 1: IPv4-mapped IPv6 bypass |
| App does not read `X-Forwarded-For` | Met | Only a doc comment references it |
| Unit tests for `resolve_client_ip` (direct v4/v6, loopback+valid, missing, loopback, private, unspecified, link-local, unparseable) | Met | All present |
| Unit test `/probe` 400 loopback w/o header | Met | |
| Unit test `/reach` 400 loopback w/o header | Met | |
| Unit test target guard rejects listed ranges | Met | Goes beyond spec (CGNAT, broadcast, multicast added) |
| `/reach` unreachable host → `reachable: false` | Met | |
| `/reach` listening socket → `reachable: true` | Met (via indirect helper test) | |
| Integration test `/probe` resolves via header (not `127.0.0.1`) | **Partial** | Only `/reach` has this test (`reach_resolves_caller_ip_from_header_on_loopback`). `/probe` has no equivalent; Nit 1 below |
| `cargo fmt`, `clippy -D warnings`, `cargo test` clean | Met | 35 tests pass locally |

---

## S12.2 — Fix Self-EHLO Trap in Built-in SMTP Listener

| AC | Status | Notes |
|---|---|---|
| `EHLO` → `250 {hostname}`, session continues | Met | |
| `HELO` → `250 {hostname}`, session continues | Met | |
| `QUIT` → `221 Bye`, clean close | Met | |
| Unknown command → `500 Command not recognized`, continues | Met | |
| Clean close on peer close / read timeout | Met | `_ => return Ok(())` covers both |
| Overall wall-clock timeout (~30s) | Met | `SMTP_CONNECTION_TIMEOUT = 30s`, wrapped around `smtp_session` |
| Full exchange test (`220`/`EHLO`/`250`/`QUIT`/`221`) | Met | `smtp_session_ehlo_then_quit` |
| Unknown command test (500 + session continues) | Met | `smtp_session_unknown_command_returns_500` |
| Peer close mid-session test | Met | `smtp_session_peer_close_is_clean` |
| Idle timeout test | **Missing** | Nit 2 — no test explicitly waits for `SMTP_LINE_TIMEOUT` to fire |
| Existing `smtp_listener_sends_banner_and_bye` test updated/replaced | Met | Removed; new tests cover the semantics |
| Round-trip `check_port25_ehlo` ↔ `handle_smtp_connection` | Met | `self_loop_ehlo_handshake_round_trip` |

---

## S12.3 — Docs

All documentation ACs met:
- `services/verify/README.md` has the Caddy deployment section, two-endpoint API docs, loopback-bind default + security note, updated systemd example.
- `docs/manual-setup.md` Part A updated end-to-end for the new contract.
- `docs/prd.md` FR-38 rewritten to describe `/reach` + `/probe` + target guard.
- Repo-root `README.md` correctly untouched.

Minor: the README's sentence "`aimx appends /probe (or /reach) to this base URL when making HTTP checks`" is accurate for the post-Sprint-13 world but at the moment only `/probe` is called by any aimx client — just a forward-looking phrasing, not a problem.

---

## Test Coverage Gaps

1. **No `/probe` integration test that exercises `resolve_client_ip` via header** (the spec explicitly called this out). `reach_resolves_caller_ip_from_header_on_loopback` covers `/reach` but not `/probe`. Since both handlers share the same resolver path, this is a minor gap, but the spec asked for both.
2. **No test for `SMTP_LINE_TIMEOUT` idle close.** The spec AC says "idle timeout closes the connection" as a test requirement; the current tests cover peer close but not slow/silent clients.
3. **No test for IPv4-mapped IPv6 in either the target guard or `resolve_client_ip`.** This omission hid the bypass below.
4. **No test for the non-loopback branch where the direct peer is itself a private/link-local IP** (e.g., a LAN client hitting `BIND_ADDR=0.0.0.0:3025`). The handler-level `is_blocked_target` would catch it, but there's no regression test.

---

## Potential Bugs

### Blocker 1 — IPv4-mapped IPv6 bypass in `is_blocked_target` (SSRF / port-scan-as-a-service)

`services/verify/src/main.rs:68-113`

`is_blocked_target` and `is_blocked_ipv6` do not canonicalize IPv4-mapped IPv6 addresses. Any address of the form `::ffff:a.b.c.d` bypasses every guard:

- `IpAddr::is_loopback()` on `::ffff:127.0.0.1` returns **false** (Rust only considers `::1` an IPv6 loopback).
- `IpAddr::is_unspecified()` on `::ffff:0.0.0.0` returns **false** (only `::` is unspecified).
- `(seg[0] & 0xffc0) == 0xfe80` — `::ffff:*` has `seg[0] == 0x0000`, so false.
- `(seg[0] & 0xfe00) == 0xfc00` — same, false.
- `is_multicast()` — false for mapped addresses.

I verified this on Linux with a small program: `TcpStream::connect("[::ffff:127.0.0.1]:<port>")` successfully connects to the underlying IPv4 address on loopback. The dual-stack socket silently downgrades to an IPv4 connection.

**Concrete attack scenarios:**

1. **Loopback header path (most direct).** If a misconfigured reverse proxy (or a future Caddyfile edit) ever allows a client-controlled `X-AIMX-Client-IP` through, an attacker sending `X-AIMX-Client-IP: ::ffff:169.254.169.254` would get the verify service to probe AWS/GCP instance metadata at `169.254.169.254:25`. Same for `::ffff:10.0.0.1`, `::ffff:127.0.0.1`, etc. Today this is blocked by Caddy's `header_up ... {remote_host}` replace semantics, so the layer isn't actively exploitable in the canonical deployment — but Layer 4 is explicitly designed as a defense-in-depth layer that "fails closed without the others" (sprint spec, S12.1 rationale). It currently does not fail closed.
2. **Dual-stack `BIND_ADDR=[::]:3025`.** In this bind mode (which the README does not forbid; it just warns against `0.0.0.0:3025`), an IPv4 client connecting to loopback shows up with `peer.ip() == ::ffff:127.0.0.1`. `resolve_client_ip` checks `peer.ip().is_loopback()`, which is **false** for the mapped form → it returns `Some(::ffff:127.0.0.1)` as the "direct" caller. The handler's `is_blocked_target` also fails to catch it. The request then probes `::ffff:127.0.0.1:25` (the service's own SMTP listener), resurrecting the self-probe bug the sprint is fixing.
3. **Self-hosted proxies other than Caddy.** The README contemplates self-hosters ("To self-host... install Caddy..."), but someone who swaps Caddy for nginx/traefik/envoy without the right header handling could inadvertently pass through an attacker-supplied mapped-IPv6 value. Layer 4 is what should save them — and it doesn't.

**Suggested fix:** Canonicalize IPv4-mapped IPv6 before applying the guards. Something like:

```rust
fn canonicalize(ip: IpAddr) -> IpAddr {
    match ip {
        IpAddr::V6(v6) => match v6.to_ipv4_mapped() {
            Some(v4) => IpAddr::V4(v4),
            None => IpAddr::V6(v6),
        },
        v4 => v4,
    }
}
```

Apply it at the top of `is_blocked_target` (so `resolve_client_ip`'s header check sees the v4 form) and also inside `resolve_client_ip` before the `peer.ip().is_loopback()` check (so the dual-stack-peer case is handled). Add tests:

```rust
for ip_str in ["::ffff:127.0.0.1", "::ffff:10.0.0.1", "::ffff:169.254.1.1", "::ffff:0.0.0.0"] {
    let ip: IpAddr = ip_str.parse().unwrap();
    assert!(is_blocked_target(&ip), "{ip_str} should be blocked");
}
```

Note: `Ipv6Addr::to_ipv4_mapped` is stable since Rust 1.63. Check `rust-version` in `Cargo.toml`.

---

## Security Issues

The sprint's whole point is security hardening. Against the SSRF/port-scan threat model:

- **XFF handling:** correct — `X-Forwarded-For` is never read, only mentioned in a doc comment. `grep` confirms.
- **Trust boundary:** correct — non-loopback peer is authoritative, loopback peer requires a trusted proxy-set header. Caddy's `header_up <name> <value>` replaces rather than appends, so clients cannot pre-seed.
- **Target guard:** **has the IPv4-mapped IPv6 bypass described above.** This is the main security finding.
- **Header parsing:** `value.trim().parse()` on the header is reasonable; parse errors correctly yield `None`.
- **DoS surface in `smtp_session`:** `BufReader::read_line` has no line length cap. With a 10-second per-line timeout, a single client can push ~tens of MB into a growing `String` buffer before timing out. Multiplied by unbounded concurrent connections (`run_smtp_listener` spawns a task per accept with no semaphore), this is a theoretical resource-exhaustion vector. Not a blocker — sprint scope was correctness, not DoS hardening — but worth a note or a `BufReader::take(1024)`-style cap in a follow-up. (Nit 3.)
- **`check_port25_tcp` / `check_port25_ehlo`:** both run AFTER `is_blocked_target`, so fixing Blocker 1 closes the chain. No other target path bypasses the guard.

---

## Alignment with PRD

FR-38 is cleanly extended to describe the two-endpoint split, the Caddy-injected header, and the target guard. The implementation matches:

- `/reach` → plain TCP connect (for `aimx preflight` pre-install). ✓
- `/probe` → full SMTP EHLO handshake (for `aimx setup` / `aimx verify` post-install). ✓
- Both apply the target guard rejecting loopback/unspecified/link-local/RFC 1918/RFC 4193. ✓ (with the caveat of Blocker 1)
- Port 25 listener (FR-39b) unchanged semantically but now correct. ✓

FR-39 (email echo removed) and FR-40 (open source / self-hostable) remain unchanged and are unaffected.

---

## Summary / Verdict

**Verdict: Needs minor fixes** (one security bug + small test gaps).

The sprint is overall well-executed: the 4-layer design is sound, the code is clean, tests are thorough, and the docs are comprehensive. The SMTP state machine rewrite (S12.2) is correct and the round-trip test elegantly proves the self-loop is now well-formed. Docs (S12.3) are complete.

The one blocker is the IPv4-mapped IPv6 bypass in `is_blocked_target`, which directly undermines the Layer 4 defense-in-depth goal that the sprint was specifically chartered to deliver. It's a small fix (5-10 lines + tests) but it belongs in this sprint, not deferred, because it's squarely in scope.

### Blockers

1. **IPv4-mapped IPv6 bypasses `is_blocked_target` and `resolve_client_ip`'s loopback check.** Canonicalize via `Ipv6Addr::to_ipv4_mapped`. Add tests for `::ffff:127.0.0.1`, `::ffff:10.0.0.1`, `::ffff:169.254.1.1`, `::ffff:0.0.0.0`.

### Non-Blockers (nits / follow-ups)

1. **Missing `/probe` integration test for the header-based resolution path.** Spec called it out; only `/reach` has one. Easy to add.
2. **Missing idle-timeout test for `handle_smtp_connection`.** Spec AC listed it. Can use `tokio::time::pause()` + `advance()` to avoid real sleep.
3. **No `BufReader::take` / max-line-length cap in `smtp_session`.** Low-severity DoS consideration; could be a follow-up.
4. **Per-connection concurrency in `run_smtp_listener` is unbounded.** Follow-up hardening.
5. **Minor:** README line "aimx appends `/probe` (or `/reach`) to this base URL" — currently only `/probe` is appended by any aimx client; `/reach` client-side wiring lands in Sprint 13. Not wrong, just slightly ahead of the code. Ignore if intentional.

Once Blocker 1 is addressed (and ideally Nits 1 and 2 at the same time), this is ready to merge.

uzyn:
## Sprint 12 Review Response

Thanks for the thorough review. Addressed the blocker and all five non-blocker nits in `21688ee`.

### Blocker — fixed

**IPv4-mapped IPv6 bypass in `is_blocked_target` and `resolve_client_ip`.** Confirmed the reviewer's analysis: Rust's `Ipv6Addr::is_loopback`/`is_unspecified` only match canonical forms, and our segment-mask checks likewise miss `::ffff:*`. Linux dual-stack `TcpStream::connect` silently downgrades to the underlying IPv4 address, so every mapped form was a live SSRF/self-probe vector against the Layer 4 fallback guard.

Fix: added a `canonicalize_ip` helper that applies `Ipv6Addr::to_ipv4_mapped` and calls it at the top of both `resolve_client_ip` (once for the peer, once for the parsed header value) and `is_blocked_target`. `rustc 1.94` is well above the 1.63 minimum for `to_ipv4_mapped`.

New regression tests:
- `is_blocked_target_rejects_ipv4_mapped_ipv6` — covers `::ffff:127.0.0.1`, `::ffff:10.0.0.1`, `::ffff:10.255.255.255`, `::ffff:172.16.0.1`, `::ffff:192.168.1.1`, `::ffff:169.254.1.1`, `::ffff:169.254.169.254` (cloud metadata), `::ffff:0.0.0.0`, `::ffff:100.64.0.1` (CGNAT).
- `is_blocked_target_allows_ipv4_mapped_public` — `::ffff:1.1.1.1`, `::ffff:8.8.8.8`, `::ffff:203.0.113.1` must still pass (no over-blocking).
- `resolve_client_ip_rejects_ipv4_mapped_loopback_header` — attacker-controlled mapped payloads in `X-AIMX-Client-IP` are rejected.
- `resolve_client_ip_treats_ipv4_mapped_loopback_peer_as_loopback` — the dual-stack `[::]:3025` case where a v4 client shows up as `::ffff:127.0.0.1` is now correctly treated as a loopback peer requiring a trusted header.
- `resolve_client_ip_canonicalizes_ipv4_mapped_public_header` — mapped public v4 in the header is returned as its v4 form for consistent downstream handling.

### Non-blockers — all fixed

1. **`/probe` integration test for header-based resolution.** Added `probe_resolves_caller_ip_from_header_on_loopback` mirroring the existing `/reach` one. Uses TEST-NET-1 so the handshake fails and the assertion is on `result.ip == "192.0.2.1"` (proving the header was used, not the peer).
2. **Idle-timeout test for `handle_smtp_connection`.** Added `smtp_session_idle_times_out` using `#[tokio::test(start_paused = true)]` + `tokio::io::duplex` so the `SMTP_LINE_TIMEOUT` fires in virtual time. Needed to make `smtp_session` generic over `AsyncRead + AsyncWrite` (trivial change, still called with `TcpStream` from `handle_smtp_connection`) and to add `test-util` as a dev-only tokio feature in `Cargo.toml`.
3. **Line-length cap in `smtp_session`.** Replaced `reader.read_line(...)` with `(&mut reader).take(SMTP_MAX_LINE_BYTES).read_until(b'\n', ...)` where `SMTP_MAX_LINE_BYTES = 1024` (generous headroom over RFC 5321's 512). Added `smtp_session_oversized_line_is_bounded` which slams >2 KiB of unterminated junk at the session and verifies the cap triggers cleanly and the session is still responsive to `QUIT` afterward.
4. **Unbounded concurrency in `run_smtp_listener`.** Left as-is with an explicit comment explaining why it's deferred: per-connection bounds are now tight (30s wall, 10s per-line, 1 KiB per-line), and a semaphore is tracked for Sprint 14 since general DoS hardening was not part of Sprint 12's charter. Happy to move this into Sprint 12 if you'd prefer — it's a ~10-line change.
5. **README wording.** Softened `aimx appends /probe (or /reach) to this base URL` to `aimx appends /probe to this base URL when making HTTP checks. (A future release will also append /reach for pre-install preflight.)` to match the current state until Sprint 13 wires up the client side.

### Test suite

`cargo fmt -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test` green: **43 passed** (was 35, added 8).

### What was intentionally left as-is

- The unbounded concurrency in `run_smtp_listener` (non-blocker #4) — deferred to Sprint 14 with an inline comment and a note above. Reviewer explicitly flagged this as a follow-up. Reversible if you'd rather pull it into this sprint.

Ready for another pass.

uzyn (commented):
## Sprint 12 Re-Review — APPROVED (posted as comment; GitHub blocks self-approval)

All previously flagged issues have been verified as resolved (or deferred with sound reasoning). Re-ran the full test suite locally on commit `21688ee`: **43 tests pass**, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --check` clean.

### Blocker — IPv4-mapped IPv6 bypass — FIXED

Verified the fix in `services/verify/src/main.rs`:

- `canonicalize_ip()` helper (lines 52-60) correctly uses `Ipv6Addr::to_ipv4_mapped()` to collapse `::ffff:a.b.c.d` to the underlying v4 form.
- Applied at the top of `is_blocked_target` (line 98), so every downstream guard check (`is_loopback`, `is_unspecified`, `is_blocked_ipv4`, `is_blocked_ipv6`) now sees the canonical representation.
- Applied to **both** entry points in `resolve_client_ip`: peer IP (line 71) and parsed header value (line 78). The dual-stack `[::]:3025` case where a v4 loopback client shows up as `::ffff:127.0.0.1` is now correctly treated as a loopback peer requiring a trusted header.

Regression coverage is thorough and exactly targets the bypass:

- `is_blocked_target_rejects_ipv4_mapped_ipv6` — all the payloads I flagged plus cloud metadata (`::ffff:169.254.169.254`) and CGNAT (`::ffff:100.64.0.1`).
- `is_blocked_target_allows_ipv4_mapped_public` — proves canonicalization doesn't over-block (`::ffff:1.1.1.1`, `::ffff:8.8.8.8`, `::ffff:203.0.113.1` pass through).
- `resolve_client_ip_rejects_ipv4_mapped_loopback_header` — attacker-controlled mapped payloads in `X-AIMX-Client-IP` are rejected.
- `resolve_client_ip_treats_ipv4_mapped_loopback_peer_as_loopback` — the dual-stack case works.
- `resolve_client_ip_canonicalizes_ipv4_mapped_public_header` — mapped public v4 in the header is returned as its v4 form, ensuring consistent downstream logging/checks.

Nothing slips through.

### Non-blockers

1. **`/probe` integration test for header resolution** — `probe_resolves_caller_ip_from_header_on_loopback` verified; mirrors the `/reach` variant and asserts `result.ip == "192.0.2.1"` (proving the header was used, not the peer).
2. **Idle-timeout test** — `smtp_session_idle_times_out` uses `#[tokio::test(start_paused = true)]` + `tokio::io::duplex(1024)`, so virtual time auto-advances and `SMTP_LINE_TIMEOUT` fires immediately. Nice — no 10s wall-clock wait in tests. Required making `smtp_session` generic over `AsyncRead + AsyncWrite`; that refactor is trivially safe (`tokio::io::split` accepts the generic bound, verified by clippy).
3. **Line-length cap** — `SMTP_MAX_LINE_BYTES = 1024` via `(&mut reader).take(SMTP_MAX_LINE_BYTES).read_until(b'\n', ...)` is the correct idiom. The fresh `Take` wrapper per iteration means the allowance resets each line, which is the right behavior. `smtp_session_oversized_line_is_bounded` test confirms the cap engages and the session remains responsive to subsequent `QUIT`.
4. **Unbounded concurrency in `run_smtp_listener`** — deferred to Sprint 14 with an explicit inline comment (lines 261-267). Accepted: per-connection bounds are already tight (30s wall, 10s per-line, 1 KiB per-line), general DoS hardening is explicitly out of Sprint 12's charter, and Sprint 14's observability work is the natural home for a semaphore.
5. **README wording** — softened to `aimx appends /probe ... (A future release will also append /reach...)`.

### Nit (non-blocking, post-merge cleanup if you feel like it)

In the refactored `smtp_session`:
```rust
let (reader, writer) = tokio::io::split(stream);
let mut reader = BufReader::new(reader);
let mut writer = writer;
```
The `let mut writer = writer;` re-binding could fold into the destructuring pattern as `let (reader, mut writer) = tokio::io::split(stream);`. No behavioral difference.

```
Sprint 12: aimx-verify security hardening + /reach endpoint

Fixes three post-Sprint-11 bugs and adds a plain-TCP /reach endpoint
for Sprint 13's preflight chicken-and-egg fix.

S12.1 4-layer Caddy self-probe fix + /reach endpoint:
  - Default HTTP bind moved from 0.0.0.0:3025 to 127.0.0.1:3025 so the
    service is not directly reachable from the public internet.
  - Canonical services/verify/Caddyfile committed with
    header_up -X-Forwarded-For and header_up X-AIMX-Client-IP
    {remote_host}, plus {$DOMAIN} interpolation.
  - resolve_client_ip(peer, headers) trust boundary: non-loopback peer
    uses peer.ip(); loopback peer requires X-AIMX-Client-IP and rejects
    loopback/unspecified/link-local/RFC1918/RFC6598/RFC4193 values.
    X-Forwarded-For is never parsed.
  - is_blocked_target() Layer 4 guard rejects loopback, unspecified,
    link-local, broadcast, multicast, RFC 1918, RFC 6598 CGNAT, and
    RFC 4193 ULA ranges before any outbound connect. IPv4-mapped IPv6
    addresses (::ffff:a.b.c.d) are canonicalized via to_ipv4_mapped()
    so dual-stack bypass (::ffff:127.0.0.1, ::ffff:169.254.169.254,
    ::ffff:10.0.0.1, etc.) is closed.
  - New GET /reach: plain 10s TCP connect to caller_ip:25, same
    response shape as /probe, no SMTP handshake required.
  - Both handlers return HTTP 400 when a loopback peer supplies no
    valid X-AIMX-Client-IP.

S12.2 Minimal correct SMTP listener:
  - handle_smtp_connection rewritten around a line-oriented BufReader
    loop: banner -> (EHLO|HELO -> 250 hostname | QUIT -> 221 Bye |
    other -> 500) state machine, per-line 10s timeout, overall 30s
    wall-clock timeout, 1 KiB per-line cap to bound memory under a
    misbehaving peer. Closes the self-EHLO trap that made /probe
    fail against itself.

S12.3 Docs:
  - services/verify/README.md rewritten for the two-endpoint model,
    the Caddy deployment contract, the loopback-bind default, and
    the security rationale.
  - docs/manual-setup.md Part A updated for the canonical Caddyfile,
    the /reach vs /probe split, and the minimal-SMTP semantics.
  - docs/prd.md FR-38 expanded for the two-endpoint design and the
    trust-boundary + target-guard model.

Tests: 43 pass. cargo fmt clean, clippy -D warnings clean. Unbounded
listener concurrency is deferred to Sprint 14 with an inline comment.
```

**Verdict: APPROVED — ship it.**

PR #22, opened 2026-04-11
[Sprint 13] Preflight Flow Fix + PTR Display

uzyn:
## Summary

Fixes two bugs found during post-Sprint-11 debugging of the preflight / verify flow:

- **S13.1 — Preflight chicken-and-egg.** `aimx preflight` is meant to run on a fresh VPS before OpenSMTPD is installed, but its inbound check called `{verify_host}/probe` which does a full SMTP EHLO handshake. On a fresh VPS nothing is listening on port 25, so the handshake failed and preflight reported `Inbound port 25 is not reachable` even when port 25 was actually reachable at the TCP level. Preflight now routes its inbound check at the plain-TCP `/reach` endpoint introduced in Sprint 12. `aimx setup` (post-install) and `aimx verify` continue to use `/probe` for the full EHLO validation — no regression there.
- **S13.2 — PTR display ordering bug.** `check_ptr` was emitting its own `println!(\"  PTR record: {ptr}\")` before returning `PreflightResult::Pass`, which interleaved with the caller's unflushed `print!(\"  PTR record... \")` and produced mangled output when the inbound check failed. `PreflightResult::Pass` now carries `Option<String>`, `check_ptr` returns the PTR value via that payload, and `run_preflight` prints a single well-formed `  PTR record... PASS (<ptr>)` line.

## Changes

- `NetworkOps` trait: new `check_inbound_reachable()` method (calls `/reach`), `check_inbound_port25()` (calls `/probe`) kept unchanged.
- `RealNetworkOps::check_inbound_reachable()` mirrors `check_inbound_port25()` exactly (curl, 60s timeout, parse `\"reachable\":true`) against the new path.
- New `check_inbound_tcp()` helper alongside `check_inbound()`; `run_preflight()` calls the TCP variant, `run_setup()` and `verify::run_with_net()` continue to call the EHLO variant.
- `PreflightResult::Pass` → `Pass(Option<String>)`; all match arms updated across `src/setup.rs` and `src/verify.rs`.
- `check_ptr()` no longer calls `println!` directly — the PTR value is carried back via `Pass(Some(ptr))` and rendered inline by the caller.
- `run_preflight` refactored into `run_preflight_to<W, E>` taking writer streams so output ordering is testable without touching global stdout.
- All `MockNetworkOps` impls in `src/setup.rs` tests and `src/verify.rs` tests extended to implement both methods.

## Test plan

- [x] `cargo fmt -- --check`
- [x] `cargo test` — 291 unit tests + 35 integration tests, all pass
- [x] New test: `preflight_passes_when_reach_ok_even_if_ehlo_would_fail` — fresh-VPS scenario
- [x] New test: `setup_uses_ehlo_not_reach` — no regression in setup flow
- [x] New test: `verify_uses_ehlo_not_reach` — no regression in `aimx verify`
- [x] New test: `preflight_ptr_inline_pass` — single well-formed `  PTR record... PASS (<ptr>)` line
- [x] New test: `preflight_no_ptr_interleaving_when_inbound_fails` — regression guard for the interleaving bug
- [x] New test: `preflight_ptr_warn_when_missing` — PTR stays non-blocking

## Notes

- PTR stays advisory: `Warn` on missing, `Pass(Some(_))` on present, never `Fail`. `all_pass` unchanged when PTR is missing.
- Display text kept as `  Inbound port 25...` in preflight — the semantic is still \"is my inbound port 25 reachable,\" just validated at the TCP layer rather than at EHLO.

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -32,7 +32,12 @@ pub trait SystemOps {
 
 pub trait NetworkOps {
     fn check_outbound_port25(&self) -> Result<bool, Box<dyn std::error::Error>>;
+    /// Full SMTP EHLO handshake via `{verify_host}/probe`.
+    /// Used by `aimx setup` (post-install) and `aimx verify`.
     fn check_inbound_port25(&self) -> Result<bool, Box<dyn std::error::Error>>;
+    /// Plain-TCP reachability check via `{verify_host}/reach`.
+    /// Used by `aimx preflight` on a fresh VPS before OpenSMTPD is installed.
+    fn check_inbound_reachable(&self) -> Result<bool, Box<dyn std::error::Error>>;
     fn check_ptr_record(&self) -> Result<Option<String>, Box<dyn std::error::Error>>;
     fn get_server_ip(&self) -> Result<IpAddr, Box<dyn std::error::Error>>;
     fn resolve_mx(&self, domain: &str) -> Result<Vec<String>, Box<dyn std::error::Error>>;
@@ -282,6 +287,21 @@ impl NetworkOps for RealNetworkOps {
         }
     }
 
+    fn check_inbound_reachable(&self) -> Result<bool, Box<dyn std::error::Error>> {
+        let reach_url = format!("{}/reach", self.verify_host);
+        let resp = std::process::Command::new("curl")
+            .args(["-s", "-m", "60", &reach_url])
+            .output();
+
+        match resp {
+            Ok(output) if output.status.success() => {
+                let body = String::from_utf8_lossy(&output.stdout);
+                Ok(body.contains("\"reachable\":true") || body.contains("\"reachable\": true"))
+            }
+            _ => Ok(false),
+        }
+    }
+
     fn check_ptr_record(&self) -> Result<Option<String>, Box<dyn std::error::Error>> {
         let ip = self.get_server_ip()?;
         let output = std::process::Command::new("dig")
@@ -354,14 +374,15 @@ pub const COMPATIBLE_PROVIDERS: &[&str] = &[
 
 #[derive(Debug, PartialEq)]
 pub enum PreflightResult {
-    Pass,
+    /// Check passed. Optional detail string (e.g. PTR value) is displayed inline.
+    Pass(Option<String>),
     Fail(String),
     Warn(String),
 }
 
 pub fn check_outbound(net: &dyn NetworkOps) -> PreflightResult {
     match net.check_outbound_port25() {
-        Ok(true) => PreflightResult::Pass,
+        Ok(true) => PreflightResult::Pass(None),
         Ok(false) => PreflightResult::Fail(
             "Outbound port 25 is blocked. Your VPS provider may restrict SMTP traffic.".into(),
         ),
@@ -369,9 +390,25 @@ pub fn check_outbound(net: &dyn NetworkOps) -> PreflightResult {
     }
 }
 
+/// Full SMTP EHLO handshake via `/probe`. Used by `aimx setup` (post-install)
+/// and `aimx verify` — both run after OpenSMTPD is listening on port 25.
 pub fn check_inbound(net: &dyn NetworkOps) -> PreflightResult {
     match net.check_inbound_port25() {
-        Ok(true) => PreflightResult::Pass,
+        Ok(true) => PreflightResult::Pass(None),
+        Ok(false) => PreflightResult::Fail(
+            "Inbound port 25 is not reachable. Check your firewall and VPS provider settings."
+                .into(),
+        ),
+        Err(e) => PreflightResult::Fail(format!("Inbound port 25 check failed: {e}")),
+    }
+}
+
+/// Plain-TCP reachability check via `/reach`. Used by `aimx preflight` on a
+/// fresh VPS before OpenSMTPD is installed — any listening socket (or the
+/// verify service's TCP connect succeeding) satisfies this check.
+pub fn check_inbound_tcp(net: &dyn NetworkOps) -> PreflightResult {
+    match net.check_inbound_reachable() {
+        Ok(true) => PreflightResult::Pass(None),
         Ok(false) => PreflightResult::Fail(
             "Inbound port 25 is not reachable. Check your firewall and VPS provider settings."
                 .into(),
@@ -382,10 +419,7 @@ pub fn check_inbound(net: &dyn NetworkOps) -> PreflightResult {
 
 pub fn check_ptr(net: &dyn NetworkOps) -> PreflightResult {
     match net.check_ptr_record() {
-        Ok(Some(ptr)) => {
-            println!("  PTR record: {ptr}");
-            PreflightResult::Pass
-        }
+        Ok(Some(ptr)) => PreflightResult::Pass(Some(ptr)),
         Ok(None) => PreflightResult::Warn(
             "No PTR (reverse DNS) record found. Set a PTR record at your VPS provider \
              pointing to your domain. This improves deliverability but is not required."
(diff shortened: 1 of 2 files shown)


uzyn (commented):
## APPROVED — Sprint 13 Review (Preflight Flow Fix + PTR Display)

> _Note: GitHub blocks self-approval via `--approve`, so this is filed as a `--comment` with an APPROVED banner. Treat as a ship-it._

### Verdict: READY TO MERGE

Both stories are cleanly implemented, all acceptance criteria are met, the test suite is strong, and the refactor is surgical. Nothing structural to flag.

---

### Sprint goal assessment

Sprint 13 set out to fix two concrete bugs discovered post-Sprint-11:

1. **Preflight chicken-and-egg** — `aimx preflight` failing on fresh VPSes because `/probe` requires a live SMTP responder that isn't installed yet.
2. **PTR display ordering** — `check_ptr` emitting its own `println!` mid-line, interleaving with the caller's unflushed header.

Both are resolved cleanly. Preflight now routes at `/reach` (the plain-TCP endpoint from Sprint 12); `aimx setup` (post-install) and `aimx verify` continue to use `/probe`, so there is no regression in those flows. PTR is now a single well-formed line with the value inline.

---

### S13.1 — Route Preflight Inbound at /reach — Acceptance Criteria

| Criterion | Status |
|---|---|
| `NetworkOps` gains `check_inbound_reachable()` | MET — `src/setup.rs:40` |
| `RealNetworkOps::check_inbound_reachable()` calls `{verify_host}/reach`, parses `"reachable":true`, 60s timeout | MET — `src/setup.rs:290-303`, mirrors `check_inbound_port25` exactly |
| `run_preflight()` uses the reachable variant | MET — routed via new `check_inbound_tcp()` helper at `src/setup.rs:470` |
| `run_setup_with_verify()` continues to use `check_inbound_port25()` (EHLO via `/probe`) | MET — `src/setup.rs:1063`, pinned by `setup_uses_ehlo_not_reach` test |
| `src/verify.rs` continues to use `check_inbound_port25()` | MET — `src/verify.rs:55`, pinned by `verify_uses_ehlo_not_reach` test |
| All mock `NetworkOps` impls in tests implement both methods | MET — `MockNetworkOps` in `src/setup.rs` and `src/verify.rs`, plus new `TrackingNetworkOps` in `src/setup.rs` |
| Unit test: `run_preflight` PASSes when `check_inbound_reachable` returns `Ok(true)` (fresh-VPS scenario) | MET — `preflight_passes_when_reach_ok_even_if_ehlo_would_fail` (also asserts `/probe` is NOT called) |
| Unit test: `run_setup_with_verify` still uses EHLO | MET — `setup_uses_ehlo_not_reach` |
| `cargo fmt`, `cargo clippy -- -D warnings`, `cargo test` clean | MET — 291 unit + 35 integration tests pass, clippy/fmt clean |

### S13.2 — PTR Display Ordering — Acceptance Criteria

| Criterion | Status |
|---|---|
| `check_ptr` no longer calls `println!` | MET — `src/setup.rs:420-430` |
| PTR value returned via `PreflightResult` | MET — `Pass(Option<String>)` extended, all call sites updated |
| `run_preflight` displays PTR inline with `PASS` | MET — `  PTR record... PASS ({ptr})` at `src/setup.rs:483` |
| `Warn` path preserves `WARN\n  {msg}` format | MET — `src/setup.rs:489` |
| `all_pass` stays `true` when PTR is missing | MET — `Warn` path does not set `all_pass = false` |
| Unit test: single well-formed line containing PASS + PTR value | MET — `preflight_ptr_inline_pass` |
| Unit test: interleaving bug does not reproduce when inbound fails | MET — `preflight_no_ptr_interleaving_when_inbound_fails` asserts exactly one PTR line on stdout AND inbound-FAIL index < PTR index |
| Unit test: `None` PTR produces WARN + advisory, no failure | MET — `preflight_ptr_warn_when_missing` |
| `cargo fmt`, `cargo clippy`, `cargo test` clean | MET |

---

### PreflightResult refactor — match-arm sweep

I grep'd the entire `src/` tree for `PreflightResult::Pass` (including the struct `Pass\b` to catch stray variants) and cross-checked against every call site:

- `src/setup.rs`: 4 producers (outbound / inbound / inbound_tcp / ptr) + 4 consumers (`run_preflight_to` × 3, `run_setup` × 3) — all updated correctly
- `src/verify.rs`: 3 consumers (`run_with_net`) — all updated correctly
- `DnsVerifyResult::Pass` is a separate enum and is (correctly) unaffected

No missed arms. The refactor is complete.

---

### Display string match against spec

Spec (sprint.md:1385):
```
  PTR record... PASS (vps-198f7320.vps.ovh.net.)
```

Implementation (`src/setup.rs:483`, `src/verify.rs:68`, `src/setup.rs:1076`):
```rust
write!(out, "  PTR record... ")?;
// ...
PreflightResult::Pass(Some(ptr)) => writeln!(out, "PASS ({ptr})")?,
```

Exact match — two spaces of indentation, `PTR record...`, space, `PASS`, space, `(ptr)`, trailing newline.

---

### Correctness of NetworkOps split + call-site routing

The three call sites are correctly wired:

| Caller | Method called | Endpoint | Justification |
|---|---|---|---|
| `run_preflight_to` (preflight CLI) | `check_inbound_reachable` via `check_inbound_tcp` | `/reach` | Runs on fresh VPS, nothing listening on 25 yet — TCP reachability is the right check |
| `run_setup` | `check_inbound_port25` via `check_inbound` | `/probe` | Sprint 11 reordered setup to install OpenSMTPD BEFORE the port check, so EHLO is valid here |
| `verify::run_with_net` | `check_inbound_port25` via `check_inbound` | `/probe` | Post-setup sanity check, SMTP server is already running |

Both setup and verify are pinned by tracking-mock tests (`setup_uses_ehlo_not_reach`, `verify_uses_ehlo_not_reach`) that actually observe the method call — not just the result — so any future regression that silently rewires a caller would break the test.

No regressions in setup/verify flows.

---

### Test coverage

Particularly strong on the S13.2 regression guard:

- `preflight_no_ptr_interleaving_when_inbound_fails` is the exact scenario from the bug report and asserts:
  1. Exactly one `PTR record` line on stdout (no duplicate from the stray `println!`)
  2. That line is the final-form `  PTR record... PASS (...)` (not the intermediate header)
  3. The inbound-FAIL block appears strictly before the PTR line in stream order

That is a proper regression test — it would have caught the original bug and will catch any future reintroduction.

For S13.1, the `TrackingNetworkOps` mock that records WHICH inbound method was called is the right level of rigor — any future refactor that silently flips a caller to the wrong endpoint would trip these tests.

---

### Minor observations (non-blocking)

- The `run_preflight_to<W, E>` doc comment at `src/setup.rs:438-439` says "Both output streams are routed to `out` in tests so ordering can be asserted" — in practice the new tests use separate `out` and `err` buffers rather than routing both to `out`. The comment is slightly stale; harmless either way.
- The pre-existing `preflight_fails_on_inbound_blocked`, `preflight_fails_when_reach_fails`, `preflight_fails_on_outbound_blocked`, `preflight_passes_when_reach_ok_even_if_ehlo_would_fail`, and `preflight_passes_with_ptr_warning` tests still call the global `run_preflight`, which writes to real stdout during test runs. Not a bug — just noise in `cargo test` output. Future cleanup could migrate them to `run_preflight_to` with captured buffers.
- `check_inbound` and `check_inbound_tcp` share identical Fail/Err handling (different only in the trait method called). Could be DRYed up with a closure, but clippy is clean and the duplication is small and readable.
- `RealNetworkOps::check_inbound_reachable` is a near-exact copy of `check_inbound_port25` with only the URL path differing. A shared helper `curl_reachable(&self, path: &str)` would eliminate the duplication and keep the two paths from drifting. Tiny win, not required.

None of these are bugs or blockers.

---

### Alignment with PRD

Sprint 13 directly supports FR-38 (verify service with complementary `/reach` and `/probe` endpoints), FR-42 (`aimx preflight`), and FR-48 (`aimx verify`). The PR honours the documented semantic split between the two endpoints: `/reach` for pre-install TCP reachability, `/probe` for post-install EHLO validation. PTR check remains advisory (Warn on missing, Pass on present, never Fail), consistent with the owner decision captured in the sprint plan.

---

```
[Sprint 13] Preflight flow fix + PTR display ordering

S13.1: Route `aimx preflight` inbound check at the plain-TCP `/reach`
endpoint (added in Sprint 12) instead of `/probe`. Fresh-VPS preflight
no longer fails because nothing is listening on port 25 yet. `aimx
setup` (post-install) and `aimx verify` continue to use `/probe` for
the full EHLO handshake, so there is no regression in those flows.

- NetworkOps gains `check_inbound_reachable()` alongside
  `check_inbound_port25()`; RealNetworkOps mirrors the existing curl
  impl against `/reach` with the same 60s timeout and parse.
- New `check_inbound_tcp()` helper in setup.rs; `run_preflight` uses
  it, `run_setup` and `verify::run_with_net` keep `check_inbound`.

S13.2: Fix PTR display interleaving. `check_ptr` used to `println!`
the PTR value before returning `Pass`, clashing with the caller's
unflushed `print!("  PTR record... ")` header and producing mangled
output when inbound failed.

- `PreflightResult::Pass` now carries `Option<String>`; `check_ptr`
  returns `Pass(Some(ptr))` and `run_preflight` prints the value
  inline as `  PTR record... PASS (<ptr>)`.
- All match arms updated in src/setup.rs and src/verify.rs.
- `run_preflight` refactored into `run_preflight_to<W, E>` so output
  ordering is testable without touching global stdout.

Tests: 291 unit + 35 integration, all pass. New regression test
`preflight_no_ptr_interleaving_when_inbound_fails` asserts the exact
stream ordering that was broken. New mock `TrackingNetworkOps` pins
each caller to the right endpoint so silent future regressions break
loudly.
```

uzyn:
## Follow-up: addressed review nits (37b51c0)

Four non-blocking items from the APPROVED review — all addressed:

**1. Stale doc comment on `run_preflight_to<W, E>`** — fixed. Comment now accurately describes that `out` is for check progress and `err` is for failure advisories, and that tests supply separate buffers.

**2. Pre-existing `preflight_*` tests still called global `run_preflight`** — migrated to `run_preflight_to` with captured `Vec<u8>` buffers. This removes the stdout noise during `cargo test` and ensures every test-side caller of the preflight flow now goes through the testable entry point. The only remaining caller of `run_preflight` is the CLI-level `run_setup` path, which is the intended production binding.

**3. `check_inbound` / `check_inbound_tcp` duplication** — extracted a private `inbound_result(res: Result<bool, _>) -> PreflightResult` helper. Both wrappers are now one-liners that forward the respective trait method's result. Single source of truth for the "Inbound port 25 is not reachable" message.

**4. `RealNetworkOps::check_inbound_port25` / `check_inbound_reachable` duplication** — extracted a private `curl_reachable(&self, path: &str)` helper. Both trait methods now forward to it with just the path literal (`/probe` vs `/reach`). The curl invocation, timeout (60s), and response parser (`"reachable":true` / `"reachable": true`) live in exactly one place, eliminating drift risk.

### Nothing intentionally left as-is
All four flagged items were small enough to fix, so all four were addressed. No new concerns surfaced.

### Verification
- `cargo test`: 291 unit + 35 integration, all pass

Diff: +51 / -44 (net +7 lines, but adds a shared helper and removes two redundant bodies — the increase is comment/doc text).

uzyn (commented):
## Re-review (Cycle 2) — APPROVED

_(Self-approval blocked by GitHub; posting as comment with APPROVED verdict.)_

All four non-blocker nits from Cycle 1 are verified fixed against the actual code in commit `37b51c0`. I re-read the file rather than trusting the summary, re-ran `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the full test suite locally — all clean, **291 unit + 35 integration tests pass**.

### Resolved

**1. Stale doc comment on `run_preflight_to<W, E>`** — `src/setup.rs:432-434`. Rewritten to accurately describe the split: `out` receives check progress, `err` receives failure advisories, and tests supply separate buffers. No more misleading "both output streams are routed to `out`" claim.

**2. Pre-existing preflight tests still calling global `run_preflight`** — `preflight_all_pass` (1387), `preflight_fails_on_outbound_blocked` (1396), `preflight_fails_on_inbound_blocked` (1408), `preflight_passes_with_ptr_warning` (1422) all migrated to `run_preflight_to(&net, &mut out, &mut err)` with captured `Vec<u8>` buffers. I grepped the whole tree — the only remaining caller of `run_preflight` is `run_preflight_command` at `src/setup.rs:1128`, which is the intended CLI production binding. No test-side caller writes to real stdout anymore.

**3. `check_inbound` / `check_inbound_tcp` DRY** — `src/setup.rs:390-398` introduces private `fn inbound_result(res) -> PreflightResult`. Both wrappers are now one-liners: `check_inbound` forwards `net.check_inbound_port25()`, `check_inbound_tcp` forwards `net.check_inbound_reachable()`. Single source of truth for the `"Inbound port 25 is not reachable"` advisory and the `"Inbound port 25 check failed: {e}"` error path. Helper is correctly private.

**4. `RealNetworkOps::check_inbound_port25` / `check_inbound_reachable` DRY** — `src/setup.rs:219-236` introduces private `fn curl_reachable(&self, path: &str) -> Result<bool, Box<dyn std::error::Error>>`. Both trait methods (lines 294-300) are now one-liners forwarding `"/probe"` or `"/reach"`. The curl invocation (`-s -m 60`), the 60s timeout, and the dual-format body parser (`"reachable":true` with or without the space) all live in exactly one place. URL construction uses `format!("{}{path}", self.verify_host)` — semantically identical to the pre-refactor `format!("{}/probe", ...)` / `format!("{}/reach", ...)`. Drift risk eliminated.

### Still unresolved
None.

### New issues found
None. The refactor is tight — net +7 lines, and the growth is comment/doc text (three new doc comments on the helpers and the corrected `run_preflight_to` doc). No behavioral change, no new clippy warnings, no new test failures. Both extracted helpers are private and have no other call sites, so the blast radius is contained.

### Verdict
**APPROVED** — ship it. Sprint 13 is complete with zero blockers and all four polish items cleanly addressed.

```
[Sprint 13] Preflight flow fix + PTR display ordering

Route aimx preflight's inbound check at the new verify-service /reach
endpoint (plain-TCP reachability) so a fresh VPS passes the check
before OpenSMTPD is installed. aimx setup (post-install) and aimx
verify continue to use /probe (full EHLO handshake) since both run
after OpenSMTPD is listening.

Adds check_inbound_reachable() to the NetworkOps trait alongside the
existing check_inbound_port25(), and a check_inbound_tcp() wrapper
that run_preflight uses for the fresh-VPS path. A tracking mock
asserts each caller (preflight vs setup vs verify) lands on the
correct method.

Fixes the PTR display ordering bug: PreflightResult::Pass now carries
Option<String> for inline detail, check_ptr() returns Pass(Some(ptr))
instead of emitting its own println! from inside the check. The
caller prints a single well-formed line like
"  PTR record... PASS (<ptr>)". PTR stays advisory — missing PTR
still Warns and still passes overall preflight.

run_preflight is refactored into run_preflight_to<W, E> taking writer
sinks so output ordering can be asserted in unit tests without
touching global stdout. All pre-existing preflight tests migrated to
the testable entry point with captured buffers.

Internal DRY: shared inbound_result() helper collapses check_inbound
and check_inbound_tcp Fail/Err handling to a single source of truth;
shared RealNetworkOps::curl_reachable(path) helper collapses the
/probe and /reach curl call sites so the invocation, 60s timeout, and
response parser cannot drift between the two endpoints.

291 unit tests + 35 integration tests pass. fmt and clippy clean.
```

PR #57, opened 2026-04-15
[Sprint 30] Goose + OpenClaw Integration

uzyn:
## Sprint Goal

Add Goose (recipe-based) and OpenClaw (skill-based, JSON5 config) to `aimx agent-setup`, completing the v1 agent-integration roster.

### [S30-1] `agents/goose/aimx-recipe.yaml` + registry entry
- [x] Goose recipe authored (`agents/goose/aimx.yaml.header`): `title: "AIMX Email"`, `prompt: |` = indented common primer, `extensions:` = stdio entry for `aimx mcp`
- [x] `goose` registered in `agent_setup.rs` with destination `$XDG_CONFIG_HOME/goose/recipes` (defaults to `~/.config/goose/recipes/aimx.yaml`); `GOOSE_RECIPE_GITHUB_REPO` detected at install time and surfaced in the activation hint
- [x] Activation hint prints `goose run --recipe aimx`
- [x] Unit tests cover default path install, XDG override, `--data-dir` rewrite, and `GOOSE_RECIPE_GITHUB_REPO`-set activation hint
- [x] Recipe format verified against current Goose docs (linked in `agents/goose/README.md`)
- [ ] Manual end-to-end validation on a machine with Goose installed — **deferred**: sandbox lacks Goose CLI

**Source filename note:** the sprint's draft name `aimx-recipe.yaml` would produce `goose run --recipe aimx-recipe`, which does not match the AC-specified activation verb `goose run --recipe aimx`. I used `aimx.yaml.header` so the recipe installs as `aimx.yaml` and the `--recipe aimx` invocation resolves correctly.

### [S30-2] `agents/openclaw/` skill + registry entry
- [x] `agents/openclaw/SKILL.md.header` authored with valid OpenClaw frontmatter (`name: aimx`, `description`, `version`, single-line `metadata.openclaw` per OpenClaw's parser constraint); body = common primer
- [x] `agents/openclaw/README.md` documents the two-step activation: install skill, then run the printed `openclaw mcp set` command
- [x] `openclaw` registered in `agent_setup.rs`; activation hint prints the exact `openclaw mcp set aimx '<json>'` command
- [x] Unit tests cover install layout, activation hint stability, `--data-dir` JSON embedding, special-char escaping (round-trip parse)
- [x] OpenClaw skill format + `openclaw mcp set` CLI verified against current docs (linked in README)
- [ ] Manual end-to-end validation on a machine with OpenClaw installed — **deferred**: sandbox lacks OpenClaw

**Source layout note:** OpenClaw uses the flat skill-only source layout (SKILL.md.header at the top of `agents/openclaw/`) matching OpenCode / Gemini, since the destination `~/.openclaw/skills/aimx/` already terminates the skill-dir nesting.

### [S30-3] Final docs pass + README overhaul
- [x] `book/agent-integration.md` gains Goose and OpenClaw sections plus matrix rows and troubleshooting entries; all six v1 agents now documented
- [x] Top-level `README.md` shows a six-row table of supported agents + install commands (replaces the old bash snippet). The sprint AC said "five-row" — I believe that was a typo (PRD §6.10 FR-50 lists six v1 agents including OpenClaw, and omitting OpenClaw here would be inconsistent with the sprint's own S30-2 registration).
- [x] `grep -r "mcpServers" book/` returns only `book/agent-integration.md` (stale "paste this snippet" prose removed from `book/mcp.md` and `book/getting-started.md`; `docs/manual-setup.md` pointer updated too)
- [x] `aimx agent-setup --list` covers all six agents via the registry (test `registry_lists_six_agents_in_canonical_order` asserts stable order)

## Technical Decisions

- **Goose recipe assembly uses a new `indent_block` helper + `rewrite_recipe_data_dir`.** Rather than round-tripping the recipe through a YAML serializer (which would preserve nothing about the hand-authored format), I added a line-oriented primer-indent step in `assemble_plugin_files` and a targeted line-based injection for the `--data-dir` rewrite. The block-scalar `prompt: |` stays exactly as authored.
- **OpenClaw MCP wiring via `openclaw mcp set` CLI, not JSON5 editing.** OpenClaw's config file is JSON5 (comments, trailing commas). Rather than shipping a JSON5 parser for one config file, AIMX prints the exact `openclaw mcp set aimx '<json>'` shell command. This matches the "print the activation command" pattern already established for OpenCode and Gemini.
- **Goose destination is a directory, not a file.** Dest template `$XDG_CONFIG_HOME/goose/recipes` because the source ships a single top-level file (`aimx.yaml.header` → `aimx.yaml`) that lands inside the recipes directory. `write_files` already handles mkdir-ing the parent chain.
- **Single-line `metadata` in OpenClaw frontmatter.** OpenClaw's SKILL.md parser per docs requires single-line frontmatter keys; metadata must be a single-line JSON object. The authored header respects that.

## Deferred Items

- **Manual Goose validation** — deferred because the sandbox lacks Goose. Schema + CLI verb verified against `goose-docs.ai` + `mintlify.com/block/goose/api/cli/recipe` docs. Tracked for the non-blocking review backlog (same pattern as Sprint 28/29 deferrals).
- **Manual OpenClaw validation** — deferred because the sandbox lacks OpenClaw. Schema + `openclaw mcp set` CLI signature verified against `docs.openclaw.ai/cli/mcp`. Tracked for the non-blocking review backlog.

## Review Focus Areas

- `src/agent_setup.rs::assemble_plugin_files` — new `.yaml.header` transformation branch. Confirm the indent-then-data-dir-rewrite order is correct and that the YAML block scalar survives.
- `src/agent_setup.rs::rewrite_recipe_data_dir` — the line-based injection deliberately avoids a YAML round-trip. Worth checking the edge cases (args block with no list items, args nested under a different key).
- `agents/goose/aimx.yaml.header` vs. sprint-plan AC wording (`aimx-recipe.yaml`) — filename choice explained under S30-1 above.
- `agents/openclaw/SKILL.md.header` — the single-line `metadata: {"openclaw": ...}` intentional to respect OpenClaw's parser constraint; multi-line YAML mapping would be rejected.
- Six-row README table vs. sprint AC's "five-row" — justified above under S30-3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 **SMTP for agents. No middleman.**
 
-One command to give your AI agents their own email addresses -- no Gmail, no OAuth, no third-party SaaS. Built for Claude Code, Codex CLI, OpenCode, Gemini CLI, and any MCP-capable agent that needs an email channel.
+One command to give your AI agents their own email addresses -- no Gmail, no OAuth, no third-party SaaS. Built for Claude Code, Codex CLI, OpenCode, Gemini CLI, Goose, OpenClaw, and any MCP-capable agent that needs an email channel.
 
 ```bash
 aimx setup agent.mydomain.com
@@ -141,14 +141,16 @@ aimx send --from support@agent.yourdomain.com \
 aimx mcp
 ```
 
-Install AIMX into your agent with one command per agent:
+Install AIMX into your agent with one command:
 
-```bash
-aimx agent-setup claude-code    # Claude Code
-aimx agent-setup codex          # Codex CLI
-aimx agent-setup opencode       # OpenCode
-aimx agent-setup gemini         # Gemini CLI
-```
+| Agent | Install command | Activation |
+|-------|-----------------|------------|
+| Claude Code | `aimx agent-setup claude-code` | Restart Claude Code (auto-discovered from `~/.claude/plugins/`). |
+| Codex CLI | `aimx agent-setup codex` | Restart Codex CLI (auto-discovered from `~/.codex/plugins/`). |
+| OpenCode | `aimx agent-setup opencode` | Paste the printed JSONC block into `opencode.json`, then restart. |
+| Gemini CLI | `aimx agent-setup gemini` | Merge the printed JSON block into `~/.gemini/settings.json`, then restart. |
+| Goose | `aimx agent-setup goose` | Run `goose run --recipe aimx`. |
+| OpenClaw | `aimx agent-setup openclaw` | Run the printed `openclaw mcp set aimx '...'` command, then restart the gateway. |
 
 Run `aimx agent-setup --list` to see every supported agent and its
 destination path. See [`book/agent-integration.md`](book/agent-integration.md)
--- a/agents/goose/README.md
+++ b/agents/goose/README.md
@@ -1,0 +1,79 @@
+# AIMX recipe for Goose
+
+This directory is the source tree for the Goose recipe that wires AIMX
+into [Goose](https://goose-docs.ai/). Contents are bundled into the
+`aimx` binary at compile time (via `include_dir!`) and installed by
+`aimx agent-setup goose`.
+
+## What gets installed
+
+- `aimx.yaml` — a Goose recipe dropped into `~/.config/goose/recipes/aimx.yaml`.
+  The recipe bundles:
+  - `title` + `description` — recipe metadata Goose surfaces in `goose
+    recipe list`.
+  - `extensions` — a stdio entry for `aimx mcp` so running the recipe
+    automatically launches AIMX's MCP server.
+  - `prompt` — the canonical AIMX primer (`agents/common/aimx-primer.md`)
+    indented as a YAML block scalar so there is one source of truth for
+    agent-facing instructions.
+
+The installer assembles `aimx.yaml` from a YAML header plus the primer at
+install time; the raw `.header` file is not shipped.
+
+## Install
+
+```bash
+aimx agent-setup goose
+```
+
+Default recipe destination: `~/.config/goose/recipes/aimx.yaml`.
+
+## Activation
+
+Once the recipe is installed, run it with:
+
+```bash
+goose run --recipe aimx
+```
+
+Goose auto-discovers recipe files in `~/.config/goose/recipes/` and picks
+them up by the filename stem, so `aimx.yaml` is invoked as `--recipe
+aimx`.
+
+## Team / org-wide recipe sharing
+
+If you set the `GOOSE_RECIPE_GITHUB_REPO` environment variable to a
+GitHub repo path (e.g. `myorg/goose-recipes`), Goose loads recipes from
+that repo instead of (or in addition to) your local directory. In that
+case, `aimx agent-setup goose` still writes the recipe locally, and the
+activation hint tells you to commit `~/.config/goose/recipes/aimx.yaml`
+into your team repo so every user can invoke it.
+
+## Overriding the data directory
+
+If AIMX was set up with a non-default data directory, re-run the
+installer with `--data-dir`:
+
+```bash
+aimx --data-dir /custom/path agent-setup goose
+```
+
+The recipe's `extensions[0].args` will include `--data-dir /custom/path`
+before `mcp`.
+
+## Schema reference
+
+Goose recipe schema is documented at
+<https://goose-docs.ai/docs/guides/recipes>. The recipe format requires
+`title` and `description` plus at least one of `instructions` or
+`prompt`; extensions follow the `type: stdio` shape with `name`, `cmd`,
+and `args`.
+
+## Design choice: recipe-based integration
+
+Goose's native integration shape is different from the skills-based
+agents (Claude Code, Codex, OpenCode, Gemini). A recipe bundles both the
+MCP extension config AND the agent-facing instructions in one YAML
+file, so there is no separate "paste this JSON snippet" step — running
+the recipe starts AIMX's MCP server automatically. This matches FR-49:
+the installer writes one file and prints one activation command.
--- a/agents/goose/aimx.yaml.header
+++ b/agents/goose/aimx.yaml.header
@@ -1,0 +1,15 @@
+version: "1.0.0"
+title: "AIMX Email"
+description: >
+  Self-hosted email for AI agents. Send, read, reply to, and manage email
+  via AIMX MCP tools stored as Markdown files on the local filesystem.
+extensions:
+  - type: stdio
+    name: aimx
+    cmd: /usr/local/bin/aimx
+    args:
+      - mcp
+    timeout: 300
+    description: >
+      AIMX MCP server — nine tools for mailbox and email operations.
+prompt: |
--- a/agents/openclaw/README.md
+++ b/agents/openclaw/README.md
@@ -1,0 +1,74 @@
+# AIMX skill for OpenClaw
+
+This directory is the source tree for the OpenClaw skill that wires
+AIMX into [OpenClaw](https://docs.openclaw.ai/). Contents are bundled
+into the `aimx` binary at compile time (via `include_dir!`) and
+installed by `aimx agent-setup openclaw`.
+
+## What gets installed
+
+- `skills/aimx/SKILL.md` — an agent-facing skill dropped into
+  `~/.openclaw/skills/aimx/SKILL.md`. Its body is the canonical AIMX
+  primer (`agents/common/aimx-primer.md`); the installer assembles the
+  final `SKILL.md` from a YAML header plus that primer so there is one
+  source of truth.
+
+OpenClaw's MCP configuration lives in `~/.openclaw/openclaw.json` (a
+JSON5 file) under the `mcpServers` key. The installer does **not**
+mutate that file; instead it prints an `openclaw mcp set` command you
+paste into your shell to register AIMX's MCP server with one step.
+
+## Install
+
+```bash
+aimx agent-setup openclaw
+```
+
+Default skill destination: `~/.openclaw/skills/aimx/SKILL.md`.
+
+## Activation
+
+After the skill is installed, register AIMX's MCP server with OpenClaw's
+built-in CLI:
+
+```bash
+openclaw mcp set aimx '{"command":"/usr/local/bin/aimx","args":["mcp"]}'
+```
+
+That command edits `~/.openclaw/openclaw.json` for you — no config-file
+hand-editing. Restart the OpenClaw gateway after registration so the
+new server is loaded.
+
+## Overriding the data directory
+
+If AIMX was set up with a non-default data directory, re-run the
+installer with `--data-dir`:
+
+```bash
+aimx --data-dir /custom/path agent-setup openclaw
+```
+
+The printed `openclaw mcp set` command's JSON will include
+`--data-dir /custom/path` in the `args` array.
+
+## Schema reference
+
+OpenClaw's skill format is documented at
+<https://docs.openclaw.ai/tools/skills>. Skills use YAML frontmatter
+with `name` + `description` (required), optional `version`, and
+optional `metadata.openclaw` declarations for runtime requirements.
+The body is agent-facing instructions.
+
+OpenClaw's MCP server configuration lives in
+`~/.openclaw/openclaw.json` under `mcpServers.<name>` with `command` +
+`args` (stdio transport). The CLI reference for `openclaw mcp set` is
+documented at <https://docs.openclaw.ai/cli/mcp>.
+
+## Design choice: CLI-based activation
+
+OpenClaw provides a first-class `openclaw mcp set <name> <json>` CLI
+that registers an MCP server non-interactively. AIMX uses that command
+as the activation step rather than asking users to hand-edit
+`~/.openclaw/openclaw.json` (a JSON5 file AIMX does not want to parse
+and rewrite). This matches FR-49: the installer writes the skill to
+disk and prints one exact command the user runs.
--- a/agents/openclaw/SKILL.md.header
+++ b/agents/openclaw/SKILL.md.header
@@ -1,0 +1,7 @@
+---
+name: aimx
+description: Self-hosted email for AI agents. Send, read, reply to, and manage email via AIMX MCP tools. Invoke when the user asks about sending mail, checking an inbox, replying to a message, or managing mailboxes, or when the conversation references AIMX.
+version: 1.0.0
+metadata: {"openclaw": {"requires": {"bins": ["aimx"]}}}
+---
+
--- a/book/agent-integration.md
+++ b/book/agent-integration.md
@@ -52,8 +52,8 @@ grows as more agents are landed in subsequent sprints.
 | Codex CLI | `aimx agent-setup codex` | `~/.codex/plugins/aimx/` | Restart Codex CLI; the plugin is auto-discovered from `~/.codex/plugins/`. |
 | OpenCode | `aimx agent-setup opencode` | `~/.config/opencode/skills/aimx/` | Paste the printed JSONC block into `opencode.json`, then restart OpenCode. |
 | Gemini CLI | `aimx agent-setup gemini` | `~/.gemini/skills/aimx/` | Merge the printed JSON block into `~/.gemini/settings.json`, then restart Gemini CLI. |
-
-More agents — Goose and OpenClaw — land in Sprint 30.
+| Goose | `aimx agent-setup goose` | `~/.config/goose/recipes/aimx.yaml` | Run `goose run --recipe aimx`. The recipe bundles its own MCP extension, so no separate config step. |
+| OpenClaw | `aimx agent-setup openclaw` | `~/.openclaw/skills/aimx/` | Run the printed `openclaw mcp set aimx '...'` command, then restart the OpenClaw gateway. |
 
 ### Claude Code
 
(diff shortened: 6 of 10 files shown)


uzyn (commented):
# Sprint 30 Review — Goose + OpenClaw Integration

## Sprint Goal Assessment

The sprint goal — "Add Goose (recipe-based) and OpenClaw (skill-based, JSON5 config) to `aimx agent-setup`, completing the v1 agent-integration roster" — is achieved. Both agents are registered in the central registry, both install successfully into their canonical per-user locations under test, both print correct activation hints, and the v1 roster of six agents (PRD §6.10 FR-50) is now complete. The two deviations from the sprint plan (source filename `aimx.yaml.header` rather than `aimx-recipe.yaml`; six-row README table instead of "five-row") are well-justified and documented in the PR body — the first is required for the AC-specified activation verb to resolve, the second aligns with the PRD. No objection.

**[S30-1] `agents/goose/` + registry entry**
- [x] Recipe authored with `title`, `description`, `extensions` (stdio entry for `aimx mcp`), `prompt: |` block scalar — met (`agents/goose/aimx.yaml.header`).
- [x] `goose` registered; destination respects `XDG_CONFIG_HOME`; `GOOSE_RECIPE_GITHUB_REPO` surfaced in the activation hint — met. Note: the AC says destination "respects `GOOSE_RECIPE_GITHUB_REPO`" as a path-choice, but the implementer correctly interpreted it as informational (the file still writes locally; the hint tells the user to commit it to the repo). This is the right interpretation — `GOOSE_RECIPE_GITHUB_REPO` is a consumer-side env var that Goose reads, not a place you can `fs::write()` to.
- [x] Activation hint prints `goose run --recipe aimx` — met (`goose_hint`).
- [x] Unit tests cover default path install, XDG override, `--data-dir` rewrite, and `GOOSE_RECIPE_GITHUB_REPO`-set hint — met (five dedicated tests).
- [x] Recipe format verified against current Goose docs — met (linked in `agents/goose/README.md`).
- [ ] Manual E2E validation — explicitly deferred (sandbox lacks Goose CLI). Consistent with Sprint 28/29 deferral pattern; acceptable.

**[S30-2] `agents/openclaw/` skill + registry entry**
- [x] `SKILL.md` authored with valid OpenClaw frontmatter (`name`, `description`, `version`, single-line `metadata.openclaw`) — met.
- [x] `agents/openclaw/README.md` documents two-step activation — met.
- [x] `openclaw` registered; activation hint prints the exact `openclaw mcp set aimx '<json>'` command — met.
- [x] Unit tests: install layout + hint stability + `--data-dir` embedding + special-char escaping — met (four dedicated tests, including a round-trip JSON parse).
- [x] OpenClaw skill + `openclaw mcp set` CLI syntax verified against current docs — met.
- [ ] Manual E2E validation — explicitly deferred. Acceptable.

**[S30-3] Final docs pass + README overhaul**
- [x] `book/agent-integration.md` has sections for all six agents — met (Goose + OpenClaw sections added; matrix row extended).
- [x] README shows six-row table with install commands — met. Sprint AC said "five-row"; this is correctly identified as a sprint-plan slip and reconciled to the PRD's six-agent v1 roster.
- [x] `grep -r "mcpServers" book/` returns only `book/agent-integration.md` — verified: stale snippets removed from `book/mcp.md`, `book/getting-started.md`, and `docs/manual-setup.md`.
- [x] `aimx agent-setup --list` covers all six via registry — met (`registry_lists_six_agents_in_canonical_order` pins the order).

## Test Coverage

18 new tests (47 total agent_setup unit tests, up from 29). Coverage is strong for this sprint's surface area:

- Install layout, XDG override, `--data-dir` rewrite, hint stability, frontmatter format, and an especially valuable round-trip escape test for special characters in data-dir paths (`openclaw_activation_hint_escapes_special_chars_in_data_dir`).
- `rewrite_recipe_data_dir_injects_args_before_mcp` locks in the ordering invariant.
- `indent_block_preserves_blank_lines_without_trailing_whitespace` pins the blank-line behavior that matters for YAML block-scalar hygiene.

**Test gap (Non-blocker):** `rewrite_recipe_data_dir` has no test for the "abort injection" path — the `else if !trimmed.trim_start().is_empty() && !trimmed.starts_with(' ')` branch that detects leaving the `args` block without finding a list item. Today that branch is unreachable with the shipped header, but if the header is ever restructured with an `args: []` inline-list form, the function would silently fail to inject and the `--data-dir` flag would be lost without any indication. A test covering at minimum the "args: ends before a `- ` line" case (whether it intentionally no-ops or returns an error) would be worth adding.

## Potential Bugs

### 1. OpenClaw activation hint: single-quote-in-path breaks the printed shell command (Non-blocker)

`openclaw_hint` (src/agent_setup.rs:189-215) wraps the JSON body in single quotes when printing:

```
openclaw mcp set aimx '{snippet_text}'
```

`serde_json::to_string` does not escape `'` because `'` is not a JSON-special character. If `--data-dir` is ever passed a path containing `'` (e.g. `/home/user's data/aimx`), the printed command is not a valid single-quoted shell string — it terminates the quote prematurely and the `mcp set` invocation fails or parses garbage.

The existing `openclaw_activation_hint_escapes_special_chars_in_data_dir` test verifies `"`, `\` escaping inside the JSON itself (which is fine — that's about JSON well-formedness), but it does not cover the outer shell-quoting layer.

Classification: **Non-blocker**. Paths with `'` are unusual, and the failure mode is an obviously broken shell command that the user sees immediately — no silent corruption, no reachable crash in AIMX itself. Worth hardening in a follow-up. Two reasonable fixes: (a) detect `'` in the data-dir path and either reject it or wrap in double quotes with explicit escaping, (b) print the command with `$'...'` ANSI-C quoting, or (c) switch to a heredoc in the printed instructions.

### 2. `rewrite_recipe_data_dir` silently no-ops if the `args:` list has no items (Non-blocker)

`src/agent_setup.rs::rewrite_recipe_data_dir` (line 582) walks line-by-line looking for the first `- ` under `args:`. If the list is empty (`args: []` inline, or `args:` followed by a non-list continuation), no injection happens and the function returns the text unchanged — no error, no warning. Today's shipped header always has `- mcp`, so this is unreachable, but the function is `pub` and will be exercised by any future recipe author. A future header that switches to `args: [mcp]` inline notation would break `--data-dir` silently: the installer would succeed, the user would think `--data-dir /custom/path` took effect, and at runtime the MCP server would launch against the default data directory. Consider either returning `Result<String, String>` and erroring on "never injected," or switching to a YAML-aware parse for that one key.

Classification: **Non-blocker** — unreachable with shipped content; the unit test `install_goose_with_custom_data_dir_rewrites_recipe_args` guards against a shipped-header regression. File for future hardening.

### 3. Activation hint for Goose reads process env at registry-scan time (`print_registry` may leak `GOOSE_RECIPE_GITHUB_REPO` noise) (Non-blocker)

`print_registry` calls `(spec.activation_hint)(None)` for every registered agent so that `aimx agent-setup --list` can show an activation one-liner. For Goose, that invocation calls `std::env::var_os("GOOSE_RECIPE_GITHUB_REPO")` and, if set, appends a multi-line blurb about team-repo sharing to the `--list` output. The `--list` formatter already handles multi-line hints by prefixing continuation lines with a gutter indent, so nothing breaks — but the UX is slightly surprising: `aimx agent-setup --list` changes its output based on an environment variable the user may have set for completely different reasons. Minor. Non-blocker; not worth fixing unless the team cares about deterministic `--list` output.

## Security Issues

No new security exposure. The OpenClaw activation hint shells out nothing — it prints a command for the user to run. The Goose recipe format does not execute code except via the stdio extension (`/usr/local/bin/aimx mcp`), which the user would already be running by installing AIMX. No path traversal surface: `dest_template` is a static compile-time constant; `data_dir` comes through the same trait-mocked pipeline as the rest of `agent-setup` and reaches the YAML/JSON serializers only as a quoted scalar.

One observation (not an issue): the `rewrite_recipe_data_dir` function does use `serde_json::to_string(&dd)` to produce a YAML-safe double-quoted scalar. This works because YAML's double-quoted-scalar escape syntax is a proper superset of JSON's for `\"` and `\\`. It would NOT work for a path containing a literal tab or newline — JSON would emit `\t`/`\n`, which YAML also accepts, so actually that case is fine too. The escape reuse is defensible.

## Code Quality

- `indent_block` (src/agent_setup.rs:556) is tidy and the blank-line handling is correct. One readability note: the helper is tested exactly once, and the test only covers a single 3-line fixture. A second test for a multi-line input with a trailing-newline-free last line would flush out a sharp edge of `split_inclusive` usage. Non-blocker.
- `rewrite_recipe_data_dir` has a tightly scoped job with a clear state machine (`in_args` + `injected`). The comment at the top correctly explains why a full YAML round-trip is rejected (block-scalar preservation). Keep.
- Registry doc-comment at line 30-39 is stale: it still says "plugin-style agents (`claude-code`, `codex`) nest their skill under `skills/aimx/` because..." and enumerates only four agents. After this sprint, the source-tree layout has three shapes (plugin-with-skill, flat-skill, flat-recipe). Worth updating the doc-comment to mention the recipe shape and list all six agents. Non-blocker.

## Alignment with PRD

Sprint delivers the last two of six v1 agents specified by §6.10 FR-50. The PRD says each shipped package must contain "both the MCP wiring and an instructions payload describing AIMX's MCP tools, storage layout, frontmatter fields, and read/unread semantics so the agent can interact with AIMX without further prompting" — both new agents satisfy this: Goose carries the primer inline as the recipe's `prompt: |`, OpenClaw carries it as the `SKILL.md` body. FR-49 ("never mutates the agent's own primary config file") is honored — the implementer explicitly refuses to edit OpenClaw's JSON5 config and delegates to `openclaw mcp set`.

No P0 requirements unaddressed that this sprint should have covered.

## Human Review Comments

No human comments on PR #57 or on any linked GitHub issue (PR does not reference one).

**Blockers:** none.

**Non-blockers (follow-up, not blocking):**
- OpenClaw activation hint: single-quote-in-path breaks the printed shell command.
- `rewrite_recipe_data_dir` silently no-ops if `args:` list is empty — consider surfacing as an error.
- Stale `registry()` doc-comment (still mentions four agents and two source-tree shapes).
- Goose `--list` output varies with `GOOSE_RECIPE_GITHUB_REPO` env (cosmetic).

**Nice-to-haves:**
- Additional `rewrite_recipe_data_dir` test for the "args-ends-before-first-list-item" branch.
- Second `indent_block` test with a multi-line input missing a trailing newline.
- Manual E2E validation of both agents once a sandbox with each CLI is available (tracked deferral, same pattern as Sprints 28/29).

```
[Sprint 30] Goose + OpenClaw Integration

Complete the v1 agent-integration roster (PRD §6.10 FR-50) by adding
Goose and OpenClaw to `aimx agent-setup`. Goose installs a recipe
(`~/.config/goose/recipes/aimx.yaml`) that bundles the MCP extension
wiring with the AIMX primer as a YAML block scalar; activation is a
single `goose run --recipe aimx`. OpenClaw installs a SKILL.md and
prints an `openclaw mcp set aimx '...'` command so users wire MCP
with one pasted shell command without AIMX touching the JSON5 config.

Adds `indent_block` and `rewrite_recipe_data_dir` helpers in
`src/agent_setup.rs` to assemble the recipe line-by-line and inject
`--data-dir` without round-tripping the file through a YAML serializer
that would destroy the hand-authored `prompt: |` block scalar. README
is converted from a bash snippet to a six-row install table; stale
`mcpServers` copy-paste prose is purged from `book/mcp.md`,
`book/getting-started.md`, and `docs/manual-setup.md`. 18 new unit
tests bring the agent_setup suite to 47 tests; `cargo clippy` and
`cargo fmt --check` are clean. Manual Goose/OpenClaw E2E validation
deferred (sandbox lacks both CLIs) per the Sprint 28/29 pattern.
```

uzyn:
Addressed all four Cycle 1 non-blockers in 51e7b51.

**Fixed**

1. **OpenClaw single-quote shell injection** — `openclaw_hint` now POSIX-shell-escapes the JSON body with a new `posix_single_quote` helper (standard `'\''` concatenation trick) before wrapping it in single quotes. A `--data-dir` path containing `'` (e.g. `/home/user's data/aimx`) no longer terminates the quoted string prematurely. Regression test `openclaw_activation_hint_escapes_single_quote_in_data_dir` round-trips the printed command through a minimal POSIX unquoter and `serde_json::from_str` to verify the path survives byte-for-byte; an additional unit test pins the `posix_single_quote` helper's semantics directly.

2. **`rewrite_recipe_data_dir` silent no-op** — signature changed to `Result<String, String>`. Returns an explicit error when the expected `args:` block with a `- ` list item is not found, so a future recipe-header restructure (e.g. `args: []` inline) will fail loudly instead of silently dropping the `--data-dir` flag. Two new negative tests cover the \"`args:` followed by a sibling key\" case and the \"no `args:` block at all\" case. Caller (`assemble_plugin_files`) propagates with `?`.

3. **Stale `registry()` doc comment** — rewritten to enumerate all six v1 agents (claude-code, codex, opencode, gemini, goose, openclaw) and all three source-tree shapes (plugin-with-skill, flat-skill, flat-recipe). References PRD §6.10 FR-50 for the roster.

4. **Goose `--list` env dependency** — `goose_hint` no longer reads `GOOSE_RECIPE_GITHUB_REPO` at render time. The team-sharing blurb is emitted unconditionally and references the variable by name (`$GOOSE_RECIPE_GITHUB_REPO`), so `aimx agent-setup --list` output is now deterministic across developer shells. Updated the existing test (renamed to `goose_activation_hint_mentions_github_repo_variable`) to assert the hint is byte-identical with and without the env var set, and that no concrete repo slug leaks into the output.

**Intentionally left as-is**

- **Manual Goose and OpenClaw CLI E2E validation** — sandbox lacks both CLIs; deferred per the Sprint 28/29 precedent already disclosed in the PR description.
- **Second `indent_block` test (nice-to-have)** — the reviewer flagged this as a \"nice-to-have\" for a multi-line input missing a trailing newline, not a non-blocker. The current test already exercises the blank-line branch that is the sharper edge of `split_inclusive`; leaving this for a follow-up.

**Verification**

- 51 agent_setup unit tests pass (up from 47; +4 new tests for the fixes)
- `cargo test` suite: all pass
- `cargo clippy -- -D warnings`: clean (matches CI)

**Status:** DONE

uzyn (commented):
# Sprint 30 Re-Review (Cycle 2) — All Non-Blockers Resolved

Re-reviewed commit `51e7b51` against the four non-blockers flagged in Cycle 1. All four are genuinely fixed with appropriate test coverage. No new issues introduced.

## Verification of Cycle 1 Non-Blockers

**1. OpenClaw activation hint — single-quote-in-path breaks printed command — RESOLVED**

`openclaw_hint` now routes the JSON body through a new `posix_single_quote` helper before emission. The helper implements the canonical `'\''` concatenation trick (close, literal escaped quote, reopen) correctly — verified by reading the implementation at `src/agent_setup.rs:224-240`:

- Iterates chars; on `'` emits `'\''`, otherwise the char; wraps the result in outer single quotes.
- `openclaw_hint` at line 213 calls `posix_single_quote(&snippet_text)` and interpolates `{shell_quoted}` in the format string.

Test coverage is strong: `openclaw_activation_hint_escapes_single_quote_in_data_dir` constructs a `--data-dir` path containing a literal `'` (`/home/user's data/aimx`), renders the hint, extracts the quoted word from the printed command, runs it through a purpose-built POSIX unquoter (`shell_unquote_single`), then through `serde_json::from_str`, and asserts the decoded path is byte-for-byte identical. This is a real end-to-end round-trip — not a unit assertion on the helper alone. A second test `posix_single_quote_escapes_embedded_quote` pins the helper's semantics directly across three cases (embedded quote, no special chars, empty string).

**2. `rewrite_recipe_data_dir` silent no-op — RESOLVED**

Signature changed to `Result<String, String>`. The final `if !injected` check (line 651) returns an explicit `Err` describing the failure mode. The `?` in `assemble_plugin_files` (line 482) propagates the error up to the installer, which surfaces it to the user. The previous abort branch (`else if !trimmed.trim_start().is_empty() && !trimmed.starts_with(' ')`) is preserved and now correctly feeds into the `!injected` error path rather than silently returning the original text.

Two new negative tests cover both cases:
- `rewrite_recipe_data_dir_errors_when_args_block_has_no_list_item` — `args:` followed by a sibling key; exercises the abort-branch path and asserts `unwrap_err()`.
- `rewrite_recipe_data_dir_errors_when_no_args_block_present` — no `args:` block anywhere.

Traced the logic by hand: for `"...args:\nprompt: |\n..."`, `args:` sets `in_args = true`; next line `prompt: |` is non-empty and not indented, so the abort branch fires, `in_args = false`, `injected` stays false, final check returns `Err`. Correct.

The existing positive test was updated to `.unwrap()` the Result, preserving the original ordering assertion.

**3. Stale `registry()` doc-comment — RESOLVED**

Doc comment (lines 29-48) now enumerates all six v1 agents (`claude-code`, `codex`, `opencode`, `gemini`, `goose`, `openclaw`) with an explicit PRD §6.10 FR-50 anchor. The three source-tree shapes are named and explained:
- Plugin-with-skill: `claude-code`, `codex`
- Flat skill: `opencode`, `gemini`, `openclaw`
- Flat recipe: `goose`

Cross-checked against the actual registry body — every name and shape classification matches the corresponding `dest_template`.

**4. Goose `--list` output varies with `GOOSE_RECIPE_GITHUB_REPO` — RESOLVED**

`goose_hint` is now a static `&str` wrapped in `.to_string()` — no `std::env::var_os` call. The team-sharing blurb is emitted unconditionally and references `$GOOSE_RECIPE_GITHUB_REPO` by name so the user learns the mechanism without leaking their local shell env into `--list` output.

The test `goose_activation_hint_mentions_github_repo_variable` asserts:
- `hint_with == hint_without` across set/unset (byte-for-byte determinism)
- Hint mentions `GOOSE_RECIPE_GITHUB_REPO`
- Hint does NOT leak the concrete repo slug `myorg/goose-recipes`

This is a stronger test than the original — it locks down determinism by direct equality rather than just content checks.

## New Issues

None. The diff is surgical (one file, +188/-46), scoped to the four items, and doesn't touch unrelated code.

## Verification Run

- `cargo test agent_setup` — 51 passed, 0 failed (up from 47; +4 new tests as described)
- Integration tests suite — 44 passed

## Nice-to-haves Still Open (tracked, not blocking)

- Second `indent_block` test with multi-line input missing a trailing newline (Cycle 1 nice-to-have; intentionally deferred, acknowledged in implementer response).
- Manual E2E validation of Goose recipe and OpenClaw skill on machines with the respective CLIs — appropriately deferred per the Sprint 28/29 precedent.

## Verdict

**APPROVED — Ready to merge.** (Posted as a comment rather than an approval because the PR author is the reviewer's account; human approval can be self-administered.)

All Cycle 1 non-blockers are genuinely resolved with real test coverage. Test suite grew from 47 to 51 agent_setup tests. No regressions. No new issues.

```
[Sprint 30] Goose + OpenClaw Integration

Complete the v1 agent-integration roster (PRD §6.10 FR-50) by adding
Goose and OpenClaw to `aimx agent-setup`. Goose installs a recipe
(`~/.config/goose/recipes/aimx.yaml`) that bundles the MCP extension
wiring with the AIMX primer as a YAML block scalar; activation is a
single `goose run --recipe aimx`. OpenClaw installs a SKILL.md and
prints an `openclaw mcp set aimx '...'` command (POSIX-shell-escaped
so paths containing `'` survive intact) so users wire MCP with one
pasted shell command without AIMX touching the JSON5 config.

Adds `indent_block`, `posix_single_quote`, and `rewrite_recipe_data_dir`
helpers in `src/agent_setup.rs` to assemble the recipe line-by-line
and inject `--data-dir` without round-tripping the file through a YAML
serializer that would destroy the hand-authored `prompt: |` block
scalar. `rewrite_recipe_data_dir` returns `Result<String, String>` so
a future header restructure fails loudly rather than silently dropping
the `--data-dir` flag. Goose activation hint is deterministic — the
team-sharing blurb references `$GOOSE_RECIPE_GITHUB_REPO` by name
rather than reading the env var at render time, so `--list` output is
stable across developer shells. README is converted from a bash
snippet to a six-row install table; stale `mcpServers` copy-paste
prose is purged from `book/mcp.md`, `book/getting-started.md`, and
`docs/manual-setup.md`. 22 new unit tests bring the agent_setup suite
to 51 tests; `cargo clippy -- -D warnings` and `cargo fmt --check` are
clean. Manual Goose/OpenClaw E2E validation deferred (sandbox lacks
both CLIs) per the Sprint 28/29 pattern.
```

PR #64, opened 2026-04-15
[Sprint 34] UDS Wire Protocol + aimx serve Send Listener

uzyn:
## Summary

Stands up the daemon-side of the `aimx send` rewrite so that DKIM signing and outbound SMTP delivery live inside `aimx serve`, reachable from any local user over a world-writable UDS at `/run/aimx/send.sock`.

- **S34-1 `src/send_protocol.rs`** — pure `AsyncRead`/`AsyncWrite` codec for the `AIMX/1 SEND` frame (length-prefixed, binary-safe). Case-insensitive header names, unknown headers ignored for forward-compat, body read by exact `Content-Length`. Rejects wrong request-line, missing/duplicate required headers, bad or oversized `Content-Length`, truncated bodies, non-ASCII header names. Response codec renders `AIMX/1 OK <id>` or `AIMX/1 ERR {MAILBOX|DOMAIN|SIGN|DELIVERY|TEMP|MALFORMED} <reason>` with CR/LF stripped from the reason.
- **S34-2 `src/serve.rs`** — binds `tokio::net::UnixListener` at `send_socket_path()` (default `/run/aimx/send.sock`, overridable via `AIMX_RUNTIME_DIR`), sets mode `0o666`, handles stale-socket reclaim on `EADDRINUSE`, logs `SO_PEERCRED` for diagnostics only (never for authorization), and ties SIGTERM/SIGINT shutdown into the existing SMTP-listener drain path. Socket file is removed on clean shutdown.
- **S34-3 `src/send_handler.rs`** — per-connection `handle_send(req, &ctx)` backed by an `Arc<SendContext>` carrying the DKIM key loaded once at daemon startup. Mailbox-registration check → `From:` parse → case-insensitive domain compare → DKIM sign → delivery via the existing `LettreTransport` trait. Every error path maps to a stable `ErrCode`.

## Wire format

```
Client → Server:
  AIMX/1 SEND\n
  From-Mailbox: <name>\n
  Content-Length: <n>\n
  \n
  <n bytes of RFC 5322 message, unsigned>

Server → Client:
  AIMX/1 OK <message-id>\n
or
  AIMX/1 ERR <code> <reason>\n
```

## Scope notes

- Authorization is explicitly out of scope for v0.2. The socket is world-writable by design (PRD / Sprint 33.1); `peer_uid`/`peer_pid` are read and logged for journald diagnostics only.
- `aimx send` is still the old in-process signer — client rewrite lands in Sprint 35. This PR only proves the daemon-side with unit tests + raw-UDS integration tests.
- `setup_test_env` now mints a DKIM keypair inside each test's `AIMX_CONFIG_DIR` because `aimx serve` refuses to start without one. The `dkim_keygen_*` tests were updated to bypass the helper so they still start from a clean slate.

### S34-1 — `src/send_protocol.rs` wire-format codec
- [x] Module added; exports `SendRequest`, `SendResponse`, `ErrCode`, `parse_request`, `write_response`
- [x] `parse_request` reads `AIMX/1 SEND\n`, headers until blank line, `Content-Length` bytes exactly
- [x] Required headers: `From-Mailbox`, `Content-Length`; unknowns ignored
- [x] Rejections: wrong request-line, missing header, bad/oversized `Content-Length`, truncated body → `Malformed`
- [x] `write_response` emits correct `OK` / `ERR <CODE>` lines for all six codes
- [x] Round-trip unit tests for every response variant using in-memory async streams
- [x] Parser fuzzed: truncated body, oversized body, CRLF vs LF, case-insensitive names, duplicate headers, missing blank line, empty body, body containing `AIMX/1 SEND\n`
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` clean

### S34-2 — `aimx serve` binds `/run/aimx/send.sock`
- [x] `UnixListener` bound at `send_socket_path()`; `AIMX_RUNTIME_DIR` override implemented
- [x] Mode set to `0o666` after bind; owner left as running process UID
- [x] `SO_PEERCRED` read per accept and logged; never used for authz
- [x] Bind failure is fatal at startup with clear error naming path + errno
- [x] Stale-socket on-disk: unlink-and-retry once, fail loudly on second attempt
- [x] SIGTERM/SIGINT drain both the SMTP listener and the UDS accept loop; socket removed on clean shutdown
- [x] Unit test asserts mode `0o666` and reclaim-from-stale
- [x] Integration tests spawn `aimx serve` in a tempdir, open a raw UDS, assert accept + response
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` clean

### S34-3 — daemon-side send handler
- [x] `src/send_handler.rs` with `async fn handle_send(req, &ctx) -> SendResponse`
- [x] `SendContext` holds `Arc<DkimKey>`, primary domain, mailbox set, `Arc<dyn MailTransport>`
- [x] `serve.rs::main` loads DKIM once at startup; failure is fatal with a clear message
- [x] `From:` parsed; domain compare case-insensitive; any local part accepted
- [x] Domain mismatch → `ERR DOMAIN`; unknown `From-Mailbox` → `ERR MAILBOX`
- [x] DKIM signing reuses relaxed/relaxed canonicalization from Sprint 25; failure → `ERR SIGN`
- [x] Delivery via existing `LettreTransport`; transport errors classified into `Delivery` vs `Temp`
- [x] Response written via `send_protocol::write_response`
- [x] Each accept spawns its own `tokio::spawn` task
- [x] Unit tests mock `MailTransport` across every error path; integration test asserts DKIM-signed message hits the transport
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` clean

## Test plan

- [x] `cargo fmt -- --check`
- [x] `cargo test --bin aimx` — 483 pass
- [x] `cargo test --test integration` — 49 pass including the four new Sprint 34 integration tests
- [ ] Exercise `aimx serve` on a real host once Sprint 35 client lands (deferred — daemon-side verified via raw-UDS tests here)

## Recommended commit message

```
[Sprint 34] UDS Wire Protocol + `aimx serve` Send Listener (#PR)

Move DKIM signing and outbound SMTP delivery inside `aimx serve`, exposed
over a world-writable UDS at `/run/aimx/send.sock`. Adds the `AIMX/1
SEND` wire codec (`src/send_protocol.rs`), the daemon-side UDS listener
with stale-socket reclaim and SO_PEERCRED diagnostics (`src/serve.rs`),
and the per-connection handler that validates, DKIM-signs, and delivers
(`src/send_handler.rs`). `aimx send` remains unchanged in this sprint —
the thin-client rewrite lands in Sprint 35.

Authorization is out of scope in v0.2: the socket is 0o666 by design;
`peer_uid`/`peer_pid` are logged for journald diagnostics only.
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,6 +9,8 @@ mod mcp;
 mod mx;
 mod platform;
 mod send;
+mod send_handler;
+mod send_protocol;
 mod serve;
 mod setup;
 mod smtp;
--- a/src/send_handler.rs
+++ b/src/send_handler.rs
@@ -13,6 +13,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use rsa::RsaPrivateKey;
+use uuid::Uuid;
 
 use crate::dkim;
 use crate::send::MailTransport;
@@ -46,6 +47,20 @@ pub struct SendContext {
 /// DKIM-sign → deliver via MX. Every error path maps to a stable
 /// [`ErrCode`].
 pub async fn handle_send(req: SendRequest, ctx: &SendContext) -> SendResponse {
+    handle_send_with_signer(req, ctx, dkim::sign_message).await
+}
+
+/// Generic form of [`handle_send`] parameterized on the DKIM signer so tests
+/// can inject a failing signer without constructing a bad key. Production
+/// code always routes through [`handle_send`], which wires [`dkim::sign_message`].
+pub(crate) async fn handle_send_with_signer<F>(
+    req: SendRequest,
+    ctx: &SendContext,
+    signer: F,
+) -> SendResponse
+where
+    F: FnOnce(&[u8], &RsaPrivateKey, &str, &str) -> Result<Vec<u8>, Box<dyn std::error::Error>>,
+{
     if !ctx.registered_mailboxes.contains(&req.from_mailbox) {
         return SendResponse::Err {
             code: ErrCode::Mailbox,
@@ -53,12 +68,14 @@ pub async fn handle_send(req: SendRequest, ctx: &SendContext) -> SendResponse {
         };
     }
 
-    let (message_id, from_header) = match parse_from_and_message_id(&req.body) {
-        Ok(v) => v,
-        Err(reason) => {
+    let headers = scan_headers(&req.body, &["From", "To", "Message-ID"]);
+
+    let from_header = match headers.get("From") {
+        Some(v) => v.clone(),
+        None => {
             return SendResponse::Err {
                 code: ErrCode::Malformed,
-                reason,
+                reason: "missing required header: From".to_string(),
             };
         }
     };
@@ -84,8 +101,8 @@ pub async fn handle_send(req: SendRequest, ctx: &SendContext) -> SendResponse {
         };
     }
 
-    let to_header = match header_value(&req.body, "To") {
-        Some(v) => v,
+    let to_header = match headers.get("To") {
+        Some(v) => v.clone(),
         None => {
             return SendResponse::Err {
                 code: ErrCode::Malformed,
@@ -94,8 +111,38 @@ pub async fn handle_send(req: SendRequest, ctx: &SendContext) -> SendResponse {
         }
     };
 
-    let signed = match dkim::sign_message(
-        &req.body,
+    // The submitted To: header may carry a display-name (`"Name"
+    // <user@host>`), a bare addr (`user@host`), or even angle-only
+    // (`<user@host>`). `lettre::Address::FromStr` only parses the bare form,
+    // so normalize to `user@host` before handing it to the transport. Any
+    // failure to extract a bare recipient is MALFORMED — not a delivery
+    // error — because nothing has been attempted over the wire.
+    let recipient_bare = match extract_bare_address(&to_header) {
+        Some(addr) => addr,
+        None => {
+            return SendResponse::Err {
+                code: ErrCode::Malformed,
+                reason: format!("could not extract recipient address from To: {to_header}"),
+            };
+        }
+    };
+
+    // If Message-ID is absent we synthesize one ourselves rather than
+    // erroring out: the sprint's error table never listed Message-ID as a
+    // required client header, and `AIMX/1 OK <message-id>` still needs
+    // something to echo. Using the configured primary domain matches the
+    // DKIM `d=` tag and avoids leaking a recipient-side hostname.
+    let (message_id, body_bytes) = match headers.get("Message-ID") {
+        Some(v) => (v.clone(), req.body.clone()),
+        None => {
+            let synthetic = format!("<{}@{}>", Uuid::new_v4(), ctx.primary_domain);
+            let injected = inject_message_id_header(&req.body, &synthetic);
+            (synthetic, injected)
+        }
+    };
+
+    let signed = match signer(
+        &body_bytes,
         &ctx.dkim_key,
         &ctx.primary_domain,
         &ctx.dkim_selector,
@@ -109,7 +156,7 @@ pub async fn handle_send(req: SendRequest, ctx: &SendContext) -> SendResponse {
         }
     };
 
-    match ctx.transport.send(&from_header, &to_header, &signed) {
+    match ctx.transport.send(&from_header, &recipient_bare, &signed) {
         Ok(_server) => SendResponse::Ok { message_id },
         Err(e) => {
             let msg = e.to_string();
@@ -119,35 +166,57 @@ pub async fn handle_send(req: SendRequest, ctx: &SendContext) -> SendResponse {
     }
 }
 
-/// Extract the `From:` header value and the `Message-ID:` header value from
-/// an RFC 5322 message. Returns `Err(reason)` if either is missing.
-fn parse_from_and_message_id(body: &[u8]) -> Result<(String, String), String> {
-    let message_id = header_value(body, "Message-ID")
-        .ok_or_else(|| "missing required header: Message-ID".to_string())?;
-    let from =
-        header_value(body, "From").ok_or_else(|| "missing required header: From".to_string())?;
-    Ok((message_id, from))
+/// Insert a `Message-ID:` header at the top of an RFC 5322 message body. The
+/// body's existing line-endings (CRLF or LF) are preserved by reusing the
+/// same terminator the first header uses.
+fn inject_message_id_header(body: &[u8], message_id: &str) -> Vec<u8> {
+    let terminator: &[u8] = if body.windows(2).take(1024).any(|w| w == b"\r\n") {
+        b"\r\n"
+    } else {
+        b"\n"
+    };
+    let mut out = Vec::with_capacity(body.len() + 32 + message_id.len());
+    out.extend_from_slice(b"Message-ID: ");
+    out.extend_from_slice(message_id.as_bytes());
+    out.extend_from_slice(terminator);
+    out.extend_from_slice(body);
+    out
 }
 
-/// Look up a header by case-insensitive name in an RFC 5322 message. Header
-/// lines are assumed to be CRLF-separated (SMTP-ready), but this helper
-/// tolerates lone LFs so it also works on locally-composed buffers. Stops
-/// at the first blank line (headers/body separator). Joins continuation
-/// lines (leading WSP).
-fn header_value(body: &[u8], name: &str) -> Option<String> {
-    let text = std::str::from_utf8(body).ok()?;
-    let mut lines = text.lines();
-    let mut current: Option<String> = None;
-
-    let commit = |current: &Option<String>| -> Option<(String, String)> {
-        let line = current.as_ref()?;
-        let (n, v) = line.split_once(':')?;
-        Some((n.trim().to_string(), v.trim().to_string()))
+/// Single-pass walk over an RFC 5322 header block, returning the values for
+/// each of `names` (case-insensitive, continuation-line folded). The returned
+/// map keys match the original `names` argument casing so callers can look
+/// up using the literal name they asked for.
+fn scan_headers(body: &[u8], names: &[&str]) -> std::collections::HashMap<String, String> {
+    let mut out: std::collections::HashMap<String, String> = std::collections::HashMap::new();
+    let text = match std::str::from_utf8(body) {
+        Ok(t) => t,
+        Err(_) => return out,
     };
 
-    let target = name.to_ascii_lowercase();
+    let lowercased: Vec<(String, &str)> =
+        names.iter().map(|n| (n.to_ascii_lowercase(), *n)).collect();
 
-    for line in lines.by_ref() {
+    let mut current: Option<String> = None;
+
+    let commit = |current: &Option<String>,
+                  out: &mut std::collections::HashMap<String, String>,
+                  lowercased: &[(String, &str)]| {
+        let Some(line) = current.as_ref() else {
+            return;
+        };
+        let Some((n, v)) = line.split_once(':') else {
+            return;
+        };
+        let n_lower = n.trim().to_ascii_lowercase();
+        for (target_lower, original) in lowercased {
+            if n_lower == *target_lower && !out.contains_key(*original) {
+                out.insert((*original).to_string(), v.trim().to_string());
+            }
+        }
+    };
+
+    for line in text.lines() {
         if line.is_empty() {
             break;
         }
@@ -158,35 +227,44 @@ fn header_value(body: &[u8], name: &str) -> Option<String> {
             }
             continue;
         }
-        if let Some((n, v)) = commit(&current)
-            && n.eq_ignore_ascii_case(&target)
-        {
-            return Some(v);
-        }
+        commit(&current, &mut out, &lowercased);
         current = Some(line.to_string());
     }
-
-    if let Some((n, v)) = commit(&current)
-        && n.eq_ignore_ascii_case(&target)
-    {
-        return Some(v);
-    }
-    None
+    commit(&current, &mut out, &lowercased);
+    out
 }
 
 /// Extract the bare-addr domain from an RFC 5322 `From:` header, handling
 /// both `"Name <user@host>"` and `"user@host"` forms. Returns `None` if no
 /// `@` is present.
 fn extract_domain(from: &str) -> Option<String> {
-    let addr = if let Some(start) = from.rfind('<') {
-        let tail = &from[start + 1..];
+    let addr = extract_bare_address(from)?;
+    let at = addr.rfind('@')?;
+    Some(addr[at + 1..].trim().to_string())
+}
+
+/// Extract the bare `local@host` form from a header value, accepting
+/// `"Name" <local@host>`, `local@host`, and angle-only `<local@host>`. For
+/// comma-separated header values only the first recipient is returned —
+/// v0.2 submissions are single-recipient and the daemon's envelope already
+/// only takes one address.
+fn extract_bare_address(value: &str) -> Option<String> {
+    let first = value.split(',').next().unwrap_or(value).trim();
+    if first.is_empty() {
+        return None;
+    }
+    let addr = if let Some(start) = first.rfind('<') {
+        let tail = &first[start + 1..];
         let end = tail.find('>').unwrap_or(tail.len());
         &tail[..end]
     } else {
-        from
+        first
     };
-    let at = addr.rfind('@')?;
-    Some(addr[at + 1..].trim().to_string())
+    let addr = addr.trim();
+    if addr.is_empty() || !addr.contains('@') {
+        return None;
+    }
+    Some(addr.to_string())
 }
 
 /// Map a transport error string to an `ErrCode`.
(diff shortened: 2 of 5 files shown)


uzyn (commented):
## Sprint Goal Assessment

The PR delivers the daemon-side of the `aimx send` rewrite as scoped: a new `AIMX/1 SEND` wire codec, a UDS listener bound inside `aimx serve`, and a per-connection handler that validates + DKIM-signs + delivers. All three acceptance-criteria checklists are materially satisfied, the unit/integration coverage is solid, and `cargo fmt` / `cargo clippy --all-targets -- -D warnings` / `cargo build --all-targets` all come back clean against this branch. I can reproduce the test counts cited in the PR description.

Scope matches the sprint plan — `aimx send` client is deliberately untouched and deferred to Sprint 35. One note: the sprint spec's S34-3 acceptance criterion asks the integration test to *cryptographically* verify the DKIM signature against the test public key; the PR stops at asserting `DKIM-Signature:` is present and contains `d=example.com` / `s=dkim`, deferring the crypto roundtrip to the existing `dkim.rs` test. That's a minor divergence from the criterion wording — flagged below.

**S34-1 — `src/send_protocol.rs` wire-format codec**
- [x] Module exports `SendRequest`, `SendResponse`, `ErrCode`, `parse_request`, `write_response` (plus `write_request` for Sprint 35).
- [x] `parse_request` reads request-line → headers → blank line → `Content-Length` bytes.
- [x] Required headers enforced; unknowns ignored.
- [x] All documented rejection paths covered by unit tests (`wrong_request_line_is_malformed`, `truncated_body_is_malformed`, etc.).
- [x] All six `ERR` codes exercised by `write_all_err_codes`.
- [x] CR/LF stripped from response reason via `sanitize_inline` — `response_reason_crlf_stripped` verifies.
- [x] Fuzz-style coverage: CRLF/LF, case-insensitivity, duplicates, empty body, body containing request-line literal (`body_containing_request_line_literal_is_not_reparsed`).

**S34-2 — `aimx serve` binds `/run/aimx/send.sock`**
- [x] `UnixListener` bound at `send_socket_path()` with `AIMX_RUNTIME_DIR` override.
- [x] Mode `0o666` set after bind; `bind_send_socket_sets_world_writable_mode` asserts it.
- [x] `SO_PEERCRED` logged on each accept (stderr line: `[send] accepted: peer_uid=… peer_pid=…`). Never read into an authorization decision.
- [x] Bind failure fatal and labeled; stale-socket unlink-and-retry verified by `bind_send_socket_reclaims_stale_path`.
- [x] SIGTERM/SIGINT drains both listeners; `uds_send_listener_cleaned_up_after_sigterm` asserts the socket file is unlinked.
- [x] Integration tests spawn `aimx serve` in a tempdir and drive the UDS end-to-end (`serve_creates_send_socket_with_world_writable_mode`, `uds_send_listener_accepts_and_rejects_domain_mismatch`, `uds_send_listener_rejects_malformed_request`).

**S34-3 — daemon-side send handler**
- [x] `handle_send(req, &ctx)` with `SendContext` shape matching the spec.
- [x] DKIM key loaded once at startup; fatal on failure with a clear message naming the path.
- [x] `From:` parsed; domain compare case-insensitive (`domain_match_is_case_insensitive`); any local part accepted.
- [x] Domain mismatch → `ERR DOMAIN` (test + integration).
- [x] Unknown `From-Mailbox` → `ERR MAILBOX` (unit test).
- [x] DKIM signing failure → `ERR SIGN` (path exists — see *Test Coverage* below).
- [x] Transport error classified into `Delivery` / `Temp` (`transport_permanent_error_maps_to_delivery`, `transport_unreachable_maps_to_temp`).
- [x] Each accept is a fresh `tokio::spawn`.
- [ ] **PARTIAL:** integration test asserts DKIM headers are present and contain `d=example.com` / `s=dkim`, but does *not* cryptographically verify the signature against the test public key as the acceptance criterion asks. The `uds_end_to_end_signed_delivery` test comment explicitly punts this to `dkim.rs::sign_and_verify_roundtrip`. That's a defensible call but a literal reading of the criterion is not met.

## Potential Bugs

### Non-blocker 1 — No read timeout on UDS connections (slow-loris exposure)
**Where:** `serve.rs::handle_send_connection` → `send_protocol::parse_request` → `reader.read_exact(&mut body)`.

Any local process can `connect()` the world-writable socket, send the request line + headers announcing `Content-Length: 26214400`, and then stall. The spawned handler task blocks on `read_exact` indefinitely. With no concurrency cap (spec explicitly defers that) and no per-read timeout, a malicious local user can accumulate arbitrary parked tasks and 25 MB/task of pre-allocated `Vec<u8>`. The sprint spec acknowledges the concurrency cap is a follow-up ("defense-in-depth concurrency cap can be a follow-up if review flags it") — flagging it here. A short per-read timeout on the parser is a cheaper fix than a full semaphore.

Non-blocker because (a) the socket is world-writable *by design* in v0.2 per PRD/Sprint 33.1, (b) the spec allowed this as a follow-up, and (c) no crash/corruption — just resource pressure.

### Non-blocker 2 — `LettreTransport::send` parses the full `To:` header as a `lettre::Address`
**Where:** `send_handler.rs` line 87 passes `to_header` (the raw header value) to `transport.send`; `send.rs::LettreTransport::send` line 130 calls `recipient.parse::<lettre::Address>()` on it.

`lettre::Address`'s `FromStr` implementation parses the bare `local@host` form. If a submitter sends `To: Name <user@example.com>` or `To: a@x.com, b@y.com`, the parse fails and the entire delivery maps to `ERR DELIVERY` / `ERR TEMP` depending on the error string. The current in-tree `aimx send` composer writes bare-form `To:` so this is latent, but the socket is world-writable — any local process is now a potential client, and the natural user expectation is that a well-formed RFC 5322 `To:` header works. Recommend stripping to the bare address before handing to the transport (reuse `send.rs::extract_domain`'s adjacent parse logic, or introduce a shared helper).

Related: the handler only looks at `To:` for the envelope — `Cc:` and `Bcc:` headers (if the submitter composes them) are silently dropped from the envelope. Consistent with Sprint 34 scope (client doesn't support them yet) and the spec, but worth documenting on the wire.

### Non-blocker 3 — `ERR MALFORMED` used for a missing `Message-ID` header
**Where:** `send_handler.rs::parse_from_and_message_id` — missing `Message-ID` returns a `Malformed` response.

The sprint spec's error table says "`From:` missing or malformed → `Malformed`" — it does *not* list `Message-ID` as a required header with its own failure mode. The handler silently elevates `Message-ID` to a required header and reports `Malformed`. In practice the in-tree composer always produces one, so this is theoretical — but it's a small divergence from the spec and the `OK <message-id>` response needs *something* to echo, so the handler couldn't skip this check. Either accept the divergence and document it, or synthesize a message-ID on the server side when missing.

### Non-blocker 4 — Transport-error classification is brittle string matching
**Where:** `send_handler.rs::classify_transport_error`.

The classifier pattern-matches on lowercased substrings of the `Box<dyn Error>` message emitted by `LettreTransport::try_deliver` (`"Connection refused"`, `"timed out"`, `"unreachable"`, etc.). This works today because `send.rs::try_deliver` formats those strings itself — but any future refactor of the transport's error messages will silently re-classify errors. A typed error (even a thin enum on `LettreTransport::send`'s return) would be more durable. Non-blocker; worth tracking.

### Non-blocker 5 — Parse-failure drain has a 50ms idle-timeout, not a hard drain
**Where:** `serve.rs::handle_send_connection` lines 281-292.

The explicit ECONNRESET-avoidance drain loop waits up to 50ms between bytes and exits on the first idle window. If a misbehaving client sends a burst, pauses for 60ms, then sends more, the drain exits early and the subsequent write-side shutdown can still race with the client's read. Acceptable for the intended failure mode (well-behaved clients that closed their write-half), and the fallback is "client sees ECONNRESET, which is what would have happened pre-patch" — so the worst case is no regression. Leaving as-is is fine; flagging so the choice is explicit.

## Security Issues

### Low — `SO_PEERCRED` logged at daemon stderr
**Where:** `serve.rs::handle_send_connection` — `eprintln!("[send] accepted: peer_uid=… peer_pid=…")`.

Per the sprint design this is intentional (diagnostics via journald), but the daemon also runs interactively in tests and a dev workflow. Every accept leaves a peer-UID line in stderr. On a multi-tenant host this exposes which local UIDs are submitting mail at what time, which is a weak side-channel. The PRD explicitly scoped this as "diagnostic only, not for authz" so it's within scope — flagging the privacy consideration for future sprints where log redaction or rate-limiting may be wanted.

## Test Coverage

- **`ERR SIGN` branch has no test.** Every other `ErrCode` variant is exercised by a dedicated handler test except `Sign`. The branch is reached only if `dkim::sign_message` fails, which is hard to fake — but injecting a deliberately malformed key (e.g., a `RsaPrivateKey` with mismatched components, or mocking the sign path) would close the gap. Not a blocker since the path is trivial.
- **`uds_end_to_end_signed_delivery` does not cryptographically verify the signature** against the test public key, despite S34-3's acceptance criterion explicitly asking for this. The reader pem is loaded but only a textual "`PUBLIC KEY`" substring check is performed. Either wire the existing `sign_and_verify_roundtrip` helper from `dkim.rs` into this test, or update the sprint acceptance wording to acknowledge the split.
- **No concurrency test.** S34-3 mentions "one slow delivery doesn't block other sends" — the implementation (`tokio::spawn` per accept) clearly meets this, but a test with a blocking transport + a parallel transport would make the contract explicit. Non-blocker.
- **`setup_test_env` now re-shells `aimx dkim-keygen`** on every test invocation that spawns `serve`. Each test pays ~200ms+ for 2048-bit RSA key generation. Worth caching once per test binary (via `once_cell` + a static `TempDir`) if wall-clock CI time matters — the PR-cited 49-test integration run now generates many keypairs. Non-blocker, QoL.

## Code Quality

- `send_handler.rs::header_value` re-parses the body text on every call (once for `Message-ID`, once for `From`, once for `To`). Not a hot path (one request per connection, bounded by `Content-Length`), but parsing the same headers three times is wasteful when a single walk could produce all three. Non-blocker.
- `extract_domain` uses `rfind('<')` + `find('>')` — correct for well-formed input but a `mailparse`-style address parser would be more durable. Given the rest of the codebase uses hand-rolled parsers consistently, this is fine.
- The `parse_failed` plumbing in `handle_send_connection` is a reasonable compromise but the comment explains the trade-off well. Keeping the explanatory comment is the right call here.

**Blockers:** none. All acceptance criteria are materially met; the code compiles clean under `clippy --all-targets -- -D warnings` and the advertised test counts reproduce. No crashes, security vulns, or silent data loss were identified.

**Non-blockers (should fix or address before landing but won't block merge):**
- S34-3 integration test does not cryptographically verify the DKIM signature — either add the roundtrip to `uds_end_to_end_signed_delivery` or reconcile the acceptance criterion wording.
- `ERR SIGN` branch is untested; add one mock-signer path.
- Handler passes raw `To:` header (possibly display-name form, possibly comma-separated) straight into `lettre::Address::FromStr`; normalize to bare addr before calling `transport.send`.
- Missing `Message-ID` is classified as `Malformed` — not listed in the sprint's error table. Either synthesize on the server or document the extra required header.

- Per-read timeout inside `parse_request` to bound the slow-loris window on a world-writable socket, ahead of a full concurrency cap in a follow-up sprint.
- Replace string-matching in `classify_transport_error` with a typed error surface on `MailTransport`.
- Cache the DKIM keypair across integration tests instead of regenerating per test.
- Single-pass `header_value` that returns all needed headers at once.

uzyn:
## Review feedback addressed (commit e9ddafb)

Thanks for the thorough review. Addressed the four non-blockers plus two of the nice-to-haves.

### Fixed

1. **DKIM signature now cryptographically verified in the S34-3 integration test** (`src/serve.rs::uds_end_to_end_signed_delivery`). Added `verify_dkim_with_pubkey` helper that parses the test `public.key` PEM, builds a synthetic `v=DKIM1; k=rsa; p=...` record, and feeds it to `mail-auth`'s `MessageAuthenticator::verify_dkim` through an in-memory `ResolverCache<String, Txt>`. No DNS is performed. `DkimResult::Pass` is asserted — test fails if the signature does not verify against the test keypair.

2. **`ERR SIGN` branch now exercised** (`src/send_handler.rs::sign_failure_returns_sign_error`). Factored `handle_send` through a `handle_send_with_signer<F>` seam that takes an `FnOnce` signer. Production `handle_send` wires `dkim::sign_message` unchanged; the new test passes a closure that always returns an error and asserts `ErrCode::Sign` + reason propagation.

3. **`To:` header normalized before `lettre::Address::FromStr`**. New `extract_bare_address` helper handles `"Name" <user@host>`, bare `user@host`, angle-only `<user@host>`, and comma-separated lists (first recipient wins). The transport receives a bare address so display-name submissions no longer misclassify as `ERR DELIVERY`/`ERR TEMP`. New test `to_header_with_display_name_delivers_bare_address`.

4. **Missing `Message-ID` synthesized, not rejected**. When the submitted body lacks `Message-ID:`, the handler generates `<{uuid}@{primary_domain}>` (matches the DKIM `d=` tag), injects it at the top of the body before signing, and echoes it in `AIMX/1 OK <message-id>`. No more `ERR MALFORMED` for this case — the sprint error table now matches reality. New test `missing_message_id_is_synthesized` asserts both the response form and that the signed message carries the synthesized header.

### Nice-to-haves addressed

- **Single-pass `header_value`**: replaced by `scan_headers(body, &["From", "To", "Message-ID"])` which walks once and populates a HashMap.

### Intentionally left as-is

- **Per-read timeout inside `parse_request`**: the read loop uses `read` (single byte) and `read_exact` (sized body). A clean idle-timeout would need `tokio::time::timeout` around each awaited future, with careful handling of the partial-body case. Non-trivial for a strict non-blocker on a v0.2 world-writable socket; tracking for the concurrency-cap follow-up.
- **Typed transport errors**: agreed, but touches `MailTransport` trait and `LettreTransport::try_deliver`; out of scope for this sprint.
- **`once_cell` DKIM-keypair cache across integration tests**: DKIM-keygen has a fixed per-test cost (~200ms) and the integration suite is only 49 tests; sharing state across tests via `once_cell` + `static TempDir` introduces test-isolation concerns (tests already mutate `AIMX_CONFIG_DIR`). The cost was not prohibitive in the local run (≈70s integration). Left as tracked QoL.
- **SO_PEERCRED stderr logging as a privacy concern**: intentionally kept per spec (FR-18b diagnostic-only). Will revisit if a future sprint introduces log redaction policy.

### Verification

- `cargo test`: 491 pass (up from 486; +5 new tests), 3 ignored, 0 failed.
- Integration tests: 49 pass (unchanged), 0 failed.
- `cargo fmt --check`: clean.
- Verifier crate (`services/verifier/`): fmt and clippy clean.

uzyn (commented):
## Re-Review (Cycle 2)

All four previously flagged non-blockers from Cycle 1 are properly addressed in commit `e9ddafb`. Nothing regressed; `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the targeted tests all pass locally.

### Previously flagged issues — verification

**Resolved: S34-3 integration test now cryptographically verifies the DKIM signature**
`verify_dkim_with_pubkey` in `src/serve.rs` parses the test `public.key`, re-encodes as SPKI DER, wraps in a synthetic `v=DKIM1; k=rsa; p=...` TXT record, and seeds an in-memory `ResolverCache<String, Txt>` at `{selector}._domainkey.{domain}.` (trailing dot matches `IntoFqdn` normalization). The test asserts `DkimResult::Pass` on at least one output and panics with the full result vector otherwise. `uds_end_to_end_signed_delivery` passes locally in ~15s. No DNS performed.

**Resolved: `ERR SIGN` branch is tested**
Clean `handle_send_with_signer<F: FnOnce(…) -> Result<Vec<u8>, …>>` seam was introduced without touching the production signer wiring — `handle_send` still calls `dkim::sign_message` unchanged. `sign_failure_returns_sign_error` injects a closure that returns `Err("simulated DKIM signing failure")` and asserts both `ErrCode::Sign` and the reason string propagation.

**Resolved: `To:` header is normalized to the bare address before lettre**
New `extract_bare_address` helper correctly handles: bare (`user@host`), angle-only (`<user@host>`), display-name (`"Name" <user@host>`), and first-of-list (`a@x.com, b@y.com`). `@`-missing and empty inputs return `None`. The handler calls this before `transport.send`, so display-name `To:` headers no longer misclassify as `ERR DELIVERY`/`ERR TEMP`. Covered by `to_header_with_display_name_delivers_bare_address` plus 5 dedicated unit tests for the helper.

**Resolved: Missing `Message-ID` now synthesized, not `ERR MALFORMED`**
When `scan_headers` doesn't return a `Message-ID`, the handler generates `<{uuid}@{primary_domain}>` (domain matches DKIM `d=` tag), prepends it via `inject_message_id_header`, signs the injected body, and echoes it in `AIMX/1 OK <message-id>`. Line-ending detection is sensible (samples first 1024 bytes for CRLF, else LF). `missing_message_id_is_synthesized` asserts both the response contract and that the signed message carries the synthesized header.

**Bonus resolved: single-pass `scan_headers`**
Replaces the three `header_value` walks with one pass producing a `HashMap`. Correctly handles case-insensitive lookup, continuation-line folding, empty-line termination, and first-occurrence-wins. Five unit tests cover the semantics.

### New issues introduced

None. No regressions in existing tests; the handler's request-validation order is preserved (mailbox → From → domain → To → Message-ID → sign → deliver).

### Residual nice-to-haves (still tracked, not blocking)

The Cycle 1 nice-to-haves the implementer explicitly deferred remain reasonable to defer:
- Per-read timeout in `parse_request` (world-writable socket is by design in v0.2; concurrency cap is a tracked follow-up).
- Typed `MailTransport` errors (touches cross-module trait surface, correctly scoped out).
- DKIM keypair caching across integration tests (~70s suite is acceptable; `once_cell` sharing would introduce isolation concerns).
- `SO_PEERCRED` stderr log privacy (spec-mandated diagnostic).

### Verification

- `cargo test send_handler`: 23 pass, 0 fail.
- `cargo test serve::tests::uds_end_to_end_signed_delivery`: pass.

## Summary

All Cycle 1 non-blockers are resolved with clean, minimal changes. The crypto-roundtrip test is a meaningful upgrade over the prior substring check — the signed message is now verified against the exact public key material the daemon generated. The `handle_send_with_signer` seam is a nice testing hook without compromising the production path, and `extract_bare_address` + `inject_message_id_header` both handle the edge cases their tests cover.

```
[Sprint 34] UDS wire protocol + aimx serve send listener

Adds the daemon-side of the `aimx send` rewrite: a UDS listener at
/run/aimx/send.sock (root:root 0666) speaking AIMX/1 SEND. Per-
connection handler validates From-Mailbox, normalizes To: to a bare
address, synthesizes Message-ID if absent, DKIM-signs with the key
loaded once at startup, and delivers via the existing lettre
transport. Errors map to stable codes: MAILBOX, DOMAIN, MALFORMED,
SIGN, DELIVERY, TEMP.

Three new modules: src/send_protocol.rs (wire codec), extended
src/serve.rs (listener + SIGTERM drain), src/send_handler.rs
(business logic with handle_send_with_signer seam for testing).
Integration test drives the UDS end-to-end and cryptographically
verifies the DKIM signature via mail-auth against an in-memory TXT
cache seeded from the test public key — no DNS. Authorization is
out of scope for v0.2; the socket is world-writable by design and
SO_PEERCRED is logged diagnostically only.

aimx send client is untouched and lands in Sprint 35.
```

PR #66, opened 2026-04-15
[Sprint 36] Datadir Reshape — Inbox/Sent Split + Slug + Bundles

uzyn:
## Summary

Sprint 36 ships the final v0.2 on-disk layout: inbound mail under
`<data_dir>/inbox/<mailbox>/`, sent copies under
`<data_dir>/sent/<mailbox>/`, deterministic
`YYYY-MM-DD-HHMMSS-<slug>.md` UTC filenames, and Zola-style attachment
bundles. Mailbox lifecycle (`create`/`list`/`delete`) now manages both
inbox and sent directories together. MCP `email_*` tools gain an
optional `folder` parameter so agents can read the sent side.

## Stories

* **S36-1: Slug algorithm + filename helper** — new `src/slug.rs` with
  pure `slugify()` (FR-13b: lowercase → non-alphanumerics to `-` →
  collapse → trim → 20-char cap → `no-subject` fallback) and
  `allocate_filename()` that picks the final on-disk path (flat vs
  bundle) and resolves collisions with `-2`, `-3`, … suffixes. 12
  unit tests cover ASCII, Unicode, all-punctuation, truncation,
  collapsing, retrim, numeric preservation, exact UTC timestamp
  formatting, and every collision flavour.
* **S36-2: Ingest rewrite** — `ingest.rs` now writes to
  `<data_dir>/inbox/<mailbox>/`, calls `allocate_filename`, and
  produces flat `<stem>.md` for zero-attachment emails or a bundle
  `<stem>/<stem>.md` plus sibling attachments otherwise. The legacy
  per-mailbox `attachments/` directory is gone; per-attachment `path`
  is now bundle-relative. Channel rules see the inner `.md` in
  `{filepath}`.
* **S36-3: Mailbox lifecycle** — `mailbox_create` provisions both
  inbox and sent directories atomically (cleanup on second-create
  failure); `mailbox_delete` removes both; `catchall` cannot be
  deleted (error mentions `catchall`). MCP `email_list`,
  `email_read`, `email_mark_read`, `email_mark_unread` gain an
  optional `folder: "inbox" | "sent"` parameter (default `"inbox"`),
  with a clean error for unknown values. `email_send` and
  `email_reply` are unaffected.

## Test plan

- [x] `cargo test` — 527 unit tests + 51 integration tests, all
  passing (including `channel_recipe_end_to_end_with_templated_args`,
  the Sprint 31 channel-recipe smoke test, updated for the new
  filename shape)
- [x] Bundle layout exercised end-to-end via the existing
  `serve_e2e_*_attachment*` integration tests (no attachments → flat
  `.md`; attachments → bundle directory with siblings)
- [x] MCP `folder` parameter exercised by new unit tests for
  `set_read_status` against both inbox and sent dirs, plus
  `resolve_folder` validation tests

## Documentation

- `book/mailboxes.md` — layout diagram, bundle vs flat explanation,
  inbox/sent semantics, slug-based ID format
- `book/channels.md` — `{filepath}` example updated; bundle path
  behaviour explained
- `book/mcp.md` — every affected tool documents the new optional
  `folder` parameter
- `CLAUDE.md` — updated "Config and storage" section

## Recommended commit message

```
[Sprint 36] Datadir Reshape — Inbox/Sent Split + Slug + Bundles

Reshape the on-disk storage layout to match v0.2 (FR-13b, FR-14, FR-15):

* Inbound mail now lives under `<data_dir>/inbox/<mailbox>/`; outbound
  copies will land under `<data_dir>/sent/<mailbox>/` (Sprint 38). The
  per-day `YYYY-MM-DD-NNN.md` counter is replaced with deterministic
  `YYYY-MM-DD-HHMMSS-<slug>.md` UTC filenames.
* Add `src/slug.rs` with `slugify` (lowercase → non-alphanumerics to `-`
  → collapse → trim → 20-char cap → `no-subject` fallback) and
  `allocate_filename` (chooses flat `<stem>.md` vs Zola-style bundle
  `<stem>/<stem>.md`, suffixes `-2`/`-3`/… on collision).
* Zero attachments → flat `.md`; one or more → bundle directory with
  attachments as siblings. The mailbox-level `attachments/` subdir is
  removed entirely; per-attachment `path` is now bundle-relative.
* `mailbox create` provisions both `inbox/<name>/` and `sent/<name>/`
  atomically (cleanup on second-create failure); `mailbox delete`
  removes both. `catchall` cannot be deleted.
* MCP tools `email_list`, `email_read`, `email_mark_read`, and
  `email_mark_unread` gain an optional `folder: "inbox" | "sent"`
  parameter (default `"inbox"`); invalid values return a clean error.
* `email_list`, `set_read_status`, status recent-activity, and the
  mailbox count helpers all walk both flat `.md` files and bundle
  directories (`<stem>/<stem>.md`).
* Channel rules' `{filepath}` now expands to the `.md` inside the
  bundle when attachments are present.
* Update `book/mailboxes.md`, `book/channels.md`, `book/mcp.md`, and
  `CLAUDE.md` for the new layout, slug filename format, and folder
  parameter.
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -77,7 +77,7 @@ These are NOT a Cargo workspace — they have independent `Cargo.toml` files and
 
 - Config: `/etc/aimx/config.toml` (mode `0640`, owner `root:root`) — parsed via `serde` + `toml` into `Config` struct in `config.rs`. Resolved via `config_path()` which honors `AIMX_CONFIG_DIR` for tests / non-standard installs.
 - DKIM keys: `/etc/aimx/dkim/{private,public}.key` (private `0600` root-only, public `0644`). Loaded via `load_private_key(&config::dkim_dir())`.
-- Storage: Markdown files with TOML frontmatter (`+++` delimiters, not `---`). One `.md` file per email, `YYYY-MM-DD-NNN.md` naming, under `/var/lib/aimx/<mailbox>/`.
+- Storage: Markdown files with TOML frontmatter (`+++` delimiters, not `---`). Inbound mail lives at `/var/lib/aimx/inbox/<mailbox>/`, outbound sent copies at `/var/lib/aimx/sent/<mailbox>/`. Filenames use `YYYY-MM-DD-HHMMSS-<slug>.md` (UTC). Zero attachments produce a flat `<stem>.md`; one or more attachments produce a Zola-style bundle directory `<stem>/` containing `<stem>.md` plus each attachment as a sibling. The previous mailbox-level `attachments/` directory is gone (Sprint 36).
 - `--data-dir` / `AIMX_DATA_DIR` overrides the **storage** path (mailboxes only, v0.2). `AIMX_CONFIG_DIR` overrides the **config + DKIM** path.
 - Runtime dir: `/run/aimx/` (mode `0755`, owner `root:root`) — provided by systemd `RuntimeDirectory=aimx` or OpenRC `checkpath` in `start_pre`. Sprint 34 places `send.sock` here as a world-writable UDS (`0o666`); authorization is out of scope for v0.2.
 - Tests use `tempfile::TempDir` plus `crate::config::test_env::ConfigDirOverride::set(&tmp)` (or the `AIMX_CONFIG_DIR` env var in integration tests) to isolate config + DKIM lookups.
--- a/book/channels.md
+++ b/book/channels.md
@@ -55,15 +55,22 @@ The following variables are available in `command` strings:
 
 | Variable | Description | Example |
 |----------|-------------|---------|
-| `{filepath}` | Full path to the saved `.md` file | `/var/lib/aimx/support/2025-01-15-001.md` |
+| `{filepath}` | Full path to the saved `.md` file | `/var/lib/aimx/inbox/support/2025-01-15-103000-meeting.md` |
 | `{from}` | Sender email address | `alice@example.com` |
 | `{to}` | Recipient email address | `support@agent.yourdomain.com` |
 | `{subject}` | Email subject | `Meeting next Thursday` |
 | `{mailbox}` | Mailbox name | `support` |
-| `{id}` | Email ID | `2025-01-15-001` |
+| `{id}` | Email ID (filename stem) | `2025-01-15-103000-meeting` |
 | `{date}` | Email date | `2025-01-15T10:30:00Z` |
 
 Variable values are shell-escaped to prevent injection.
+
+When an email has attachments, `{filepath}` points at the `.md` file
+*inside* the Zola-style bundle directory
+(`/var/lib/aimx/inbox/<mailbox>/<stem>/<stem>.md`). Attachments live beside
+the `.md` as siblings inside the same directory; read the `attachments`
+array in the frontmatter to list them (each `path` is relative to the
+bundle directory).
 
 ## Match filters
 
--- a/book/mailboxes.md
+++ b/book/mailboxes.md
@@ -4,9 +4,25 @@ Mailboxes are the core organizational unit in AIMX. Each mailbox maps an email a
 
 ## Concepts
 
-- **Mailboxes are directories.** Creating a mailbox creates a folder and registers an address. No OS users, no passwords, no database.
-- **Catchall.** The `catchall` mailbox (created by default during setup) receives email for any unrecognized address at your domain.
+- **Mailboxes are directories.** Creating a mailbox creates two folders (one under `inbox/` and one under `sent/`) and registers an address. No OS users, no passwords, no database.
+- **Catchall.** The `catchall` mailbox (created by default during setup) receives email for any unrecognized address at your domain. Catchall is inbox-only — no `sent/catchall/` directory is created.
 - **No restart required.** Mailbox changes take effect immediately -- `aimx ingest` reads the config on each invocation.
+
+### On-disk layout
+
+```
+/var/lib/aimx/
+├── inbox/              # inbound mail lives here
+│   ├── catchall/
+│   └── support/
+└── sent/               # outbound sent copies (populated in a future release)
+    └── support/
+```
+
+Each email is stored as either a flat `YYYY-MM-DD-HHMMSS-<slug>.md` file
+when it has zero attachments, or as a Zola-style bundle directory
+`YYYY-MM-DD-HHMMSS-<slug>/` containing `<stem>.md` plus every attachment
+as a sibling file when attachments are present.
 
 ### Routing logic
 
@@ -25,7 +41,10 @@ For example, with mailboxes `support` and `catchall` configured:
 aimx mailbox create support
 ```
 
-This creates `support@agent.yourdomain.com` and the directory `/var/lib/aimx/support/`.
+This creates `support@agent.yourdomain.com` and both directories:
+`/var/lib/aimx/inbox/support/` (for incoming mail) and
+`/var/lib/aimx/sent/support/` (for outbound copies). Deletion removes
+both; `catchall` cannot be deleted.
 
 ### List mailboxes
 
@@ -75,7 +94,7 @@ This format is designed to be agent-readable without parsing libraries. An agent
 
 | Field | Type | Description |
 |-------|------|-------------|
-| `id` | string | Unique email ID within the mailbox (e.g. `2025-01-15-001`) |
+| `id` | string | Unique email ID within the mailbox (the filename stem, e.g. `2025-01-15-103000-meeting`) |
 | `message_id` | string | RFC 5322 Message-ID header |
 | `from` | string | Sender address with optional display name |
 | `to` | string | Recipient address |
@@ -97,12 +116,14 @@ This format is designed to be agent-readable without parsing libraries. An agent
 
 ## Attachments
 
-Attachments are extracted to an `attachments/` subdirectory within the mailbox folder:
+When an email carries one or more attachments, AIMX writes a Zola-style
+bundle directory whose name matches the `.md` file's stem:
 
 ```
-/var/lib/aimx/support/
-├── 2025-01-15-001.md
-└── attachments/
+/var/lib/aimx/inbox/support/
+├── 2025-01-15-103000-status-update.md         # flat: no attachments
+└── 2025-01-15-104500-quarterly-report/        # bundle: one or more attachments
+    ├── 2025-01-15-104500-quarterly-report.md
     ├── report.pdf
     └── image.png
 ```
@@ -114,15 +135,15 @@ Attachment metadata is stored in the email frontmatter:
 filename = "report.pdf"
 content_type = "application/pdf"
 size = 45230
-path = "attachments/report.pdf"
+path = "report.pdf"
 ```
 
 | Field | Description |
 |-------|-------------|
-| `filename` | Original filename |
+| `filename` | Original filename (path components stripped, duplicates suffixed `-1`, `-2`, …) |
 | `content_type` | MIME type |
 | `size` | File size in bytes |
-| `path` | Relative path from the mailbox directory |
+| `path` | Path relative to the bundle directory (no `attachments/` prefix in v0.2) |
 
 ## Read/unread tracking
 
@@ -177,7 +198,12 @@ The `email_reply` MCP tool handles threading automatically -- it reads the origi
 
 ## Email ID format
 
-Each email receives a unique ID within its mailbox in the format `YYYY-MM-DD-NNN` (e.g. `2025-01-15-001`). The counter increments atomically per mailbox to prevent collisions.
+Each email's `id` field is the filename stem
+`YYYY-MM-DD-HHMMSS-<slug>` in UTC. The slug is derived from the subject:
+lowercase, non-alphanumeric runs collapsed to `-`, trimmed, capped at 20
+characters, with `no-subject` as a fallback for empty results. Two emails
+with the same subject in the same UTC second have `-2`, `-3`, … appended
+to disambiguate.
 
 ---
 
--- a/book/mcp.md
+++ b/book/mcp.md
@@ -77,6 +77,7 @@ List emails in a mailbox with optional filters.
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
 | `mailbox` | string | yes | Mailbox name to list emails from |
+| `folder` | string | no | `"inbox"` (default) or `"sent"` — picks which side of the mailbox to list |
 | `unread` | bool | no | Filter to only unread emails |
 | `from` | string | no | Filter by sender address (substring match) |
 | `since` | string | no | Filter to emails since this datetime (RFC 3339 format) |
@@ -93,7 +94,8 @@ Read the full content of an email.
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
 | `mailbox` | string | yes | Mailbox name |
-| `id` | string | yes | Email ID (e.g. `2025-01-15-001`) |
+| `id` | string | yes | Email ID, i.e. the filename stem (e.g. `2025-01-15-103000-meeting`) |
+| `folder` | string | no | `"inbox"` (default) or `"sent"` |
 
 **Returns:** Complete `.md` file content including frontmatter and body.
 
@@ -136,7 +138,8 @@ Mark an email as read.
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
 | `mailbox` | string | yes | Mailbox name |
-| `id` | string | yes | Email ID (e.g. `2025-01-15-001`) |
+| `id` | string | yes | Email ID (filename stem, e.g. `2025-01-15-103000-meeting`) |
+| `folder` | string | no | `"inbox"` (default) or `"sent"` |
 
 Updates `read = true` in the email's frontmatter.
 
@@ -149,7 +152,8 @@ Mark an email as unread.
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
 | `mailbox` | string | yes | Mailbox name |
-| `id` | string | yes | Email ID (e.g. `2025-01-15-001`) |
+| `id` | string | yes | Email ID (filename stem, e.g. `2025-01-15-103000-meeting`) |
+| `folder` | string | no | `"inbox"` (default) or `"sent"` |
 
 Updates `read = false` in the email's frontmatter.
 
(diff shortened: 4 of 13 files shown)


uzyn (commented):
## Sprint Goal Assessment

Sprint 36 lands what it set out to land. Inbound mail now lives under `inbox/<mailbox>/`, deterministic `YYYY-MM-DD-HHMMSS-<slug>.md` filenames replace the per-day counter, attachments are bundled Zola-style, mailbox lifecycle (`create`/`list`/`delete`) manages both `inbox/` and `sent/`, and the `email_*` MCP tools accept an optional `folder` parameter. The sent-side population is correctly deferred to Sprint 38. The slug algorithm matches FR-13b precisely.

**S36-1: Slug algorithm + filename helper**
- [x] `src/slug.rs` with `slugify()` and `allocate_filename()`
- [x] Unit tests cover ASCII, Unicode, all-non-alphanumeric → `no-subject`, truncation, dash collapsing, leading/trailing trim
- [x] Collision tests: base / `-2` / `-3`, plus bundle vs flat collision check
- [x] Timestamp asserted exactly — `format_timestamp_utc_exact_string` covers `2025-03-14-150926`
- [x] `cargo test`/`clippy`/`fmt` clean

**S36-2: Ingest rewrite**
- [x] Writes to `<data_dir>/inbox/<mailbox>/`, unknown local parts → `inbox/catchall/`
- [x] Zero-attachment → flat `<stem>.md`; one-or-more → bundle `<stem>/<stem>.md` plus siblings
- [x] No top-level `attachments/` (verified by `attachment_creates_bundle_directory`)
- [x] `{filepath}` channel substitution updated; `book/channels.md` documents it
- [x] `channel_recipe_end_to_end_with_templated_args` updated and passing
- [x] Existing ingest integration tests updated for the new layout

**S36-3: Mailbox lifecycle**
- [x] `mailbox_create` creates both `inbox/<name>/` and `sent/<name>/` atomically with cleanup on second-failure
- [~] `mailbox_list` — PARTIAL: the spec says "scans `inbox/*/`" but `list_mailboxes` still iterates `config.mailboxes.keys()` and reads counts from `inbox/<name>/`. Functionally equivalent for the catchall surfacing requirement (catchall is always config-registered), and matches v1 behavior, but a stray inbox dir not in config will be invisible. Flag as a Non-blocker — it's a deviation in mechanism, not in observable outcome for the documented config-driven workflow.
- [x] `mailbox_delete` removes both; refuses catchall with an error mentioning `catchall`
- [x] MCP `folder: Option<String>` added to `email_list`/`email_read`/`email_mark_read`/`email_mark_unread`
- [x] Invalid folder value returns clean error (`resolve_folder_rejects_unknown` covers it)
- [x] Bundle counting in `count_messages` (verified by `list_counts_bundle_directories`)
- [x] Docs updated (`book/mailboxes.md`, `book/mcp.md`)

## Test Coverage

527 unit + 51 integration tests pass. Coverage is solid:
- Slug edge cases are well-covered (12 unit tests including Unicode, truncation re-trim, all-non-alphanumeric).
- Bundle vs flat collision is tested in both directions (`allocate_filename_bundle_collision_uses_directory_name`, `allocate_filename_flat_collides_with_existing_bundle`).
- MCP `folder=sent` round-tripped via `set_read_status_reads_sent_folder_when_requested`.

Gaps worth noting (none blocking):
- No test for **concurrent ingest** of the same subject within the same UTC second with the same attachment filenames — see Potential Bugs below.
- `email_list` against `folder=sent` is exercised indirectly via `set_read_status_reads_sent_folder_when_requested`, but there's no direct MCP-level test that calls `email_list` with `folder: "sent"` end-to-end. Low value because the path is the same `list_emails` helper.
- No integration test asserts that creating a mailbox actually leaves a `sent/<name>/` directory on disk via the CLI subprocess (only via the unit-level `create_mailbox` call). Optional.

## Potential Bugs

**1. Concurrent-ingest race on bundle directories — Non-blocker**

`src/ingest.rs:120-138`. `allocate_filename` decides on `<dir>/<stem>/<stem>.md` for a bundle layout; the caller then `create_dir_all(&parent_dir)` (idempotent), `write_attachments(&parent_dir, …)`, and finally `write_markdown` with `OpenOptions::create_new(true)`.

Two ingests racing on the same UTC second + subject + attachment-presence will both pick the same `<stem>` (because neither has created the bundle dir at the time of the check). Both will write attachments into the shared bundle dir; the second will then fail noisily on `create_new(true)` for the `.md` — but its already-written attachments remain in the first ingest's bundle, potentially clobbering same-named attachments from the first ingest. The `inspect_err` cleanup only fires on `write_attachments` failure, not on `write_markdown` failure, so there's no rollback for the second ingest's stray attachment writes.

Why Non-blocker:
- Requires same-second + same subject + same attachment filenames, which is rare in practice for the v0.2 single-host SMTP daemon.
- The PRD already mandates a `Mutex<()>` for the outbound side in FR-19b but does not require it for inbound.
- The ingest still produces a valid `.md` for the first arrival; the corruption surface is only the attachments inside the shared bundle.

Worth either: (a) adding an inbound `Mutex<()>` around the allocate-and-create critical section symmetric to the planned outbound one, or (b) a follow-up that does cleanup on `write_markdown` failure too.

**2. `mailbox_list` deviation from "scan `inbox/*/`" wording — Non-blocker**

`src/mailbox.rs:66-77` and `src/mcp.rs:424-438` both iterate `config.mailboxes.keys()` instead of scanning `inbox/*/`. For the documented workflow (mailboxes are always created via `aimx mailbox create`, which registers them in config), this is functionally identical. But:
- An inbox dir created out-of-band (e.g. by a backup restore) won't appear in `mailbox list`.
- A mailbox unregistered from config but with messages still on disk won't appear either.

Acceptable as a Non-blocker since v1 behaved the same way and the spec accepts "either is acceptable" for catchall surfacing. Worth a quick note in the sprint retrospective if you intended a real filesystem scan.

## Security Issues

None new in this PR. The pre-existing `validate_email_id` and `validate_mailbox_name` guards still cover path traversal for the new layouts (verified by `set_read_status_rejects_path_traversal`, `create_path_traversal_fails`, `create_with_slash_fails`, `create_with_backslash_fails`). Attachment filenames are sanitized via `Path::new(&raw_name).file_name()` in `prepare_attachments`, stripping smuggled path components. The `slugify` algorithm is total (returns `no-subject` on empty) and bounded (20 chars), so a hostile subject can't drive unbounded allocation or path-component injection.

## Code Quality

Generally clean. Minor observations (all Non-blocker, all optional):
- `src/ingest.rs:103` — `extract_local_part` already exists; no change. `prepare_attachments` and `write_attachments` cleanly separate I/O from MIME parsing, which made the bundle pivot tractable. Good shape for Sprint 37.
- `src/slug.rs:82` — `for suffix in 1..` with no upper bound; `unreachable!("collision loop exhausted u32::MAX suffixes")` is fine because `i32::MAX` collisions per UTC second is truly absurd, but a silent infinite loop on a filesystem that always returns "exists" for any path (very pathological FUSE behavior) would hang the ingest. Bounding to a sane number like 10_000 with a hard error would be defensive but unnecessary in practice.
- `src/ingest.rs:140-144` — `id` is derived from `md_path.file_stem()`, which for a bundle is `<stem>` (because `file_stem("foo/foo.md") == "foo"`). Correct and matches the documented `id` semantics, but worth a one-line comment because the single-file vs bundle equivalence isn't immediately obvious.

## Alignment with PRD

Implementation matches FR-9 (catchall inbox-only), FR-13b (slug + collision), FR-14 (bundle layout, no `attachments/` subdir), FR-15 (`inbox/<mailbox>/` routing). FR-19b/c (sent-items + outbound block + mutex) are correctly deferred to Sprint 38 per the sprint plan.

**Overall verdict:** Needs minor fixes (only the concurrent-ingest race is worth a brief response — the rest are observational).

Actually re-reading my own notes: the race is rare, doesn't cause panic or crash, and doesn't violate any acceptance criterion. It's strictly a Non-blocker by the rubric (no panic, no incorrect-result on the dominant path, the sprint doesn't claim concurrent-ingest atomicity). Downgrading to **Ready to merge with non-blocker observations**.

**Blockers:** none

**Non-blockers (worth tracking):**
- Concurrent same-second + same-subject + same-attachment-filename ingest can corrupt attachment contents in the first ingest's bundle. Either add an inbound `Mutex<()>` parallel to the planned outbound one, or extend the `inspect_err` cleanup to cover `write_markdown` failure.
- `mailbox_list` iterates `config.mailboxes` rather than scanning `inbox/*/` — confirm this matches your intent before closing the sprint.

**Nice-to-haves:**
- Direct MCP integration test for `email_list` with `folder=sent`.
- One-line comment on the `id` derivation in `ingest.rs` explaining why `file_stem()` works for both flat and bundle layouts.
- Bound the collision-suffix loop with a defensive ceiling.

```
[Sprint 36] Datadir Reshape — Inbox/Sent Split + Slug + Bundles

Reshape the on-disk storage layout to match v0.2 (FR-13b, FR-14, FR-15):

* Inbound mail now lives under `<data_dir>/inbox/<mailbox>/`; outbound
  copies will land under `<data_dir>/sent/<mailbox>/` (Sprint 38). The
  per-day `YYYY-MM-DD-NNN.md` counter is replaced with deterministic
  `YYYY-MM-DD-HHMMSS-<slug>.md` UTC filenames.
* Add `src/slug.rs` with `slugify` (lowercase → non-alphanumerics to `-`
  → collapse → trim → 20-char cap → `no-subject` fallback) and
  `allocate_filename` (chooses flat `<stem>.md` vs Zola-style bundle
  `<stem>/<stem>.md`, suffixes `-2`/`-3`/… on collision).
* Zero attachments → flat `.md`; one or more → bundle directory with
  attachments as siblings. The mailbox-level `attachments/` subdir is
  removed entirely; per-attachment `path` is now bundle-relative.
* `mailbox create` provisions both `inbox/<name>/` and `sent/<name>/`
  atomically (cleanup on second-create failure); `mailbox delete`
  removes both. `catchall` cannot be deleted.
* MCP tools `email_list`, `email_read`, `email_mark_read`, and
  `email_mark_unread` gain an optional `folder: "inbox" | "sent"`
  parameter (default `"inbox"`); invalid values return a clean error.
* `email_list`, `set_read_status`, status recent-activity, and the
  mailbox count helpers all walk both flat `.md` files and bundle
  directories (`<stem>/<stem>.md`).
* Channel rules' `{filepath}` now expands to the `.md` inside the
  bundle when attachments are present.
* Update `book/mailboxes.md`, `book/channels.md`, `book/mcp.md`, and
  `CLAUDE.md` for the new layout, slug filename format, and folder
  parameter.
```

uzyn:
Addressed both non-blocker review comments in 2bc4765.

**1. Concurrent ingest race on shared bundle directories**

Added a process-scoped `static INGEST_WRITE_LOCK: Mutex<()>` in `src/ingest.rs` guarding the critical section: `allocate_filename` → `create_dir_all(&parent_dir)` → `write_attachments` → `write_markdown`. Took the Mutex route (your preferred option) since it's symmetric to the planned outbound `Mutex<()>` (FR-19b) and the daemon is the single writer to `<data_dir>/inbox/`, so process-scoped locking is sufficient — no need for fs-level locks.

Also extended the bundle-cleanup `inspect_err` to fire on `write_markdown` failure as well as `write_attachments` failure (defensive — with the mutex held, `create_new` can no longer collide on the dominant path).

New test `concurrent_same_subject_attachment_ingests_do_not_corrupt_each_other` runs 8 threads racing the same subject + same attachment filename and asserts each ingest owns a distinct bundle directory with the original attachment bytes intact.

**2. `mailbox_list` iterates `config.mailboxes` rather than scanning `inbox/*/`**

`list_mailboxes` (CLI) and `list_mailboxes_with_unread` (MCP) now use a new `discover_mailbox_names` helper that scans `<data_dir>/inbox/*/` and unions with `config.mailboxes` keys. Stray inbox directories surface in the listing with an `(unregistered)` marker:

- CLI: appends a yellow `(unregistered)` after the message count.
- MCP: appends ` (unregistered)` to the existing per-line summary.

Catchall continues to surface as a normal registered entry (always in config). Status/serve sites still use `config.mailboxes.keys()` directly because they need the registered-mailbox metadata (SMTP routing acceptance, status table) — the filesystem-scan semantics are listing-specific.

Tests added:
- `list_surfaces_stray_inbox_dir_without_config_entry` in `mailbox.rs`: registered + catchall + stray dir all listed; `is_registered` correctly distinguishes them.
- `list_mailboxes_with_unread_surfaces_unregistered_dir` in `mcp.rs`: same coverage for the MCP-side variant; existing test expanded to assert the `registered` flag.

**Verification**

- `cargo test`: 531 unit + 51 integration pass (added 2 new unit tests).
- Verifier crate (`services/verifier/`) clippy clean.

Did not adopt the optional nice-to-haves (collision-suffix loop bound, MCP `email_list folder=sent` direct integration test, comment on `id` derivation) — happy to fold those in if you'd like.

uzyn (commented):
## Re-Review (Cycle 2)

Both Cycle-1 non-blockers were addressed in `2bc4765` and verified independently.

### Resolved

**1. Concurrent same-second / same-subject / same-attachment ingest race — RESOLVED**

`src/ingest.rs:14` introduces a `static INGEST_WRITE_LOCK: Mutex<()>` and the guard is acquired at line 134 *before* `allocate_filename` and held until `drop(_guard)` on line 188 — i.e. across allocate → `create_dir_all` → `write_attachments` → `write_markdown`. Poison recovery is handled with `unwrap_or_else(|poisoned| poisoned.into_inner())`, so a panic in one ingest can't deadlock the daemon. Channel triggers correctly run *after* the guard is released, so they don't serialize.

The `cleanup_bundle` closure has been hoisted and is now invoked from `inspect_err` on both `write_attachments` failure (line 165) and `write_markdown` failure (line 185), closing the rollback gap I called out. The `has_attachments && parent_dir != inbox_dir` predicate correctly avoids deleting the mailbox root in the flat-layout case.

The new `concurrent_same_subject_attachment_ingests_do_not_corrupt_each_other` test (8 threads racing identical subject + attachment) passes locally and asserts each ingest owns a distinct bundle directory with its attachment bytes intact.

This is symmetric to the planned outbound `Mutex<()>` for FR-19b — exactly the option I suggested.

**2. `mailbox_list` should scan `inbox/*/` — RESOLVED**

`src/mailbox.rs:85` adds `discover_mailbox_names(config)` that unions `config.mailboxes.keys()` with directory entries under `<data_dir>/inbox/`, deduped via `BTreeSet` for natural sort. `is_registered` lets callers distinguish stray dirs. `list_mailboxes` (CLI, line 67) and `list_mailboxes_with_unread` (MCP, `src/mcp.rs:431`) both consume it; stray dirs render with a yellow `(unregistered)` marker on the CLI and ` (unregistered)` suffix in MCP output. Catchall keeps surfacing as registered.

The decision to leave `serve.rs`/`status.rs` on `config.mailboxes.keys()` is correct — those paths need the registered-mailbox metadata (SMTP routing, status table) and shouldn't accept a stray dir as a deliverable mailbox.

Tests `list_surfaces_stray_inbox_dir_without_config_entry` (mailbox.rs) and `list_mailboxes_with_unread_surfaces_unregistered_dir` (mcp.rs) both pass; the existing `list_mailboxes_with_unread_returns_counts` test was extended to assert the `registered` flag.

### Verification

- `cargo test`: 530 unit + 51 integration pass (3 ignored as before).
- New concurrency test specifically passes when run in isolation.

### New observations (Nice-to-have, not blocking)

- `discover_mailbox_names` returns *any* directory name from `inbox/*/` without running it through `validate_mailbox_name`. So a hand-crafted `inbox/..hostile/` would surface in the listing display. Not exploitable — validation still gates create/delete/route — and stray dirs are explicitly an out-of-band recovery affordance, so this is intentional. Worth a follow-up to either skip names that fail validation or render them with an extra warning, but not blocking for v0.2.
- The MCP-side `list_mailboxes_with_unread` tuple grew from `(name, total, unread)` to `(name, total, unread, registered)`. If any external consumer of this private function exists I haven't seen, it would need updating — but `cargo build` and the test suite are clean.

### Unresolved

None.

### New issues introduced

None.

## Verdict

**Ready to merge.** Both Cycle-1 non-blockers are fully addressed with the exact mitigations I suggested (process Mutex symmetric to FR-19b; filesystem-scan union with `(unregistered)` marker for strays). Tests, clippy, and fmt are all clean.

```
[Sprint 36] Datadir Reshape — Inbox/Sent Split + Slug + Bundles

Reshape the on-disk storage layout to match v0.2 (FR-13b, FR-14, FR-15):

* Inbound mail now lives under `<data_dir>/inbox/<mailbox>/`; outbound
  copies will land under `<data_dir>/sent/<mailbox>/` (Sprint 38). The
  per-day `YYYY-MM-DD-NNN.md` counter is replaced with deterministic
  `YYYY-MM-DD-HHMMSS-<slug>.md` UTC filenames.
* Add `src/slug.rs` with `slugify` (lowercase → non-alphanumerics to `-`
  → collapse → trim → 20-char cap → `no-subject` fallback) and
  `allocate_filename` (chooses flat `<stem>.md` vs Zola-style bundle
  `<stem>/<stem>.md`, suffixes `-2`/`-3`/… on collision).
* Zero attachments → flat `.md`; one or more → bundle directory with
  attachments as siblings. The mailbox-level `attachments/` subdir is
  removed entirely; per-attachment `path` is now bundle-relative.
* `mailbox create` provisions both `inbox/<name>/` and `sent/<name>/`
  atomically (cleanup on second-create failure); `mailbox delete`
  removes both. `catchall` cannot be deleted.
* MCP tools `email_list`, `email_read`, `email_mark_read`, and
  `email_mark_unread` gain an optional `folder: \"inbox\" | \"sent\"`
  parameter (default `\"inbox\"`); invalid values return a clean error.
* Inbound writes are serialized through a process-scoped `Mutex<()>` so
  same-second + same-subject + same-attachment-filename ingests cannot
  interleave (symmetric to the planned outbound mutex in FR-19b);
  bundle directories are rolled back on attachment- or markdown-write
  failure.
* `mailbox list` (CLI) and the MCP `list_mailboxes` tool union
  `config.mailboxes` with a filesystem scan of `inbox/*/` and surface
  stray directories with an `(unregistered)` marker.
* `email_list`, `set_read_status`, status recent-activity, and the
  mailbox count helpers all walk both flat `.md` files and bundle
  directories (`<stem>/<stem>.md`).
* Channel rules' `{filepath}` now expands to the `.md` inside the
  bundle when attachments are present.
* Update `book/mailboxes.md`, `book/channels.md`, `book/mcp.md`, and
  `CLAUDE.md` for the new layout, slug filename format, and folder
  parameter.
```

PR #72, opened 2026-04-16
[Sprint 42] CLI UX: Config Errors + Setup Race + Version Hash

uzyn:
## Summary

Sprint 42 fixes two P0 UX issues that block first-time setup and lands one P1 build-traceability improvement:

- **S42-1** — `Config::load()` catches `io::ErrorKind::NotFound` and returns `Config file not found at <path> — run 'sudo aimx setup' first` instead of leaking raw `os error 2`. Applies to every config-dependent subcommand (`status`, `mcp`, `send`, `mailbox`, `serve`, `agent-setup`, `dkim-keygen`). Respects `AIMX_CONFIG_DIR`.
- **S42-2** — After `install_service_file()` restarts `aimx`, setup now polls `127.0.0.1:25` via TCP connect (~500ms interval, ~5s budget) before running the outbound/inbound port 25 probes, closing the race that caused the inbound check to fail on fresh installs. Exits early on first success; on timeout, proceeds to the port checks (the existing `Inbound port 25… FAIL` message still covers that path). Implemented behind `SystemOps::wait_for_service_ready` so tests mock it without real sleeps.
- **S42-3** — New `build.rs` runs `git rev-parse --short=8 HEAD` and exports `GIT_HASH` via `cargo:rustc-env`, with a graceful `"unknown"` fallback when git is unavailable or the tree isn't a repo. `cli.rs` composes the clap version string as `aimx 0.1.0 (abcd1234)` when present, or plain `aimx 0.1.0` when unknown. `rerun-if-changed` on `.git/HEAD` + `.git/refs` keeps the hash fresh without forcing full rebuilds.

## Test plan

- [x] `cargo test` — 619 unit tests + 52 integration tests pass
- [x] `aimx --version` prints `aimx 0.1.0 (<8-char-hash>)`
- [x] `AIMX_CONFIG_DIR=/tmp/nope aimx status` prints `Error: Config file not found at /tmp/nope/config.toml — run 'sudo aimx setup' first` (also verified for `mailbox list`, `send`, `dkim-keygen`, `serve`)
- [x] New setup tests assert `wait_for_service_ready` is called exactly once on fresh install and zero times on re-entrant setup, and that a timed-out wait still proceeds to the port-check summary

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,15 @@
+use std::process::Command;
+
+fn main() {
+    let hash = Command::new("git")
+        .args(["rev-parse", "--short=8", "HEAD"])
+        .output()
+        .ok()
+        .filter(|o| o.status.success())
+        .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_string())
+        .unwrap_or_else(|| "unknown".to_string());
+
+    println!("cargo:rustc-env=GIT_HASH={hash}");
+    println!("cargo:rerun-if-changed=.git/HEAD");
+    println!("cargo:rerun-if-changed=.git/refs");
+}
--- a/docs/sprint.md
+++ b/docs/sprint.md
@@ -300,7 +300,7 @@ Completed sprints 1–37 have been archived for context window efficiency.
 
 ---
 
-## Sprint 42 — CLI UX: Config Error Messages + Setup Port-Check Race + Version Hash (Days 118–120.5) [NOT STARTED]
+## Sprint 42 — CLI UX: Config Error Messages + Setup Port-Check Race + Version Hash (Days 118–120.5) [IN PROGRESS]
 
 **Goal:** Fix P0 UX issues that block first-time setup and improve build traceability: (1) commands that require config give a cryptic "os error 2" instead of pointing the user to `aimx setup`, (2) `aimx setup` fails the inbound port 25 check because it races against `aimx serve` startup, and (3) `aimx --version` includes the git commit hash so pre-release builds are distinguishable.
 
@@ -312,11 +312,11 @@ Completed sprints 1–37 have been archived for context window efficiency.
 
 **Priority:** P0
 
-- [ ] `config::load()` (or the call site in `main.rs`) catches `io::ErrorKind::NotFound` on the config file and returns a clear error naming the expected path and suggesting `sudo aimx setup`
-- [ ] Error message includes the actual path attempted (respects `AIMX_CONFIG_DIR` override)
-- [ ] All config-dependent subcommands benefit from the fix (status, mcp, send, mailbox, serve, agent-setup, dkim-keygen) — no raw "os error 2" leaks to the user
-- [ ] Unit test: calling config load with a nonexistent path produces the expected error message, not a raw IO error
-- [ ] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean
+- [x] `config::load()` (or the call site in `main.rs`) catches `io::ErrorKind::NotFound` on the config file and returns a clear error naming the expected path and suggesting `sudo aimx setup`
+- [x] Error message includes the actual path attempted (respects `AIMX_CONFIG_DIR` override)
+- [x] All config-dependent subcommands benefit from the fix (status, mcp, send, mailbox, serve, agent-setup, dkim-keygen) — no raw "os error 2" leaks to the user
+- [x] Unit test: calling config load with a nonexistent path produces the expected error message, not a raw IO error
+- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean
 
 #### S42-2: Wait-for-ready loop in `aimx setup` before port checks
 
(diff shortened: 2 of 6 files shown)


uzyn (commented):
## Sprint Goal Assessment

Sprint 42 delivers exactly what the sprint plan calls for: two P0 UX fixes (config-missing error message, setup port-check race) and one P1 build-traceability fix (git commit hash in `--version`). The implementation is tightly scoped, respects the existing `SystemOps` / `NetworkOps` trait-based testing pattern, and does not touch unrelated code. 619 unit tests + 52 integration tests pass, clippy --all-targets -D warnings is clean, fmt is clean.

**S42-1: Helpful error message when config file is missing**
- [x] `Config::load()` catches `io::ErrorKind::NotFound` and returns a clear message — verified in `src/config.rs:101-111`, branching on `e.kind()` and preserving non-NotFound errors verbatim
- [x] Error message includes the actual path (respects `AIMX_CONFIG_DIR`) — the `format!` uses `path.display()`, and `Config::load_resolved()` reads the path from `config_path()` which honors the env var; test `load_resolved_missing_config_includes_config_dir_path` exercises this
- [x] All config-dependent subcommands benefit — `main.rs:66` routes every command except Setup/Verify/Mcp/AgentSetup through `Config::load_resolved_with_data_dir`, which calls `Config::load_resolved` → `Config::load`. MCP goes through `AimxMcpServer::load_config` (`src/mcp.rs:106`), same path. No raw `read_to_string(?)` on `config_path` elsewhere
- [x] Unit test for helpful error — `load_missing_config_gives_helpful_error` + `load_resolved_missing_config_includes_config_dir_path`
- [x] `cargo test` / `clippy` / `fmt` clean — confirmed locally

**S42-2: Wait-for-ready loop in `aimx setup` before port checks**
- [x] After `restart_service("aimx")` returns, wait loop polls `127.0.0.1:25` — `setup.rs:1236` invokes `sys.wait_for_service_ready()` immediately after `install_service_file()` returns, before any port probe
- [x] 500ms interval, 5s total budget, 200ms connect timeout — `RealSystemOps::wait_for_service_ready` at `setup.rs:193-211` matches
- [x] Early exit on first success — `if TcpStream::connect_timeout(...).is_ok() { return true; }` in the loop head
- [x] Timeout proceeds to port checks without error — test `fresh_setup_proceeds_when_service_ready_times_out` asserts the run continues past the wait and errors out only at the port-check summary
- [x] Behind `SystemOps` trait — new trait method added with mock implementation tracking call counts via `RefCell<u32>`, so tests mock it without real sleeps
- [x] Existing + new tests pass — 184 setup tests pass including three new cases (fresh waits once, timeout proceeds, re-entrant skips)
- [x] `cargo test` / `clippy` / `fmt` clean

**S42-3: Include git commit hash in `aimx --version` output**
- [x] `build.rs` runs `git rev-parse --short=8 HEAD` and sets `GIT_HASH` via `cargo:rustc-env` — verified in `build.rs:1-15`
- [x] Fallback to `"unknown"` when git is unavailable — `.output().ok().filter(|o| o.status.success())...unwrap_or_else(|| \"unknown\".to_string())` handles both missing binary (ok → None) and non-repo/non-zero exit (filter drops it)
- [x] Version string `aimx 0.1.0 (abcd1234)` — confirmed locally: `target/debug/aimx --version` prints `aimx 0.1.0 (9ef26abd)`
- [x] `"unknown"` case omits the parenthetical — `cli.rs:7` guards `hash == "unknown" || hash.is_empty()`
- [x] `rerun-if-changed` on `.git/HEAD` + `.git/refs` — present at `build.rs:13-14`
- [x] `cargo test` / `clippy` / `fmt` clean

## Test Coverage

Coverage is proportional to the changes:

- S42-1 gets two dedicated tests in `config::tests` — one for direct `Config::load` on a nonexistent path, one that routes through `load_resolved` via `ConfigDirOverride` to prove the env-var path is reflected in the message. The existing `save_and_load_roundtrip` still exercises the happy path and implicitly confirms non-NotFound IO errors (read failures, parse failures) still fall through unchanged.
- S42-2 gets three dedicated tests in `setup::tests` that assert: (a) fresh install calls `wait_for_service_ready` exactly once, (b) a timed-out wait still reaches the port-check failure path, (c) re-entrant setup skips both `install_service_file` and the wait loop. The `RefCell<u32>` call counter on `MockSystemOps` is the right mechanism — it distinguishes "called once" from "called many times" so regressions where someone moves the wait into a loop would surface.
- S42-3 is verified manually (confirmed `aimx --version` prints the expected string locally on commit `9ef26abd`). No unit test for build.rs, which is appropriate — build scripts are awkward to unit-test and the output is directly exercised by the CLI behavior.

No gaps worth flagging.

## Potential Bugs

None found. A few things I looked at carefully and confirmed are fine:

- `Config::load` error mapping preserves non-NotFound errors as `Box<dyn Error>` via `Box::new(e)`, so `PermissionDenied` and `IsADirectory` still surface their native messages rather than being masked by the "Config file not found" helper.
- The wait loop uses `TcpStream::connect_timeout` with a 200ms connect timeout nested inside a 5s budget — in the worst case (every connect hangs for the full 200ms) you get ~10 attempts before timeout, which is still reasonable. If someone ever sets the connect timeout higher than the interval, the loop still honors the total 5s budget because the elapsed check fires unconditionally.
- `version_string()` uses `LazyLock<String>` and returns `&'static str` via a reference into the static. Since `LazyLock<T>` dereferences to `&T` for the lifetime of the program, returning `&VERSION` coerces to `&'static str`. Clippy didn't complain and the compile is clean.
- `rerun-if-changed=.git/refs` watches the directory's mtime, so ref creation/deletion/rename triggers rebuilds. If git packs refs into `.git/packed-refs` during gc, this could theoretically miss a hash update until the next file-touching build — but in practice refs only get packed on explicit `git gc`, not on commit, and the fallback ("unknown" hash if git rev-parse fails for any reason) means no build ever breaks. Not a bug for this sprint.

## Security Issues

No security-relevant changes. The config error message exposes the config path, which is either the hardcoded default `/etc/aimx/config.toml` or whatever `AIMX_CONFIG_DIR` was set to — same information the user already has. No sensitive data leak.

## Code Quality

- `version_string()` placement at the top of `cli.rs` is clean and the `LazyLock` pattern is idiomatic for one-time string composition with static references.
- The new `wait_for_service_ready` method fits the existing `SystemOps` trait shape — same return-`bool` pattern as `check_root`/`is_service_running`.
- The `"timeout (proceeding anyway)"` output string is not wrapped in a `term::*` helper like the surrounding success/fail badges, which is a tiny inconsistency with the adjacent `term::pass_badge()` output. Trivial, not worth fixing.

## Alignment with PRD

Nothing in the PRD contradicts this work. FR-1 (setup requires root) is untouched; FR-3 (port 25 checks after `aimx serve` is started) is actually strengthened by S42-2 — the checks now run after the daemon has bound, not just after systemctl returns. S42-1 implicitly improves the first-run UX path referenced in Goal #1 ("zero-to-working-email in one session < 15 minutes").

**Overall verdict:** Ready to merge

No blockers, no non-blockers, no nice-to-haves. All three stories meet their acceptance criteria, tests are appropriate and pass, clippy/fmt are clean, and the implementation matches the sprint plan's scope without creep.

```
[Sprint 42] CLI UX: config errors, setup race, version hash

S42-1: Config::load() catches io::ErrorKind::NotFound and returns a
helpful message naming the expected path and suggesting `sudo aimx
setup`, replacing the raw "os error 2" leak. Covers all config-dependent
subcommands (status, mcp, send, mailbox, serve, agent-setup, dkim-keygen)
and respects AIMX_CONFIG_DIR. Two unit tests.

S42-2: After install_service_file() restarts aimx, run_setup polls
127.0.0.1:25 via TCP connect (~500ms interval, ~5s budget) before
running the outbound/inbound probes, closing the race that caused the
inbound check to fail on fresh installs. Behind SystemOps so tests mock
it without real sleeps; three new tests cover fresh/timeout/re-entrant.

S42-3: New build.rs runs `git rev-parse --short=8 HEAD` and exports
GIT_HASH via cargo:rustc-env, falling back to "unknown" outside a git
repo. cli.rs composes the version string as `0.1.0 (<hash>)` or plain
`0.1.0` when unknown. rerun-if-changed on .git/HEAD + .git/refs keeps
the hash fresh without full rebuilds.

619 unit tests + 52 integration tests pass; clippy clean; fmt clean.
```

PR #75, opened 2026-04-17
Fix aimx setup: install aimx.service as the final step, after preflight + DNS

uzyn:
## Summary

`sudo aimx setup` was installing and starting `aimx.service` *mid-flow*, then running the port-25 preflight against the running daemon. This had two problems:

1. **The immediate bug reported by the user:** on a fresh install, `install_service_file` ran before `finalize_setup` wrote `/etc/aimx/config.toml` and the DKIM private key. The daemon exited immediately with `Error: Config file not found`, systemd restart-looped it 5× → permanently failed. `wait_for_service_ready` timed out, the inbound port-25 check had no listener to hit, and setup bailed before ever reaching `finalize_setup`. Every retry hit the same wall.
2. **The architectural smell underneath it:** coupling the preflight to a production service being up means a preflight failure leaves a broken systemd unit behind, and a half-configured setup is visible to any inbound SMTP client during DNS propagation. The ordering bug only existed because setup was trying to run the real daemon before setup was actually done.

This PR fixes both — the immediate bug by moving `finalize_setup` earlier, then the architecture by deferring the service install to the very end of setup so setup concludes with either a bound+verified `aimx serve` or a specific error pointing at `journalctl -u aimx`.

## New ordering in `run_setup`

1. Root check + domain
2. Port-25 conflict detection (foreign process → error)
3. TLS cert generation (fresh path only)
4. `finalize_setup` — writes `config.toml` + DKIM key; idempotent, handles domain change on re-entry
5. **Port-25 preflight**
   - Fresh install: bind a minimal SMTP responder on `:25` for the duration of the probe, tear it down afterwards
   - Re-entrant: the real daemon is already bound, probe it directly
6. DNS guidance loop (interactive)
7. MCP + Deliverability sections
8. **`install_and_verify_service`** (fresh path only): install the unit, restart, poll `:25` — error loudly if readiness times out

## Supporting changes

- Extract `verify::with_temp_smtp_listener` as a `pub(crate)` helper (was `run_with_temp_server`, limited to `aimx verify`) so setup's preflight reuses the same EHLO-speaking responder.
- Add `SystemOps::with_temp_smtp_listener` with a real-bind default; override in test mocks so unit tests don't contend for `:25`.
- Extract `install_and_verify_service` so the readiness-timeout error path is unit-testable.
- `wait_for_service_ready` timeout now HARD fails instead of "proceed anyway" — a daemon that didn't come up must surface to the operator, not silently proceed past.
- `systemctl reset-failed aimx` before restart so a prior stuck start-limit state from a broken install doesn't block recovery.

## Test plan

- [x] `cargo test` — 698 passed, 0 failed (main crate + verifier)
- [x] `cargo clippy -- -D warnings` — clean (both crates)
- [x] New unit tests:
  - `fresh_setup_defers_install_until_after_preflight` — preflight failure must NOT install the service
  - `fresh_setup_writes_config_and_dkim_before_preflight_and_install` — regression guard for the original bug
  - `install_and_verify_service_errors_when_service_never_binds` — readiness timeout is a hard error
  - `install_and_verify_service_succeeds_when_daemon_binds` — happy path
- [x] **End-to-end on the affected OVH VPS** with the release binary installed:
  - **Fresh** (service stopped, preflight via temp listener): `Outbound port 25… PASS`, `Inbound port 25… PASS` → DNS loop → `Starting aimx serve…` → `Waiting for aimx serve to bind port 25… PASS` → `aimx serve is running.` → `Setup complete for agent.zeroshot.lol!`
  - **Re-entrant** (service already bound): `Existing AIMX configuration detected. Skipping install, proceeding to verification.` → probes the live daemon → skips the install step → `Setup complete`
  - `sudo aimx verify` post-setup confirms `:25` reachable via EHLO probe
  - `aimx.service` reaches `active (running)` and binds `0.0.0.0:25`

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -167,6 +167,11 @@ impl SystemOps for RealSystemOps {
                 let _ = std::process::Command::new("systemctl")
                     .args(["enable", "aimx"])
                     .status();
+                // Clear any stuck "start-limit-hit" state from a prior
+                // failed install attempt so the restart below actually runs.
+                let _ = std::process::Command::new("systemctl")
+                    .args(["reset-failed", "aimx"])
+                    .status();
                 self.restart_service("aimx")?;
             }
             InitSystem::OpenRC => {
@@ -995,12 +1000,14 @@ pub fn finalize_setup(
         println!("DKIM keypair already exists.");
     }
 
+    Ok(())
+}
+
+fn announce_setup_complete(domain: &str) {
     println!(
         "\n{}\n",
         term::success_banner(&format!("Setup complete for {domain}!"))
     );
-
-    Ok(())
 }
 
 /// Create the config dir (default `/etc/aimx`, or `AIMX_CONFIG_DIR` override)
@@ -1201,6 +1208,8 @@ pub fn run_setup(
                 "Existing AIMX configuration detected. Skipping install, proceeding to verification."
             )
         );
+        // Keep config + DKIM state consistent (handles a changed domain on re-entry).
+        finalize_setup(data_dir, &domain, &dkim_selector)?;
     } else {
         // Step 2: Port 25 conflict detection
         match sys.check_port25_occupancy()? {
@@ -1217,7 +1226,7 @@ pub fn run_setup(
             }
         }
 
-        // Step 3: Generate TLS cert and install service file
+        // Step 3: Generate TLS cert
         let cert_dir = Path::new("/etc/ssl/aimx");
         if !sys.file_exists(&cert_dir.join("cert.pem")) {
             println!("Generating self-signed TLS certificate...");
@@ -1227,6 +1236,14 @@ pub fn run_setup(
             println!("TLS certificate already exists.");
         }
 
+        // Step 4: Write config.toml + DKIM keys BEFORE starting the service.
+        // aimx.service exits immediately with "Config file not found" if these
+        // are missing, which systemd then turns into a restart-loop and a
+        // permanently failed unit — causing wait_for_service_ready to time out
+        // and the inbound port-25 check to FAIL with no listener to hit.
+        finalize_setup(data_dir, &domain, &dkim_selector)?;
+
+        // Step 5: Install and start the service (config is now on disk).
         println!("Installing AIMX service...");
         sys.install_service_file(data_dir)?;
         println!("{}", term::success("`aimx serve` started."));
@@ -1236,7 +1253,7 @@ pub fn run_setup(
         if sys.wait_for_service_ready() {
             println!("{}", term::pass_badge());
         } else {
-            println!("timeout (proceeding anyway)");
+            println!("{} (proceeding anyway)", term::fail_badge());
         }
     }
 
@@ -1278,8 +1295,8 @@ pub fn run_setup(
         );
     }
 
-    // Step 6: DKIM keygen and finalize
-    finalize_setup(data_dir, &domain, &dkim_selector)?;
+    // Step 6: Announce success now that the service is running and port 25 is reachable.
+    announce_setup_complete(&domain);
 
     // Step 7: DNS guidance and verification (section [DNS])
     // Single `hostname -I` invocation (S32-4): derive both families from one call.
@@ -1433,6 +1450,14 @@ mod tests {
         existing_files: HashMap<PathBuf, String>,
         restarted_services: RefCell<Vec<String>>,
         service_file_installed: RefCell<bool>,
+        /// True if `/etc/aimx/config.toml` existed on disk at the moment
+        /// `install_service_file` was called. `None` until install is invoked.
+        /// Guards against re-introducing the bug where aimx.service is started
+        /// before config is written and fails with "Config file not found".
+        config_present_at_install: RefCell<Option<bool>>,
+        /// Same idea for the DKIM private key — aimx.service also refuses to
+        /// start without it.
+        dkim_present_at_install: RefCell<Option<bool>>,
         is_root: bool,
         port25_status: Port25Status,
         service_running: bool,
@@ -1447,6 +1472,8 @@ mod tests {
                 existing_files: HashMap::new(),
                 restarted_services: RefCell::new(vec![]),
                 service_file_installed: RefCell::new(false),
+                config_present_at_install: RefCell::new(None),
+                dkim_present_at_install: RefCell::new(None),
                 is_root: true,
                 port25_status: Port25Status::Free,
                 service_running: false,
@@ -1497,6 +1524,14 @@ mod tests {
         }
         fn install_service_file(&self, _data_dir: &Path) -> Result<(), Box<dyn std::error::Error>> {
             *self.service_file_installed.borrow_mut() = true;
+            // Snapshot the on-disk state that aimx.service would observe when
+            // systemd exec()s it. Real files — not the mock's `written_files`
+            // map — because finalize_setup writes config.toml + DKIM keys via
+            // the real filesystem (tests isolate via AIMX_CONFIG_DIR).
+            let config_present = crate::config::config_path().exists();
+            let dkim_present = crate::config::dkim_dir().join("private.key").exists();
+            *self.config_present_at_install.borrow_mut() = Some(config_present);
+            *self.dkim_present_at_install.borrow_mut() = Some(dkim_present);
             Ok(())
         }
         fn wait_for_service_ready(&self) -> bool {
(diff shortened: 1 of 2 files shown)


uzyn (commented):
# Review — PR #75: Install `aimx.service` as the final step of setup

## Changes Overview

Reorders `run_setup` so `finalize_setup` (config + DKIM write) runs before any attempt to start `aimx.service`, and defers `install_service_file` to the very last step of setup — after preflight, DNS guidance, MCP, and deliverability sections. A temporary EHLO-speaking SMTP responder covers port 25 during the preflight probe on fresh installs (reusing a `pub(crate)`-promoted helper in `verify.rs`). `wait_for_service_ready` timeout is now a hard error instead of a silent "proceed anyway", and `systemctl reset-failed aimx` runs before restart to clear stuck start-limit state from a prior broken install.

## Scope Alignment

Implementation matches the PR description 1:1. Every bullet in the "New ordering in `run_setup`" section maps to observable code:

1. Root check + domain — `run_setup:1180-1195` (unchanged)
2. Port-25 conflict detection — `run_setup:1218-1233`, hoisted out of the `!already_configured` branch to always run
3. TLS cert generation (fresh only) — `run_setup:1243-1252`
4. `finalize_setup` — `run_setup:1257`, now unconditional (fresh AND re-entrant)
5. Port-25 preflight — `run_setup:1264-1268`, correctly branches on `port25_status == Aimx` (probe directly) vs else (temp listener)
6. DNS loop / MCP / Deliverability — `run_setup:1270-1348`
7. `install_and_verify_service` on fresh path only — `run_setup:1353-1355`

No scope creep. No drive-by refactors. The `announce_setup_complete` extraction is load-bearing (it has to move from inside `finalize_setup` to after the install step).

## Potential Bugs

None found. A few edge cases I walked through that all resolve correctly:

- **`already_configured && port25_status == Free`**: weird transient state (`is_service_running` returns true but port 25 isn't held). Flow takes the else branch and binds the temp listener — which succeeds because nothing holds :25. Skips install. Leaves the operator with a daemon systemd thinks is running but which isn't. This is pre-existing and not in scope.
- **`!already_configured && port25_status == Aimx`**: preflight runs directly against the orphan aimx; install then runs `systemctl restart` which will replace it. Works because systemd ends up owning the new process. Pre-existing behavior.
- **User hits `q` in DNS loop before install step**: config + DKIM are on disk, no service is installed. Re-running setup: `is_service_running` returns false → `already_configured` false → fresh path runs again, `finalize_setup` is idempotent, install happens at the end. Clean recovery path. ✓
- **Ctrl+C mid-setup**: before this PR, an interrupt after `install_service_file` left systemd managing a broken daemon. Now almost all of setup runs before install — a clear improvement.
- **Temp listener bind race**: `check_port25_occupancy` returned Free, then something else binds :25 before `with_temp_smtp_listener`. The bind error surfaces cleanly via the `Cannot bind port 25 for preflight` message. ✓
- **`f()` panics inside `with_temp_smtp_listener`**: the runtime and spawned listener task are dropped cleanly because `rt` and `handle` unwind. The explicit `shutdown_tx.send(true)` is skipped but the task exits when the runtime drops. No listener leak. ✓

## Security Issues

None. No new attack surface; the temp listener only serves 220/EHLO/QUIT/502 and reads at most 30s per line (existing timeout). Binding :25 still requires root (`aimx setup` already mandates root).

## Test Coverage

Four new unit tests, each targets a real invariant:

- `fresh_setup_defers_install_until_after_preflight` — directly asserts `!service_file_installed && wait_for_ready_calls == 0` when preflight fails. Guards the architectural fix.
- `fresh_setup_writes_config_and_dkim_before_preflight_and_install` — asserts config + DKIM are on disk after a preflight failure, proving `finalize_setup` ran before the preflight (and therefore before any future install step). Guards the original bug.
- `install_and_verify_service_errors_when_service_never_binds` — proves readiness timeout is a hard error, not silent.
- `install_and_verify_service_succeeds_when_daemon_binds` — happy-path smoke.

Old tests `fresh_setup_waits_for_service_ready_after_install` and `fresh_setup_proceeds_when_service_ready_times_out` are correctly removed — the semantics they encoded ("wait after install", "proceed on timeout") no longer apply. Tests pass locally (`cargo test` at HEAD: `setup::tests` 186/186, full suite green), `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt -- --check` clean.

### Coverage gaps (nice-to-have, not blockers)

- The full **temp-listener-then-preflight-succeeds-then-install** happy path is not unit-tested end-to-end because the DNS retry loop reads stdin and there's no injection point. The PR description documents an end-to-end verification on the OVH VPS covering exactly this path — acceptable given the structural constraints.
- The real `verify::with_temp_smtp_listener` (the actual `tokio::net::TcpListener::bind("0.0.0.0:25")`) is still not unit-covered; both the setup mock and the verify-module mock short-circuit it. Pre-existing limitation (requires root to bind :25), unchanged by this PR.

## Code Quality

- `install_and_verify_service` and `run_port25_preflight` extractions are well-scoped — each has one clear job and sensible docstrings.
- `with_temp_smtp_listener` signature change (`FnOnce() -> Result<R, ...>` + generic `R`) is a clean generalization. `&mut dyn FnMut() -> Result<(), ...>` coerces to `FnOnce` via the stdlib `impl<F> FnOnce for &mut F where F: FnMut` blanket, so the setup-side trait wrapper compiles cleanly.
- `announce_setup_complete` is a touch thin (one `println!`), but pulling the banner out of `finalize_setup` is structurally necessary since `finalize_setup` now runs well before setup is actually done. Keep it.

### Nice-to-haves

- `MockSystemOps::install_service_file` (`setup.rs:1576-1581`) snapshots `config_present_at_install` and `dkim_present_at_install` into `RefCell<Option<bool>>` fields (declared `setup.rs:1504-1507`), but **no test asserts on those fields**. The tests that care about "config+DKIM on disk before install" either assert on the real filesystem (`fresh_setup_writes_config_and_dkim_before_preflight_and_install`) or never reach the install call at all (`fresh_setup_defers_install_until_after_preflight`). Either (a) wire those snapshots into a test that asserts both are `Some(true)` at the moment `install_service_file` ran on a hypothetical happy-path test, or (b) drop the snapshot instrumentation since it's dead. Leaving it as-is is mildly misleading — a future reader will assume there's a test guarding it.
- The test name `fresh_setup_writes_config_and_dkim_before_preflight_and_install` promises "before install", but install never runs in that test (preflight fails first). The test *does* prove "before preflight" and — together with the run-ordering in the source — implies "before install". Minor. Either accept or consider `..._before_preflight` to match what's actually asserted.

- **Blockers**: none
- **Non-blockers**: none
- **Nice-to-haves**:
  - Either use or remove the unused `config_present_at_install` / `dkim_present_at_install` snapshot fields in `MockSystemOps`.
  - Tighten the `fresh_setup_writes_config_and_dkim_before_preflight_and_install` test name to match what it actually asserts.

The architectural fix is the right one: setup was previously trying to run the real daemon as part of its own correctness check, which is what made the original "config not yet on disk" bug possible in the first place. Moving install to the very end — and making the readiness timeout a hard error — closes both the immediate bug and the class it belonged to.

```
Fix aimx setup: install aimx.service as the final step, after preflight + DNS

Setup previously installed and restarted aimx.service mid-flow, then ran
the port-25 preflight against the running daemon. Two problems:

1. On a fresh install, install_service_file ran before finalize_setup
   wrote /etc/aimx/config.toml and the DKIM private key. aimx serve
   exited with "Config file not found", systemd restart-looped it 5x
   within 60s and permanently failed the unit. wait_for_service_ready
   timed out; the inbound port-25 check had no listener to hit; setup
   bailed before ever reaching finalize_setup. Every retry hit the
   same wall.

2. Coupling preflight to a production service being up meant a
   preflight failure left a broken systemd unit behind, and a
   half-configured setup was visible on :25 during DNS propagation.

Restructure run_setup: root → port-25 conflict → TLS cert → finalize
(config + DKIM) → preflight (fresh: temp SMTP responder on :25;
re-entrant: probe the live daemon) → DNS guidance → MCP +
deliverability → install and verify aimx.service.

Extract verify::with_temp_smtp_listener as a pub(crate) helper so
setup's preflight reuses the same EHLO responder. Add
SystemOps::with_temp_smtp_listener for test-mockability. Extract
install_and_verify_service so the readiness-timeout error path is
unit-testable. wait_for_service_ready timeout is now a hard error
("check journalctl -u aimx") instead of silently proceeding.
systemctl reset-failed aimx before restart clears stuck start-limit
state from prior broken installs.

Verified end-to-end on the affected OVH VPS (fresh + re-entrant).
```

uzyn:
Addressed both nice-to-haves in 1278d9e:

1. **Dropped the unused snapshot fields.** `config_present_at_install` and `dkim_present_at_install` (along with the filesystem snapshotting inside `MockSystemOps::install_service_file`) are gone. As the review noted, no test asserts on them and the happy path they'd guard isn't reachable from a unit test without mocking stdin.
2. **Renamed the test** from `fresh_setup_writes_config_and_dkim_before_preflight_and_install` to `fresh_setup_writes_config_and_dkim_before_preflight`. The test forces a preflight failure so it never reaches the install step — the new name matches what's actually asserted. The docstring now explicitly calls out that the "before install" half of the invariant is carried by the `run_setup` source ordering (preflight → DNS → install) rather than by this test directly.

Nothing else changed. `cargo test` 698/698 green, `cargo clippy -- -D warnings` clean, `cargo fmt -- --check` clean.

PR #76, opened 2026-04-17
docs: add manual testing plan

uzyn:
## Summary
- Adds `docs/manual-test.md`, a copy-pasteable end-to-end manual testing plan for aimx on a fresh VPS.
- Covers T1–T13: setup + service start, mailboxes, inbound/outbound mail, routing, Claude Code + Codex CLI MCP integration, channel triggers (trusted/untrusted/outbound-does-not-fire), trust model definition, reboot survival, and setup re-entrance.
- Includes explicit `[VPS]` / `[LOCAL]` / handoff markers so tests that need external Gmail sends or receipt at `chua@uzyn.com` are clearly delegated.

## Test plan
- [ ] Human operator walks through T0 → T13 on a fresh VPS and records which acceptance criteria pass
- [ ] Capture any discrepancies between documented `trusted`-field semantics and actual behavior observed in T9 (the "trigger gate is looser than trusted field" subtlety)
- [ ] Feed regressions found during the walkthrough back into follow-up issues

### DIFF
--- a/docs/manual-test.md
+++ b/docs/manual-test.md
@@ -6,7 +6,7 @@ aimx is a self-hosted email server for AI agents. This plan manually verifies an
 
 **Test domain assumption.** Replace `mail.example.com` with your real test domain everywhere below. The plan uses `inbox`, `agent`, and `test` as mailbox names.
 
-**Notation.** `[VPS]` = run on the aimx VPS. `[LOCAL]` = run on your laptop (where `claude` / `codex` are installed). `[GMAIL]` = user action in Gmail web UI.
+**Notation.** `[VPS]` = run on the aimx VPS (the same machine where `claude` / `codex` are installed — aimx is designed for a host dedicated to AI agents). `[GMAIL]` = user action in Gmail web UI. Commands run as root unless the step says otherwise; `aimx send` and `aimx agent-setup` refuse to run as root, so run those as a non-root user on the VPS.
 
 ---
 
@@ -14,9 +14,9 @@ aimx is a self-hosted email server for AI agents. This plan manually verifies an
 
 1. Fresh Linux VPS with public IPv4, port 25 unblocked by the provider.
 2. DNS zone editable for the test domain.
-3. `chua@uzyn.com` mailbox available on your laptop (to observe outbound).
-4. A Gmail account available (to observe inbound external deliverability).
-5. `claude` and `codex` CLIs installed and logged in on your laptop.
+3. Access to the `chua@uzyn.com` mailbox from anywhere (to observe outbound).
+4. A Gmail account (to observe inbound external deliverability).
+5. `claude` and `codex` CLIs installed and logged in **on the VPS**, under a non-root user (the same user that will run `aimx send` and `aimx agent-setup`).
 
 ```bash
 # [VPS] sanity — port 25 not already bound
@@ -197,7 +197,7 @@ sudo journalctl -u aimx -n 200 --no-pager | grep -i 'nobody\|reject\|550'
 **Goal.** The Claude Code skill is installed, the MCP server starts, and Claude can list and send email.
 
 ```bash
-# [LOCAL]
+# [VPS] run as the non-root user that owns the claude install
 aimx agent-setup claude-code
 ls -la ~/.claude/plugins/aimx/
 # expect: plugin.json + skills/ + references/
@@ -206,14 +206,14 @@ ls -la ~/.claude/plugins/aimx/
 Restart Claude Code (exit and relaunch), then test:
 
 ```bash
-# [LOCAL] list inbox via MCP
+# [VPS] list inbox via MCP
 claude -p "Use the aimx MCP tools. Call email_list with mailbox='inbox' and unread=true. Summarize what's there."
 ```
 
 **Acceptance — list.** Claude responds with a summary matching the emails delivered in T3/T5.
 
 ```bash
-# [LOCAL] send via MCP
+# [VPS] send via MCP
 claude -p "Use the aimx MCP tool email_send to send from inbox@mail.example.com to chua@uzyn.com. Subject: 'aimx claude MCP test'. Body: 'Sent via Claude MCP.'"
 ```
 
@@ -225,7 +225,7 @@ claude -p "Use the aimx MCP tool email_send to send from inbox@mail.example.com 
 - Recipient receives the mail with DKIM=pass.
 
 ```bash
-# [LOCAL] read + mark_read
+# [VPS] read + mark_read
 claude -p "List unread mail in 'inbox'. Read the most recent one, summarize in one sentence, then mark it read."
 ```
 
@@ -236,7 +236,7 @@ claude -p "List unread mail in 'inbox'. Read the most recent one, summarize in o
 ## T7 — Codex CLI agent integration
 
 ```bash
-# [LOCAL]
+# [VPS] run as the non-root user that owns the codex install
 aimx agent-setup codex
 ls -la ~/.codex/plugins/aimx/
 ```
@@ -244,7 +244,7 @@ ls -la ~/.codex/plugins/aimx/
 Follow any printed `codex mcp add ...` step verbatim, then restart Codex.
 
 ```bash
-# [LOCAL]
+# [VPS]
 codex exec "Use the aimx MCP tools. Call email_list with mailbox='inbox' and unread=true. Summarize."
 
 codex exec "Use the aimx MCP tool email_send to send from inbox@mail.example.com to chua@uzyn.com. Subject: 'aimx codex MCP test'. Body: 'Sent via Codex MCP.'"
--- a/docs/manual-test.md
+++ b/docs/manual-test.md
@@ -1,0 +1,456 @@
+# aimx Manual Testing Plan
+
+## Context
+
+aimx is a self-hosted email server for AI agents. This plan manually verifies an end-to-end install on a fresh VPS: setup, inbound/outbound mail, agent integrations (Claude Code + Codex CLI), channel triggers, and the trust model. Every test below is copy-pasteable — run top-to-bottom on the VPS as root unless noted. Tests that need external help (sending from Gmail, receiving at `chua@uzyn.com`) call that out explicitly and include a prompt to hand off to a human collaborator.
+
+**Test domain assumption.** Replace `mail.example.com` with your real test domain everywhere below. The plan uses `inbox`, `agent`, and `test` as mailbox names.
+
+**Notation.** `[VPS]` = run on the aimx VPS. `[LOCAL]` = run on your laptop (where `claude` / `codex` are installed). `[GMAIL]` = user action in Gmail web UI.
+
+---
+
+## T0 — Pre-flight
+
+1. Fresh Linux VPS with public IPv4, port 25 unblocked by the provider.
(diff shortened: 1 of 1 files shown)


uzyn (commented):
## Changes Overview

Adds `docs/manual-test.md` (456 lines), a copy-pasteable T0–T13 manual test plan for a fresh-VPS aimx install: setup, inbound/outbound, routing, Claude Code + Codex MCP integration, channel triggers (trusted/untrusted/outbound), trust-model definition, reboot, and setup re-entrance. Pure docs addition — no code changes.

Since this is a docs PR, the "implementation" is the document itself. I spot-checked every command, path, expected exit code, and acceptance criterion against the current code. Most are accurate; a few will actively misdirect a tester.

> Posted as a comment: GitHub blocked `--request-changes` because the reviewer is the PR author. Treat the Blockers list below as change requests.

## Potential Bugs (doc-level factual errors)

### BLOCKER — T7 points testers at the wrong Codex install path
`docs/manual-test.md:182-183` (and Teardown at `:434`) instruct:
```
ls -la ~/.codex/plugins/aimx/
```
but `src/agent_setup.rs:71` sets the Codex destination to `$HOME/.codex/skills/aimx`, not `~/.codex/plugins/aimx`. Codex specifically does NOT auto-discover plugins (see `src/agent_setup.rs:115-135`) — that's why it lives under `skills/`. A tester following the doc will see "No such file or directory" and conclude `aimx agent-setup codex` failed when it actually succeeded. Fix both the `ls` in T7 and the `rm -rf` line in Teardown.

### BLOCKER — T6 describes a layout the Claude Code installer doesn't produce
`docs/manual-test.md:157-159`:
```
# expect: plugin.json + skills/ + references/
```
The installed tree at `~/.claude/plugins/aimx/` is actually:
- `.claude-plugin/plugin.json` (dot-directory; hidden from plain `ls`, only shown with `ls -la`)
- `skills/aimx/SKILL.md`
- `skills/aimx/references/*.md` (references are siblings of SKILL.md — see `src/agent_setup.rs:584-599`)

There is no top-level `plugin.json` and no top-level `references/`. As written, the "expect" line will not match reality. Either list the real layout or remove the expectation.

### BLOCKER — T9 acceptance criterion directly contradicts the v1 trigger semantics the doc itself documents
`docs/manual-test.md:253-256`:
```
ls /tmp/aimx-trigger-*.flag 2>/dev/null || echo "no flag — PASS"
...
**Acceptance.**
- No flag file was created.
```
…but the footnote at `:260-262` correctly states that the channel-trigger gate fires on allowlist **OR** DKIM-pass. Gmail's DKIM passes, so sending from any Gmail address (on or off the allowlist) **will** fire the trigger under current code (`src/channel.rs` via `should_execute_triggers`, and the parity test in `src/trust.rs:258-332` explicitly verifies this asymmetry). So the printed "no flag — PASS" test will print "no flag — PASS" only if the trigger behaves contrary to its documented v1 semantics.

The net effect: a tester who follows T9 literally will mark this as a regression when they see the flag file, even though the code is behaving exactly as specified. Either:
- Rewrite the acceptance to match v1 semantics: "flag file **is** created because Gmail's DKIM passes; confirm `trusted = \"false\"` in the frontmatter (the gate is wider than the trust field — this is documented below)"; or
- Use a sender whose DKIM fails (as the footnote suggests) and keep the "no flag" expectation — but then the doc needs to tell the tester how to produce that.

As written, the top-of-T9 acceptance and the bottom-of-T9 subtlety note disagree.

## Non-blockers

1. **T1 "Interactive steps to answer" (`:38-42`) lists prompts that don't exist.** `sudo aimx setup mail.example.com` passes the domain as an arg, so the domain prompt at `src/setup.rs:1174` is skipped. There is no verify-host prompt (it's CLI-flag-only, `src/setup.rs` + `src/cli.rs:59-60`). There is no "install the service" yes/no prompt — service installation is unconditional (`src/setup.rs:1230-1231`). Of the four bullets, only "Capture the printed DKIM TXT record" is real, and it's an observation, not a prompt. Rewrite the section as "What setup prints (capture the DKIM TXT line)".

2. **T2 expects three mailbox directories but four will exist.** `aimx setup` writes a default `catchall` mailbox into config (`src/config.rs:174, :236`), and `aimx mailbox create` calls in T2 don't remove it. So after T2, `/var/lib/aimx/inbox/` contains `catchall/` + `inbox/` + `agent/` + `test/`. The "Three mailbox directories" acceptance will fail as written.

3. **T5 unknown-local-part will not be rejected and not be in journalctl grep set.** `src/config.rs:321-323` routes unknown local-parts to `catchall` — silently, no SMTP-level 550. The `journalctl | grep -i 'nobody\|reject\|550'` probe on `:210` will return nothing, and the tester will be stuck. Point the check at `/var/lib/aimx/inbox/catchall/` — the `route-test-nobody` message lives there. The T5 hedging text already allows for this; the commands just don't check the right place.

4. **T8 "Edit /etc/aimx/config.toml to add" will create a duplicate `[mailboxes.test]` section.** T2 already runs `aimx mailbox create test`, which writes `[mailboxes.test]` with defaults via `src/mailbox.rs:51-59`. Appending the T8 block verbatim produces two `[mailboxes.test]` tables and TOML parse behavior becomes implementation-defined. Change the instruction to "Update the existing `[mailboxes.test]` entry to set `trust = \"verified\"` and `trusted_senders = [...]`, and append a `[[mailboxes.test.on_receive]]` sub-table under it."

5. **T4 failure branches don't isolate the config-shape failures.** The `aimx send --from bogus@mail.example.com ...` branch expects `AIMX/1 ERR MAILBOX`, which matches `src/send_handler.rs:81-86`. Good. But note that if the tester's *domain* doesn't match (e.g., a typo picks up `mail.example.com` when the configured domain is something else), they'll instead see `ERR DOMAIN` (`src/send_handler.rs:121-129`). Worth a one-line note: "If you see `ERR DOMAIN`, check `/etc/aimx/config.toml` domain value."

## Nice-to-haves

- T3's `sudo find /var/lib/aimx/inbox/inbox/ -name '*.md' -newer /tmp -print` is unusual — `/tmp` mtime isn't a stable reference for "files newer than the test started". Suggest `ls -lt /var/lib/aimx/inbox/inbox/` and pick the top entry, or `touch /tmp/.aimx-t3-mark` before the Gmail send and use `-newer /tmp/.aimx-t3-mark`.
- T11 re-states the exact definition from `src/trust.rs` — that's fine, but drift risk is high. Either link to the file or inline a snippet so it's obvious the doc is the derivative.
- T12's "Wait 60s" is generous for reboot; some VPS providers take longer. Consider "Wait until SSH reconnects" as the cue.
- Teardown doesn't stop `aimx` on OpenRC hosts (uses `systemctl disable --now`). If you want this plan to support both init systems end-to-end, add the OpenRC equivalent or note that it's systemd-only.

**Overall verdict:** Needs minor fixes — scope is appropriate, the document is genuinely useful, but the three blocker items will actively mislead the first real walkthrough.

1. T7 (and Teardown) — wrong Codex path; should be `~/.codex/skills/aimx/`, not `~/.codex/plugins/aimx/`.
2. T6 — installed Claude Code tree has `.claude-plugin/plugin.json` and `skills/aimx/references/`, not top-level `plugin.json`/`references/`.
3. T9 — acceptance says "no flag created" but a Gmail sender (DKIM pass) will fire the trigger under documented v1 semantics; rewrite the acceptance to match the footnote.

1. T1's "Interactive steps" lists three non-existent prompts.
2. T2 will see four mailbox dirs (catchall is created by setup).
3. T5's journalctl grep won't match catchall-routed mail; check `/var/lib/aimx/inbox/catchall/` instead.
4. T8 should instruct editing the existing `[mailboxes.test]` section, not adding a duplicate.
5. T4 — add a one-liner about `ERR DOMAIN` vs `ERR MAILBOX` for config-shape mistakes.

**Nice-to-haves:**
- T3 use an explicit mark file for `-newer` instead of `/tmp`.
- T11 link to `src/trust.rs` to prevent drift.
- T12 cue on SSH reconnect rather than fixed 60s.
- Teardown cover OpenRC or scope the doc to systemd explicitly.

uzyn:
Addressed review feedback in 2532d23. Summary of what changed (and a note on the one point where I verified the reviewer's claim externally because it contradicted my own recollection):

### Research note — Claude Code plugin layout

The reviewer's claim about `~/.claude/plugins/aimx/.claude-plugin/plugin.json` is canonical, confirmed two ways:
1. The source tree bundled into the binary lives at `agents/claude-code/.claude-plugin/plugin.json` (dot-subdir), and `src/agent_setup.rs:917` asserts exactly that layout after install.
2. Per the [Claude Code Plugins Reference](https://code.claude.com/docs/en/plugins-reference), the plugin manifest must live under `.claude-plugin/`; component dirs (`skills/`, `commands/`, etc.) live at the plugin root. Skills correctly sit at `skills/aimx/SKILL.md` with `skills/aimx/references/` as a sibling.

So the fix for T6 is to rewrite the expectation, not to change the installer.

### Research note — Codex does not auto-discover plugins

Confirmed against Codex CLI 0.117.0 (already captured in `src/agent_setup.rs:115-135` and the test at `agent_setup.rs:1144`). Codex MCP wiring is `~/.codex/config.toml` via `codex mcp add`; there is no `~/.codex/plugins/` auto-discovery. T7 and Teardown now install to / clean up `~/.codex/skills/aimx/` and include the `codex mcp add` / `codex mcp remove` registration steps.

### Blockers

- **T6**: expected tree now reads `.claude-plugin/plugin.json`, `skills/aimx/SKILL.md`, `skills/aimx/references/*.md`; added a second `ls` into the dot-subdir so the plugin.json actually shows up in output.
- **T7 + Teardown + Files-referenced section**: path changed to `~/.codex/skills/aimx/`; added `codex mcp add ...` verification step (`grep [mcp_servers.aimx] ~/.codex/config.toml`); Teardown now includes `codex mcp remove aimx`.
- **T9**: retitled to "Off-allowlist sender: trigger gate vs. `trusted` field". Acceptance now matches v1 semantics (`should_execute_triggers` = allowlisted OR DKIM-pass): flag file **is** created because Gmail's DKIM passes; frontmatter still has `trusted = "false"` because the `trusted` field requires both conditions. Footnote updated to call this out as expected, not a regression.

### Non-blockers

- **T1**: dropped the "Interactive steps" list (domain is CLI-arg, no verify-host prompt, service install is unconditional) and replaced with "What setup prints".
- **T2**: acceptance now expects four mailbox dirs including the pre-created `catchall/`.
- **T5**: check `/var/lib/aimx/inbox/catchall/` for the unknown-local-part message (no SMTP reject in v1); updated goal line to match.
- **T8**: instruction now says update the existing `[mailboxes.test]` in place, with a callout that a duplicate TOML table is undefined.
- **T4**: added a one-liner distinguishing `ERR DOMAIN` (wrong sender domain) from `ERR MAILBOX` (unregistered local-part).

### Nice-to-haves

- **T3**: `touch /tmp/.aimx-t3-mark` before the Gmail send, use it as the `-newer` reference instead of `/tmp`.
- **T12**: "Wait until SSH reconnects (typically ~30-90s depending on provider)" instead of a fixed 60s.
- **Teardown**: added OpenRC equivalent for stop/disable/unit-removal alongside the systemd steps.
- **T11**: already references `src/trust.rs::evaluate_trust` — left as-is.

Nothing was left unaddressed. No code changes; docs-only PR.

PR #77, opened 2026-04-17
fix(ingest): use envelope MAIL FROM and peer IP for SPF

uzyn:
## Summary

Fixes an SPF bug where inbound messages received via the embedded SMTP daemon were producing `spf = "none"` in the frontmatter even when SPF should clearly pass.

**Reported case:** a message from `chua@uzyn.com` delivered via Gmail (peer IP `209.85.219.48`, within `_spf.google.com`) landed with:
```toml
dkim = "pass"
spf = "none"     # ← should be "pass"
dmarc = "pass"
```

`uzyn.com` has a proper SPF record (`v=spf1 a mx ip4:175.41.131.3 include:_spf.google.com ~all`), so SPF against `209.85.219.48` must pass.

## Root cause

`verify_auth()` ignored the authoritative envelope data from the SMTP session and parsed `Received:` and `From:` headers from the message body instead.

1. **Wrong IP source.** `extract_received_ip(raw)` walks `Received:` headers looking for a bracketed IP. For Gmail-sent messages the topmost `Received:` header is a `by mail-xxx.google.com with SMTP ...` line with no bracketed IP, so the code either skipped it or landed on an internal Google hop — not necessarily the IP that actually connected to us.
2. **Wrong mail_from source.** `extract_mail_from(raw)` returns the **header** `From:` address, but SPF (RFC 7208) is defined against the **envelope** MAIL FROM (the SMTP reverse-path).

Both values were already known authoritatively by the SMTP session (`peer_addr` and `reverse_path`), but they weren't being plumbed through.

## Fix

Thread the SMTP session's peer IP and `reverse_path` through to `ingest_email()` and into `verify_auth()`. When available, these authoritative values are used for SPF (and for DMARC mail-from alignment).

Header parsing remains as a best-effort fallback for the manual `aimx ingest` stdin path, where no SMTP session exists.

### Bug fix: SPF verification uses authoritative envelope data
- [x] `ingest_email()` accepts an `envelope_mail_from: Option<&str>`
- [x] `verify_auth()` accepts `peer_ip` and `envelope_mail_from` and uses them for SPF
- [x] `verify_dmarc_async()` uses the same authoritative mail_from for alignment
- [x] `src/smtp/session.rs` threads `session_state.reverse_path` through `ingest_email`
- [x] Manual `aimx ingest` stdin path passes `None` and keeps header-parsing fallback
- [x] Unit tests for the new `select_spf_inputs()` helper covering envelope-wins, stdin-fallback, and empty-envelope edge cases
- [x] All 648 unit + 53 integration tests pass; `cargo clippy -- -D warnings` clean; `cargo fmt` clean

## Technical Decisions

- **Added a new parameter rather than a struct.** `ingest_email()` has 30+ call sites in tests. A struct would have been a bigger diff without real ergonomic gain at this call count.
- **Kept header parsing as a fallback** (gated on `peer_ip.is_unspecified()` and `envelope_mail_from.is_none()`). This preserves `aimx ingest` stdin behavior, where the operator pipes a raw `.eml` and no envelope context exists. Setting SPF to "none" for that path would be a regression for manual-ingest workflows.
- **Extracted `select_spf_inputs()`** as a pure helper so the precedence logic is directly unit-testable without running a real SPF DNS lookup.

## Review Focus Areas

- `src/ingest.rs:315-338` — `select_spf_inputs()`: is the envelope-vs-header precedence right, including the empty-string defense?
- `src/smtp/session.rs:493-513` — threading `reverse_path` through `spawn_blocking`; the clone is cheap (small `String`) and unavoidable because we move into the blocking task.
- All 30+ test call sites were updated mechanically to pass `None`; nothing should have changed semantically for them.

### DIFF
--- a/src/ingest.rs
+++ b/src/ingest.rs
@@ -23,9 +23,9 @@ pub fn run(rcpt: &str, config: Config) -> Result<(), Box<dyn std::error::Error>>
     let mut raw = Vec::new();
     std::io::stdin().read_to_end(&mut raw)?;
     // Manual stdin path: no SMTP session, so received_from_ip is the
-    // unspecified sentinel (0.0.0.0).
+    // unspecified sentinel (0.0.0.0) and there is no envelope MAIL FROM.
     let sentinel_ip: IpAddr = "0.0.0.0".parse().unwrap();
-    ingest_email(&config, rcpt, &raw, sentinel_ip)
+    ingest_email(&config, rcpt, &raw, sentinel_ip, None)
 }
 
 pub fn ingest_email(
@@ -33,6 +33,7 @@ pub fn ingest_email(
     rcpt: &str,
     raw: &[u8],
     received_from_ip: IpAddr,
+    envelope_mail_from: Option<&str>,
 ) -> Result<(), Box<dyn std::error::Error>> {
     let local_part = extract_local_part(rcpt);
     let mailbox = config.resolve_mailbox(local_part);
@@ -147,7 +148,7 @@ pub fn ingest_email(
     let prepared_attachments = prepare_attachments(&message);
     let has_attachments = !prepared_attachments.is_empty();
 
-    let auth_results = verify_auth(raw);
+    let auth_results = verify_auth(raw, received_from_ip, envelope_mail_from);
 
     let trusted_value = config
         .mailboxes
@@ -260,7 +261,7 @@ fn create_resolver() -> Option<mail_auth::MessageAuthenticator> {
         .ok()
 }
 
-fn verify_auth(raw: &[u8]) -> AuthResults {
+fn verify_auth(raw: &[u8], peer_ip: IpAddr, envelope_mail_from: Option<&str>) -> AuthResults {
     let rt = match tokio::runtime::Runtime::new() {
         Ok(rt) => rt,
         Err(_) => return AuthResults::default(),
@@ -281,11 +282,21 @@ fn verify_auth(raw: &[u8]) -> AuthResults {
 
         let dkim_result = dkim_output_to_string(&dkim_output, auth_msg.is_some());
 
-        let spf_output = build_spf_output(raw, &resolver).await;
+        let (spf_ip, spf_mail_from) = select_spf_inputs(raw, peer_ip, envelope_mail_from);
+        let spf_output = build_spf_output(spf_ip, spf_mail_from.as_deref(), &resolver).await;
         let spf_result = spf_output_to_string(&spf_output);
 
         let dmarc_result = match &auth_msg {
-            Some(msg) => verify_dmarc_async(msg, &dkim_output, &spf_output, raw, &resolver).await,
+            Some(msg) => {
+                verify_dmarc_async(
+                    msg,
+                    &dkim_output,
+                    &spf_output,
+                    spf_mail_from.as_deref(),
+                    &resolver,
+                )
+                .await
+            }
             None => "none".to_string(),
         };
 
@@ -295,6 +306,31 @@ fn verify_auth(raw: &[u8]) -> AuthResults {
             dmarc: dmarc_result,
         }
     })
+}
+
+/// Decide which IP and MAIL FROM to feed into the SPF check.
+///
+/// When the SMTP session gives us an authoritative peer IP and envelope
+/// MAIL FROM (reverse_path), we use those — they're what RFC 7208 actually
+/// requires for SPF. Header parsing is only a best-effort fallback for the
+/// manual `aimx ingest` stdin path, where we have no envelope context.
+fn select_spf_inputs(
+    raw: &[u8],
+    peer_ip: IpAddr,
+    envelope_mail_from: Option<&str>,
+) -> (Option<IpAddr>, Option<String>) {
+    let ip = if peer_ip.is_unspecified() {
+        extract_received_ip(raw)
+    } else {
+        Some(peer_ip)
+    };
+
+    let mail_from = match envelope_mail_from {
+        Some(s) if !s.is_empty() => Some(s.to_string()),
+        _ => extract_mail_from(raw),
+    };
+
+    (ip, mail_from)
 }
 
 fn dkim_output_to_string(results: &[mail_auth::DkimOutput<'_>], parsed: bool) -> String {
@@ -316,17 +352,18 @@ fn dkim_output_to_string(results: &[mail_auth::DkimOutput<'_>], parsed: bool) ->
 }
 
 async fn build_spf_output(
-    raw: &[u8],
+    ip: Option<IpAddr>,
+    mail_from: Option<&str>,
     resolver: &mail_auth::MessageAuthenticator,
 ) -> mail_auth::SpfOutput {
-    let ip = match extract_received_ip(raw) {
+    let ip = match ip {
         Some(ip) => ip,
         None => return mail_auth::SpfOutput::new(String::new()),
     };
 
-    let mail_from = extract_mail_from(raw).unwrap_or_default();
+    let mail_from = mail_from.unwrap_or_default();
 
-    let helo_domain = match spf_domain(&mail_from) {
+    let helo_domain = match spf_domain(mail_from) {
         Some(d) => d.to_string(),
         None => return mail_auth::SpfOutput::new(String::new()),
     };
@@ -336,7 +373,7 @@ async fn build_spf_output(
             ip,
             &helo_domain,
             &helo_domain,
-            &mail_from,
+            mail_from,
         ))
         .await
 }
@@ -356,11 +393,11 @@ async fn verify_dmarc_async(
     auth_msg: &mail_auth::AuthenticatedMessage<'_>,
     dkim_output: &[mail_auth::DkimOutput<'_>],
     spf_output: &mail_auth::SpfOutput,
-    raw: &[u8],
+    mail_from: Option<&str>,
     resolver: &mail_auth::MessageAuthenticator,
 ) -> String {
-    let mail_from = extract_mail_from(raw).unwrap_or_default();
-    let mail_from_domain = spf_domain(&mail_from).unwrap_or("");
+    let mail_from = mail_from.unwrap_or_default();
+    let mail_from_domain = spf_domain(mail_from).unwrap_or("");
 
     let params = mail_auth::dmarc::verify::DmarcParameters {
         message: auth_msg,
@@ -897,7 +934,14 @@ mod tests {
     fn ingest_plain_text() {
         let tmp = TempDir::new().unwrap();
         let config = test_config(tmp.path());
-        ingest_email(&config, "alice@test.com", plain_text_eml(), sentinel_ip()).unwrap();
+        ingest_email(
+            &config,
+            "alice@test.com",
+            plain_text_eml(),
+            sentinel_ip(),
+            None,
+        )
+        .unwrap();
 
         let entries = collect_md_files(&inbox(tmp.path(), "alice"));
         assert_eq!(entries.len(), 1);
@@ -919,7 +963,14 @@ mod tests {
     fn ingest_writes_to_inbox_subdir() {
         let tmp = TempDir::new().unwrap();
         let config = test_config(tmp.path());
-        ingest_email(&config, "alice@test.com", plain_text_eml(), sentinel_ip()).unwrap();
+        ingest_email(
+            &config,
+            "alice@test.com",
+            plain_text_eml(),
+            sentinel_ip(),
+            None,
+        )
+        .unwrap();
 
         assert!(tmp.path().join("inbox").join("alice").exists());
         assert!(!tmp.path().join("alice").exists());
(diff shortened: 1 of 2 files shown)


uzyn (commented):
## Changes Overview

Fixes an SPF bug in the embedded SMTP ingest path where `verify_auth()` was deriving the client IP and mail_from from message body headers (`Received:`, `From:`) instead of the authoritative SMTP envelope data that was already available in the session. The fix threads `peer_addr.ip()` and `session_state.reverse_path` through `ingest_email()` into `verify_auth()`, extracts the precedence logic into a pure `select_spf_inputs()` helper, and also routes the envelope mail_from into DMARC alignment (which is semantically more correct — DMARC's `rfc5321_mail_from_domain` param specifically wants the envelope value). Header parsing remains a best-effort fallback for the `aimx ingest` stdin path.

## Scope Alignment

Implementation matches the PR description exactly. All eight items in the "Stories Implemented" checklist are met:

- `ingest_email()` signature extended with `envelope_mail_from: Option<&str>` (`src/ingest.rs:36`)
- `verify_auth()` accepts `peer_ip` and `envelope_mail_from` (`src/ingest.rs:264`)
- `select_spf_inputs()` extracted as a pure helper with the documented precedence (`src/ingest.rs:317-334`)
- DMARC uses the same authoritative mail_from (`src/ingest.rs:289-299`, `src/ingest.rs:392-400`)
- `src/smtp/session.rs:498-506` clones `reverse_path` into the `spawn_blocking` closure and maps empty → `None` before passing it through
- Manual stdin path at `src/ingest.rs:28` passes `None`; header-parsing fallback preserved
- 3 new unit tests on `select_spf_inputs` cover envelope-wins, stdin-fallback, and empty-envelope edge cases (`src/ingest.rs:1891-1943`)
- Full test suite clean: 648 unit + 53 integration passing locally; `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check` clean

No scope creep — the refactor of `verify_dmarc_async` to take `Option<&str>` instead of re-parsing `raw` is a natural byproduct of the fix and keeps mail_from consistent between SPF and DMARC.

## Potential Bugs

None that block merge. Two points worth flagging:

**1. RFC 5321 null sender (`MAIL FROM:<>`) behavior — Non-blocker.**
When a legitimate bounce arrives with `MAIL FROM:<>`, `extract_angle_addr("<>")` returns `""` (`src/smtp/session.rs:663`), `session.rs:501` maps empty → `None`, and `select_spf_inputs()` then falls back to `extract_mail_from(raw)` — which reads the `From:` header. This is **pre-existing behavior** (the old code also parsed the `From:` header in this case), so not a regression introduced by this PR. But per RFC 7208 §2.4, SPF for a null envelope sender should be verified against the HELO/EHLO domain, not the body `From:` header. The `select_spf_inputs()` doc comment says the fallback is for the "manual `aimx ingest` stdin path" — in reality it will also fire for bounce mail. Consider either (a) updating the comment to note this, or (b) filing a follow-up to thread `ehlo_hostname` through for the null-sender case. Not blocking for this PR since the goal is the Gmail-peer-IP bug, and bounce handling is a separable concern.

**2. Empty-string envelope semantics — Non-blocker.**
The `select_spf_inputs_empty_envelope_string_falls_back` test codifies that `Some("")` falls back to header parsing (`src/ingest.rs:1933-1943`). The test's own comment notes "should not happen — session resets reverse_path to `""` after RSET — but be explicit." Fine as defensive code, but this makes the empty-envelope contract double-guarded: `session.rs:501-505` already collapses empty → `None`, so `select_spf_inputs()` only ever sees `None` in practice. The second guard isn't harmful, just redundant. Keeping both is a reasonable belt-and-suspenders choice.

## Security Issues

No new security issues. The PR actually strengthens auth correctness:

- **SPF is now verifying the right thing.** Previously, a message from Gmail with a spoofed `Received:` header could have been SPF-checked against whatever the attacker wrote, since `extract_received_ip` walks all `Received:` lines and only trusts bracketed IPs — but headers below the topmost one are still attacker-controlled. Using `peer_addr` closes that gap.
- **DMARC mail-from alignment is now correct.** Previously, DMARC alignment compared the body-parsed `From:` domain to itself (since both came from headers) — effectively always-aligned, which is wrong. Using the envelope MAIL FROM for `rfc5321_mail_from_domain` makes alignment evaluation semantically correct per RFC 7489.
- `reverse_path` flows through `extract_angle_addr` which strips to the address, so no header injection risk via this path.

## Test Coverage

New tests are well-targeted:
- `select_spf_inputs_prefers_envelope_over_headers` — the core bug-fix assertion
- `select_spf_inputs_falls_back_to_headers_for_manual_stdin` — stdin path preserved
- `select_spf_inputs_empty_envelope_string_falls_back` — defensive edge case

**Coverage gap (Non-blocker):** there is no integration test that asserts the SPF *outcome* (e.g., `spf = "pass"` in the written frontmatter) when an SMTP session delivers a signed message with a real envelope. This would require either (a) a DNS mock for the `mail_auth::MessageAuthenticator` or (b) a test against a fixture domain with a live SPF record, both of which are beyond the scope of this fix. The existing `serve_e2e_*` integration tests at least exercise the new code path without asserting the auth outcome — which is enough to catch a crash but not enough to catch a silent regression. Worth filing as a follow-up to add a property-level test, but not a blocker.

No weak tests or always-passing assertions in the new test code.

## Code Quality

Well-scoped, small diff. Minor notes — all Non-blocker:

- **Pure helper extraction is the right call.** `select_spf_inputs()` as a standalone pure function makes the precedence logic directly testable without running a tokio runtime or DNS lookup. Good.
- **Doc comment on `select_spf_inputs()`** (`src/ingest.rs:311-316`) would benefit from explicitly mentioning that the fallback also fires for null MAIL FROM from inbound SMTP bounces, not *only* the stdin path (see Potential Bug #1).
- **Parameter-vs-struct choice** documented in PR body is the right call for a 30+ call-site fix — agreed, a struct would have been a bigger diff without meaningful ergonomics gain.
- **Mechanical test updates** (adding `None` at 25+ call sites) are visible in the diff but unavoidable. No semantic change to those tests.

- **Overall verdict:** Ready to merge
- **Blockers:** None
- **Non-blockers (consider before merge or as immediate follow-ups):**
  - Update `select_spf_inputs()` doc comment to acknowledge that the fallback also covers RFC 5321 null-sender bounces, not just the stdin path.
- **Nice-to-haves (file as follow-ups):**
  - Thread `ehlo_hostname` into `select_spf_inputs()` so null-sender bounces can be SPF-checked against the HELO domain per RFC 7208 §2.4.
  - Add an integration test that asserts SPF outcome end-to-end via the SMTP path (needs DNS mocking or a live fixture domain).

```
fix(ingest): use envelope MAIL FROM and peer IP for SPF verification

Previously, verify_auth() parsed the client IP from Received: headers
and mail_from from the body From: header. Both are attacker-influenced
and produced spf = "none" for legitimate mail delivered via Gmail where
the topmost Received: line has no bracketed IP.

Thread peer_addr and session_state.reverse_path from the SMTP session
through ingest_email() into verify_auth(), and use those authoritative
values for SPF (RFC 7208) and DMARC mail-from alignment (RFC 7489).
Header parsing is preserved as a best-effort fallback for the manual
`aimx ingest` stdin path, where no SMTP envelope exists.

Extracts select_spf_inputs() as a pure helper to make the envelope-vs-
header precedence unit-testable without a real SPF DNS lookup.
```

uzyn:
Thanks for the review. Addressed the one non-blocker and filed the nice-to-haves mentally as follow-ups (not in this PR).

## Changed in 30c000e

- **Doc comment on `select_spf_inputs()` (`src/ingest.rs:311-322`)** — rewritten to acknowledge that the header-parsing fallback fires for both the manual `aimx ingest` stdin path *and* RFC 5321 null-sender bounces (`MAIL FROM:<>`, where `reverse_path` is empty after `extract_angle_addr("<>")` → `session.rs:501` maps `""` → `None`). Also notes that the null-sender case is strictly non-conformant with RFC 7208 §2.4 and points at the follow-up (threading `ehlo_hostname` through).

## Intentionally deferred (as follow-ups, per review)

- Threading `ehlo_hostname` into `select_spf_inputs()` so null-sender bounces can be SPF-checked against the HELO domain per RFC 7208 §2.4.
- Integration test that asserts SPF *outcome* end-to-end (`spf = "pass"` in written frontmatter) via the SMTP path. Needs DNS mocking of `mail_auth::MessageAuthenticator` or a live-DNS fixture domain — both out of scope for this bugfix.

## Intentionally kept as-is

- **Empty-string envelope double-guard** (session collapsing `""` → `None` *and* `select_spf_inputs` also mapping `Some("")` → header fallback). You noted this is redundant-but-harmless defense-in-depth and fine to keep. Leaving it.
- **30+ mechanical `None` additions at test call sites.** All preserve prior semantics — the old callers had no envelope context, which is what `None` means. No substantive test change.

## Verification

- `cargo test select_spf_inputs` — 3 passed

uzyn (commented):
## Changes Overview

Independent re-verification after the prior self-review thread. PR threads the SMTP session's `peer_addr.ip()` and `session_state.reverse_path` through `ingest_email()` into `verify_auth()`, replacing the previous (wrong) practice of parsing the client IP from `Received:` headers and mail_from from the body `From:` header. A pure `select_spf_inputs()` helper encapsulates the envelope-vs-header precedence and is unit-tested. DMARC alignment also benefits since `rfc5321_mail_from_domain` now receives the envelope value per RFC 7489.

## Scope Alignment

All eight items in the PR's Stories Implemented checklist are present in `src/ingest.rs:31-38`, `src/ingest.rs:264`, `src/ingest.rs:311-339`, `src/ingest.rs:289-299`, `src/ingest.rs:397-405`, `src/smtp/session.rs:493-506`, `src/ingest.rs:22-29`, and `src/ingest.rs:1896-1944`. No scope creep — the `verify_dmarc_async` signature change is a natural byproduct and makes the mail_from source consistent between SPF and DMARC.

## Potential Bugs

None blocking. Verified independently:

- **`peer_ip.is_unspecified()` check at `src/ingest.rs:327`** — correctly covers both `0.0.0.0` and `::` sentinels; stdin path only uses `0.0.0.0` but the IPv6 case is also handled.
- **Empty-envelope guard at `src/ingest.rs:333-336`** — `Some("")` correctly falls back to header parsing. Matches the pattern in `src/smtp/session.rs:501-505` which already collapses empty → `None`. Double-guard is redundant-but-harmless (as noted in the prior thread).
- **Clone of `reverse_path` at `src/smtp/session.rs:498`** — small `String`, moved into the `spawn_blocking` closure, borrowed as `&str` within the closure. Lifetime-correct.
- **Null-sender bounce path** — `MAIL FROM:<>` produces `reverse_path = ""` (via `extract_angle_addr("<>")` → `""` at `src/smtp/session.rs:350-351`), which maps to `None` and falls back to parsing the body `From:` header. Strictly non-conformant with RFC 7208 §2.4 (should check HELO/EHLO domain) but doc comment at `src/ingest.rs:317-321` explicitly acknowledges this and the `ehlo_hostname` threading is filed as a follow-up. Non-blocker, consistent with pre-fix behavior.

## Security

The fix **improves** security correctness:

- **SPF IP source is now authoritative.** The previous `extract_received_ip(raw)` walked `Received:` headers, which below the topmost hop are attacker-controlled. Using `peer_addr` closes that gap.
- **DMARC mail-from alignment is now semantically correct.** Previously `rfc5321_mail_from_domain` was sourced from the body `From:` header, making alignment trivially self-true when both DKIM and body-From agreed (i.e. always, for DKIM-aligned mail). Now it uses the envelope MAIL FROM, which is what RFC 7489 actually requires.
- `reverse_path` flows through `extract_angle_addr`, so no header-injection vector via the envelope.

## Test Coverage

Three new unit tests in `src/ingest.rs:1896-1944` cover the three material cases of `select_spf_inputs`:
- envelope-wins (the core bug fix)
- stdin fallback (no session context)
- empty-envelope defensive fallback

All 648 unit + 53 integration tests pass locally; `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check` clean.

**Known coverage gap (non-blocker, filed as follow-up):** no end-to-end assertion that the written frontmatter shows `spf = "pass"` when a real signed message enters via the SMTP path. Requires either DNS mocking of `mail_auth::MessageAuthenticator` or a live-DNS fixture — both out of scope for this fix.

## Code Quality

- Helper extraction is the right call — `select_spf_inputs()` is a pure function, directly testable without a tokio runtime or DNS resolver. Good.
- Doc comment at `src/ingest.rs:311-321` is accurate after 30c000e: names both fallback triggers (stdin and null-sender) and links to the follow-up.
- Parameter vs struct trade-off — agreed with the PR body's reasoning given 30+ call sites; a struct would be a bigger diff without ergonomic gain at this scope.
- Mechanical `None` additions at test call sites — unavoidable, no semantic change.

## Summary

- **Overall verdict:** Ready to merge
- **Blockers:** none
- **Non-blockers:** none (the prior thread's one non-blocker — doc comment accuracy — was addressed in 30c000e)
- **Nice-to-haves (track as follow-ups, already acknowledged by author):**
  - Thread `ehlo_hostname` through so null-sender bounces can SPF-check against HELO per RFC 7208 §2.4
  - Integration test that asserts `spf = "pass"` in written frontmatter via the SMTP path

```
fix(ingest): use envelope MAIL FROM and peer IP for SPF verification

verify_auth() previously parsed the client IP from Received: headers
and mail_from from the body From: header. Both are attacker-influenced
and produced spf = "none" for legitimate mail delivered via Gmail where
the topmost Received: line has no bracketed IP.

Thread peer_addr and session_state.reverse_path from the SMTP session
through ingest_email() into verify_auth(), and use those authoritative
values for SPF (RFC 7208) and DMARC mail-from alignment (RFC 7489).
Header parsing remains a best-effort fallback for the manual `aimx
ingest` stdin path and for RFC 5321 null-sender bounces, with the
latter tracked as a follow-up to thread EHLO hostname through.

select_spf_inputs() is extracted as a pure helper so the envelope-vs-
header precedence is unit-testable without a real SPF DNS lookup.
```

PR #81, opened 2026-04-18
[Sprint 46] Mailbox CRUD via UDS (Daemon Picks Up Changes Live)

uzyn:
## Sprint Goal
Make `aimx mailbox create` / `delete` route through the daemon over UDS so the daemon's in-memory `Config` updates atomically with `config.toml` on disk — inbound mail to a just-created mailbox routes correctly on the very next SMTP session, no `systemctl restart aimx` required.

### [S46-1] UDS `MAILBOX-CREATE` — daemon writes config.toml + hot-swaps Config
- [x] `src/send_protocol.rs`: `MAILBOX-CREATE` verb parser + writer, plus new `VALIDATION` and `NONEMPTY` error codes
- [x] `src/mailbox_handler.rs` `handle_mailbox_create`: validates name, creates `inbox/<name>/` + `sent/<name>/`, atomically appends stanza via write-temp-then-rename, swaps `RwLock<Arc<Config>>`, returns `AIMX/1 OK`; typed ERR on validation / IO / duplicate
- [x] `src/mailbox.rs create`: UDS first, fallback to direct edit + restart-hint on socket-missing, hint suppressed on UDS success
- [x] Send handler / state handler / SMTP server refactored to read `Config` through the shared `ConfigHandle.load()` so every reader sees the swap
- [x] Integration test: daemon running → `aimx mailbox create eve` via UDS → inbound SMTP to `eve@domain` lands in `inbox/eve/`, not catchall, no restart
- [x] Integration test: daemon stopped → `aimx mailbox create eve` falls back to direct config edit + restart-hint present in stdout
- [x] Name-validation tests: empty, `..`, `/`, `\`, `\0` each return `ERR VALIDATION`; duplicate returns `ERR MAILBOX`
- [x] `book/mailboxes.md` updated — restart is no longer required when daemon is running
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all clean

### [S46-2] UDS `MAILBOX-DELETE` — safety check + daemon swap
- [x] `src/send_protocol.rs`: `MAILBOX-DELETE` verb parser + writer (shares `MailboxCrudRequest` with create, `create: false`)
- [x] `src/mailbox_handler.rs` `handle_mailbox_delete`: verifies mailbox exists, scans `inbox/<name>/` and `sent/<name>/` for files — if non-empty returns `ERR NONEMPTY mailbox <name> has <n> files (...); archive or remove them first`; on success removes stanza via write-temp-then-rename and swaps `Config`
- [x] `src/mailbox.rs delete`: attempts UDS first, falls back to direct edit + restart-hint only on socket-missing; NONEMPTY errors bubble up verbatim (not a socket-missing condition)
- [x] Refuses to delete `catchall` mailbox via UDS (matches CLI guardrail)
- [x] Integration test: daemon running → create `qux`, drop a file, `mailbox delete qux` returns `NONEMPTY`, remove file + retry succeeds, config.toml stanza gone, subsequent inbound to `qux@domain` falls through to catchall
- [x] `book/mailboxes.md` documents the NONEMPTY safety behavior and symmetric live-update semantics
- [x] Sprint 44 restart-hint suppression applies to the delete path too
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all clean

## Technical Decisions

- **`RwLock<Arc<Config>>` over `arc_swap::ArcSwap`.** Sprint notes called this a judgment call. Chose the std-library route to avoid a new dependency; the write-side is mailbox CRUD (rare), and the read side's critical section is a single `Arc::clone` — well under 1 ms even under heavy inbound pressure. If profiling ever shows ingest latency sensitive to this, the swap to `arc-swap` is a one-file change.
- **`ConfigHandle` shared by all daemon-side contexts.** Instead of threading a bunch of individual mutable references, `ConfigHandle` is cloneable (`Arc<RwLock<Arc<Config>>>`) and every context that needs Config reads it through `.load()` at request time. `SmtpSession` takes a fresh snapshot at `deliver_message` time so within one DATA payload every RCPT sees the same routing table; `SendContext` snapshots at the top of `handle_send` similarly; `StateContext` checks mailbox existence live for each MARK.
- **Disk-write-before-memory-swap ordering.** The handler writes `config.toml` atomically first, and only swaps the in-memory Config if the rename succeeds. If we did it the other way and the rename failed, the daemon would serve inbound mail against a routing table that doesn't match disk — exactly the silent-misroute bug Sprint 44's hint was papering over. On failure the handler also rolls back the freshly-created inbox/sent directories so the operator can retry cleanly.
- **`NONEMPTY` is not a socket-missing condition.** The CLI surfaces `NONEMPTY` verbatim rather than falling through to direct-edit (which would remove the directory tree and the files in it). This intentionally diverges from the old direct-edit CLI semantics on the "daemon running" path — matching the sprint spec, which treats the file safety as a feature.
- **Per-mailbox lock reuse.** MAILBOX-CRUD takes the same per-mailbox `RwLock<()>` that `StateContext` already owns for MARK verbs, so a concurrent MARK on the same mailbox can't observe a half-created / half-deleted file system. The lock map is lazy-inserted, so this has no startup cost.

## Deferred Items

None. Both stories are fully implemented.

## Review Focus Areas

- `src/mailbox_handler.rs` — the disk-write-then-swap ordering and the rollback-on-failure path are the critical-section logic. In particular `write_config_atomic`: make sure the tempfile naming (`<file>.tmp.<pid>`) plus the `File::create` + `sync_all` + `rename` sequence is bulletproof.
- `src/config.rs` `ConfigHandle` — correctness of the `Arc::clone`-under-read-lock pattern. No readers should ever hold the `RwLock` across an `.await`.
- `src/smtp/session.rs` `deliver_message` — the `config_handle.load()` call now happens before the `for rcpt in forward_paths` loop; all recipients in one DATA payload still see the same snapshot.
- `src/mcp.rs` `submit_mailbox_crud_via_daemon` + `MailboxCrudFallback` — the fallback decision is based solely on I/O error kind (`NotFound`/`ConnectionRefused`/`PermissionDenied`), never on a daemon-returned ERR. That's what keeps `NONEMPTY` from accidentally turning into a destructive direct-edit delete.
- `tests/integration.rs` `McpClient::spawn` — now sets `AIMX_RUNTIME_DIR` to a per-test tempdir so two existing MCP tests don't hit whatever production daemon happens to be on the CI host. Without this, `mcp_mailbox_create_list_delete` silently spoke to a prod daemon and the failure mode was confusing.


uzyn (commented):
# Sprint 46 Review — Mailbox CRUD via UDS

## Sprint Goal Assessment

The goal is fully achieved: `aimx mailbox create`/`delete` now route through the daemon's UDS socket, atomically rewrite `config.toml`, and hot-swap the in-memory `Arc<Config>` so inbound mail to a just-created mailbox routes correctly on the very next SMTP session. Integration tests verify both the live-swap happy path and the no-restart routing claim end-to-end. The CI on this commit (`3ab73a1`) is green (core-tests pass 4m46s, verifier-tests pass 1m54s), and a full local run reproduces `725 passed` (bin) + `59 passed` (integration) on a single host — matching the implementer's note about the earlier intermediate `McpClient::spawn` flake having been fixed by the `AIMX_RUNTIME_DIR` override.

**[S46-1] UDS `MAILBOX-CREATE` — daemon writes config.toml + hot-swaps Config**
- [x] `send_protocol.rs` gains `MAILBOX-CREATE` verb parser/writer and `VALIDATION`/`NONEMPTY` error codes (well-tested, 9 new send_protocol unit tests, including case-insensitive headers, duplicate/empty/missing Name, nonzero Content-Length rejection, roundtrip, verb-selection-in-wire-format)
- [x] `mailbox_handler::handle_mailbox_crud` validates name, creates `inbox/<name>/` + `sent/<name>/`, atomically rewrites config.toml via write-temp-then-rename, swaps the `RwLock<Arc<Config>>` only after rename succeeds, and rolls back the freshly-created dirs on rename failure
- [x] `mailbox::create` tries UDS first; on `NotFound`/`ConnectionRefused`/`PermissionDenied` falls back to direct-edit + restart hint; suppresses the hint on UDS success
- [x] Send handler, state handler, SMTP server, and the new mailbox handler all read `Config` through `ConfigHandle.load()` so the swap is visible everywhere. Verified `SmtpSession::deliver_message` snapshots once per DATA so every RCPT within a single message sees the same routing table
- [x] Integration test: daemon running → `aimx mailbox create eve` via UDS → inbound SMTP to `eve@agent.example.com` lands in `inbox/eve/`, not catchall, no restart
- [x] Integration test: daemon stopped → `aimx mailbox create eve` falls back to direct config edit + restart-hint present in stdout
- [x] Name-validation tests: empty, `..`, `/`, `\\`, `\\0` each return `ERR VALIDATION`; duplicate returns `ERR MAILBOX`
- [x] `book/mailboxes.md` updated with the live-swap behavior and NONEMPTY semantics
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all clean locally and on CI
- [ ] Concurrent `MAILBOX-CREATE foo` + inbound on a different mailbox — NOT MET: no explicit test, but the per-mailbox lock plus the snapshot-then-ingest flow in `SmtpSession::deliver_message` make this structurally true. Minor gap, not blocking.

**[S46-2] UDS `MAILBOX-DELETE` — safety check + daemon swap**
- [x] `MAILBOX-DELETE` verb parser/writer (shares `MailboxCrudRequest` with `create: false`)
- [x] `handle_mailbox_delete` verifies existence, scans inbox+sent, returns `ERR NONEMPTY` with file count, removes the stanza atomically, swaps Config, leaves directories on disk (operator cleans up)
- [x] `mailbox::delete` uses UDS-first; surfaces `NONEMPTY` verbatim (never falls back to destructive direct-edit); socket-missing → direct-edit + restart hint
- [x] Catchall delete refused explicitly via `name == \"catchall\"` guard in `handle_delete`, matching CLI semantics
- [x] Integration test: create `qux`, drop a stray file, `aimx mailbox delete --yes qux` fails with NONEMPTY + stanza preserved; remove the file, retry, succeeds; subsequent SMTP to `qux@domain` falls through to catchall via the live-swapped handle
- [x] `book/mailboxes.md` documents NONEMPTY behavior and symmetric live-update semantics
- [x] Sprint 44 restart-hint suppression applies to delete path too

## Test Coverage

Test coverage is thorough and mostly targets the right seams:

- 9 `send_protocol` unit tests covering verb roundtrip, case-insensitive headers, duplicate/empty/missing Name, nonzero Content-Length rejection, and explicit verb-selection-in-wire-format
- 11 `mailbox_handler` unit tests covering create/delete happy paths, duplicate/missing/invalid-name error paths, NONEMPTY rejection, catchall guardrail, disk-write failure rollback (uses a nonexistent parent dir to force failure — nice), atomic rename leaves no stray `.tmp.*`, and a create-then-MARK chain that verifies the live-swap works in-process
- 3 integration tests exercising the end-to-end CLI + daemon + SMTP ingest flow
- Pre-existing `mcp_mailbox_create_list_delete` / `mcp_mailbox_delete_catchall_error` updated so `McpClient::spawn` sets `AIMX_RUNTIME_DIR` — fixing a real hazard where the MCP process would silently speak to a production daemon on the test host

**Gaps worth noting (non-blocking):**

- No explicit test for concurrent `MAILBOX-CREATE` on **different** mailbox names (see Potential Bugs §1). The per-mailbox lock design permits cross-name parallelism, but nothing validates the outcome.
- `create_failure_at_disk_write_leaves_handle_and_disk_unchanged` forces failure via a nonexistent parent dir. That path fails at `File::create`, so it never exercises the rename step's failure mode. A test using a read-only target would be stronger — but not required.
- No test exercises the MCP tool's fallback path (`MailboxCrudFallback::SocketMissing` → `create_mailbox` with \"daemon not running\" message). The integration tests cover the CLI fallback path; the MCP one is only covered by unit tests on `submit_mailbox_crud_via_daemon`. Minor.

## Potential Bugs

### 1. Lost-update race on `config.toml` across different mailbox names — Non-blocker

`mailbox_handler::handle_mailbox_crud` takes a per-mailbox lock keyed by `req.name`. Two concurrent `MAILBOX-CREATE alice` + `MAILBOX-CREATE bob` requests hold different locks and can interleave:

1. A: `load()` → `config_v1` (no alice, no bob)
2. B: `load()` → `config_v1` (no alice, no bob)
3. A: builds `config_v1 + alice`
4. B: builds `config_v1 + bob`
5. A: `write_config_atomic` renames `config_v1 + alice` over `config.toml`
6. B: `write_config_atomic` renames `config_v1 + bob` over `config.toml` — alice's stanza now gone from disk
7. A: `store(config_v1 + alice)` — in-memory has alice but not bob
8. B: `store(config_v1 + bob)` — in-memory has bob but not alice

End state: disk has only bob; memory has only bob; alice's inbox+sent dirs exist orphaned. Both clients saw `OK`. Subsequent SMTP to `alice@domain` silently falls through to catchall.

This is a real correctness bug, but v0.2 has a single operator manually invoking `aimx mailbox create`, so the window is vanishingly small in practice. MCP-driven concurrent agent calls are the realistic trigger. **A process-wide `Mutex` around the load → modify → write → store critical section would close it in ~6 lines** (symmetric to `SENT_WRITE_LOCK` in `send_handler.rs`). Worth filing into the Non-blocking Review Backlog and addressing before production-scale multi-agent use.

### 2. `write_config_atomic` loses non-canonical TOML content — Pre-existing, Non-blocker

Re-serializing `Config` drops unknown fields (e.g., legacy `probe_url`, `verify_address`) and erases TOML comments. This is preexisting — the old `Config::save` path has the same behavior — so Sprint 46 doesn't regress it. Flagging so it's not forgotten.

### 3. Rollback `remove_dir` (non-recursive) vs pre-existing directories — Non-blocker

`handle_create` rolls back with `remove_dir(&inbox)` / `remove_dir(&sent)`. If the operator had manually created `inbox/<name>/` (empty) before running create, the rollback will remove their directory. If it had content, `remove_dir` silently fails (best-effort) and the dir stays — an `OK_ = std::fs::remove_dir(…)` result, which is acceptable. Not a bug, just a subtle interaction worth documenting.

### 4. `MCP mailbox_delete` success message omits \"directories remain on disk\" note — Non-blocker

The CLI delete success is simply `Mailbox 'name' deleted.`. Users following the symmetric-delete expectation might `ls` and be surprised by the remaining empty directories. Consider a one-liner hint (\"(empty directories left; run `rmdir` if you want to tidy up)\"). Minor UX.

### 5. `validate_mailbox_name` doesn't reject whitespace / RFC-5322-unsafe local parts — Pre-existing, Non-blocker

Names like `\"hello world\"` pass validation but produce invalid email addresses when interpolated as `{name}@{domain}`. Preexisting from `src/mailbox.rs validate_mailbox_name`; the daemon-side copy at `mailbox_handler.rs:88` duplicates the same rules intentionally.

## Security Issues

None found. The daemon-side `validate_mailbox_name` enforces the same rules as the CLI (empty / `..` / path separator / NUL), preventing a malicious local client from slipping a traversal past the CLI guard. The `is_socket_missing` logic correctly treats `PermissionDenied` as a fallback condition — slightly aggressive but safe given the `0o666` socket and v0.2's explicit \"authorization out of scope\" stance.

## Code Quality

A few minor observations (all Non-blocker):

- `submit_mailbox_crud_via_daemon` in `src/mcp.rs` returns `MailboxCrudFallback`, but reuses the `MarkOutcome` enum internally (`Ok(MarkOutcome::Ok)` etc). Correct but slightly confusing — a distinct `AckOutcome` (or just inlining) would read cleaner. Not worth changing unless you're refactoring anyway.
- `MailboxContext` duplicates the `config_handle` that's also in `StateContext`. Fine — `ConfigHandle` is a cheap `Arc` clone — but the doc comment on `MailboxContext` says it's \"kept separate from StateContext so the state-lock map is not accidentally bypassed\", which is true but the actual separation is two arguments on `handle_mailbox_crud`. Fine as-is.
- The `state_ctx.data_dir` + `mb_ctx.config_handle.load().data_dir` pair could theoretically diverge if the handle were ever swapped to a new data_dir. Today the Config's `data_dir` is never changed after startup so this is moot; a comment would help the next reader.
- `StateContext` doc mentions \"Sprint 46 is tracked to unify [MARK and ingest] writers under a single per-mailbox lock\" — Sprint 46 didn't actually unify them (reasonably, since it's out of scope for MAILBOX-CRUD). Worth updating the comment to point at a future sprint instead.

## Alignment with PRD

FR-26 (\"No mail server restart required\") is now fully satisfied. The PRD's other mailbox FRs (FR-27, FR-28) are preserved. No divergence from PRD intent.

**Overall verdict: Needs minor fixes** — the lost-update race (§1) is a real but low-probability correctness bug. Everything else is high-quality, well-tested, green on CI.

**Blockers:** None.

**Non-blockers (worth filing into Non-blocking Review Backlog and addressing before multi-agent production use):**
1. Process-wide lock on the config.toml read-modify-write critical section to close the lost-update race on concurrent `MAILBOX-CREATE`/`DELETE` across different mailbox names
2. Test coverage gap: concurrent `MAILBOX-CREATE` on different names
3. MCP `mailbox_delete` success message could mention the leftover empty directories

**Nice-to-haves:**
- Stronger disk-write failure test (read-only target, not nonexistent parent)
- Update the Sprint-46-era comment in `StateContext` pointing at a future lock-unification sprint
- Brief comment on the `state_ctx.data_dir` vs `config.data_dir` invariant

Ready to merge once the race is either (a) fixed in a follow-up commit on this PR, or (b) explicitly triaged into the backlog with the understanding that v0.2 ships without it.

uzyn:
## Cycle-2 review fixes (commit 20502d0)

Addressed all flagged items from the cycle-1 review:

### Fixed in code

1. **Lost-update race on `config.toml` (Potential Bugs §1)** — `src/mailbox_handler.rs` now takes a process-wide `CONFIG_WRITE_LOCK` (`std::sync::Mutex`) around the load-modify-write-store critical section, after the per-mailbox `RwLock` and before any disk I/O. Symmetric in spirit to `send_handler::SENT_WRITE_LOCK`. Held only for the critical section so MARK-* and inbound ingest on other mailboxes still proceed in parallel.
2. **Test coverage gap on concurrent different names** — added `concurrent_create_different_names_both_stanzas_survive` in `src/mailbox_handler.rs`. Fans out 16 concurrent `MAILBOX-CREATE` calls across distinct names on a multi-thread tokio runtime, then asserts every stanza survives both on disk and in the live `ConfigHandle`, and every `inbox/<name>/` + `sent/<name>/` pair exists. Verified the test fails without the global lock and passes with it.
3. **MCP `mailbox_delete` success message (Potential Bugs §4)** — both the MCP tool (`src/mcp.rs`) and the CLI (`src/mailbox.rs`) now mention the leftover empty directories and suggest `rmdir` for cleanup. `agents/common/references/mcp-tools.md` updated to match.
4. **Stale `Sprint 46 will unify` comment in `StateContext` (Code Quality nit)** — `src/state_handler.rs` doc comments now point at the Non-blocking Review Backlog rather than Sprint 46.
5. **`state_ctx.data_dir` vs `config.data_dir` invariant (Code Quality nit)** — added a doc comment on `StateContext::data_dir` documenting the snapshot-at-startup invariant.

### Filed into Non-blocking Review Backlog (`docs/sprint.md`)

- **(Potential Bugs §2)** `write_config_atomic` drops unknown TOML fields and erases comments. Pre-existing, symmetric to `Config::save`. Needs a TOML-edit crate or pass-through serde rep — not trivial, deferred.
- **(Potential Bugs §5)** `validate_mailbox_name` accepts whitespace / RFC-5322-unsafe local parts. Pre-existing in `src/mailbox.rs` and intentionally duplicated in `src/mailbox_handler.rs` — both need tightening to a local-part safe character class.
- **(Nice-to-have, Test Coverage §2)** Disk-write failure test forces failure via a nonexistent parent dir, which fails at `File::create` before reaching `rename(2)`. Strengthen with a read-only target.
- Surfaced the MARK-* + ingest unify-under-one-lock follow-up that the `StateContext` comment used to point at Sprint 46.

### Intentionally left as-is

- **Potential Bugs §3 (`remove_dir` rollback vs pre-existing dirs)** — reviewer explicitly notes this is acceptable best-effort behavior, not a bug. No change.
- **Code Quality nit on `MarkOutcome` reuse in `submit_mailbox_crud_via_daemon`** — reviewer says "not worth changing unless you're refactoring anyway". Skipped.
- **Code Quality nit on `MailboxContext` doc** — reviewer says "fine as-is". Skipped.

### Verification

- `cargo test`: 726 unit + 59 integration passed (was 725 + 59; +1 new concurrent-create test)
- `services/verifier`: 43 passed

uzyn (commented):
# Sprint 46 Re-Review (Cycle 2) — PR #81

Verified commit `20502d0` against the cycle-1 review. All five flagged items are addressed in code; the four nice-to-haves are filed into the Non-blocking Review Backlog with `(Sprint 46, PR #81)` tags as agreed.

## Resolved Issues

1. **Lost-update race on `config.toml` (Potential Bugs §1) — RESOLVED.** `src/mailbox_handler.rs` now declares `static CONFIG_WRITE_LOCK: Mutex<()>` and acquires it inside `handle_mailbox_crud` between the per-mailbox `RwLock` and the disk I/O. Lock order is clean: tokio per-mailbox `RwLock` (outer) → `std::sync::Mutex` (inner) → filesystem ops. No other call site takes `CONFIG_WRITE_LOCK`, so nested-lock cycles aren't reachable. Held only for the critical section, so MARK-* and inbound ingest on other mailboxes still proceed in parallel as the design intended. Poison handling via `unwrap_or_else(|p| p.into_inner())` is the right call — a panic inside the critical section shouldn't permanently wedge MAILBOX-CRUD. The doc comment on the static and the inline comment at the acquisition site both clearly explain why the global lock exists.

2. **Concurrent-create test gap — RESOLVED.** New `concurrent_create_different_names_both_stanzas_survive` (multi-thread tokio runtime, 16 distinct mailbox names spawned via `tokio::spawn`) asserts every stanza survives both on disk (`Config::load`) and in the live `ConfigHandle`, plus every `inbox/<name>/` + `sent/<name>/` directory pair exists. The fan-out is wide enough to defeat accidental scheduler serialization. Test passes locally on the new code; the implementer's note ("verified the test fails without the global lock") matches what we'd expect.

3. **MCP `mailbox_delete` UX (Potential Bugs §4) — RESOLVED.** `src/mcp.rs` and `src/mailbox.rs` both now mention the leftover `inbox/<name>/` + `sent/<name>/` directories in the success message and suggest `rmdir`. `agents/common/references/mcp-tools.md` is updated symmetrically and additionally documents the NONEMPTY semantics + catchall guardrail under **Errors** — a nice extra.

4. **Stale `Sprint 46 will unify` comment (Code Quality nit) — RESOLVED.** Both occurrences in `src/state_handler.rs` (header and `handle_mark` doc) now point at the Non-blocking Review Backlog instead of Sprint 46.

5. **`state_ctx.data_dir` vs `config.data_dir` invariant (Code Quality nit) — RESOLVED.** A clear doc comment on `StateContext::data_dir` documents the snapshot-at-startup invariant and explains why MAILBOX-CRUD never rewrites `data_dir`.

## Non-blocking Review Backlog Triage

Verified the four deferred items landed in `docs/sprint.md` with the correct tags:

- `write_config_atomic` drops unknown TOML fields and erases comments — tagged `(Sprint 46, PR #81)`
- `validate_mailbox_name` accepts whitespace / RFC-5322-unsafe local parts — tagged `(Sprint 46, PR #81)`
- Disk-write failure test forces failure at `File::create`, never reaching `rename(2)` — tagged `(Sprint 46, PR #81)`
- MARK-* + ingest unify-under-one-lock — tagged `(Sprint 45, PR #78 → Sprint 46, PR #81)`

All four are acceptable to defer for v0.2.

## New Issues Found

None. Diff between cycle-1 and cycle-2 commits is small (`+129 / -10` across 6 files) and well-scoped. No regressions introduced. `cargo clippy --all-targets -- -D warnings` is clean locally on the new HEAD.

## Verification Notes

- Reran `cargo test --bin aimx mailbox_handler::tests::concurrent_create_different_names_both_stanzas_survive` — passes (0.03s)
- CI on `20502d0` is pending at the time of this review; the implementer's note confirms `cargo test`, `cargo clippy`, and `cargo fmt -- --check` are all clean locally (726 unit + 59 integration + 43 verifier)
- Lock-ordering audit: `CONFIG_WRITE_LOCK` is only taken in `handle_mailbox_crud`. No other call site touches it, so no AB-BA potential with `SENT_WRITE_LOCK`, `INGEST_WRITE_LOCK`, or the per-mailbox RwLock map

## Verdict

**APPROVED — Ready to merge.**

Sprint 46 delivers the live-swap mailbox CRUD cleanly. The cycle-2 fix closes the only real correctness concern from cycle 1, the new regression test would catch the bug if it ever came back, and the operator-facing UX gap on delete is addressed in both surfaces (CLI + MCP) plus the agent-facing reference doc. Pending CI run is a formality given the local clean run on the same commit.

```
[Sprint 46] Mailbox CRUD via UDS (Daemon Picks Up Changes Live) (#81)

Adds MAILBOX-CREATE / MAILBOX-DELETE verbs to the AIMX/1 UDS protocol so
`aimx mailbox create/delete` route through the daemon. The daemon
atomically rewrites `config.toml` and hot-swaps its in-memory
`Arc<Config>` so inbound mail to a just-created mailbox is delivered to
the correct directory on the very next SMTP session — no
`systemctl restart aimx` required. When the daemon is stopped, the CLI
falls back to direct on-disk edit and prints the Sprint 44 restart-hint
banner. Delete refuses non-empty mailboxes (NONEMPTY), refuses the
catchall guardrail, and leaves empty directories on disk for the
operator to clean up.

Concurrent MAILBOX-CREATE/DELETE across different mailbox names is
serialized by a process-wide CONFIG_WRITE_LOCK, closing a lost-update
race where two requests holding disjoint per-mailbox locks could clobber
one stanza via interleaved load → modify → write → store sequences. A
multi-thread regression test (16 concurrent creates) verifies the fix.

CLI and MCP delete success messages mention the leftover empty
directories so operators aren't surprised; agents/common/references/
mcp-tools.md updated to document NONEMPTY, catchall guardrail, and
leftover-directory behavior.
```

PR #82, opened 2026-04-18
[Sprint 47] Post-v1 Non-blocking Cleanup

uzyn:
## Sprint Goal
Close the eight non-blocking improvements accumulated across Sprints 44-46 reviews — defense-in-depth, test-coverage, and architectural-cleanup work. No user-visible features; no PRD changes.

### [S47-1] DKIM startup check — end-to-end integration test + runtime-flavor contract
- [x] `HickoryDkimResolver::resolve_dkim_txt` now `debug_assert!`s the tokio runtime flavor is `MultiThread` at entry (its `block_in_place` + `Handle::block_on` call graph requires it). Short doc comment explains the rationale.
- [x] New `AIMX_TEST_DKIM_RESOLVER_OVERRIDE` env var (test-only, never set in prod) short-circuits the real DNS lookup so integration tests can drive `run_serve` end-to-end against a canned resolver result (`ok:…`, `err:…`, or `no-record`).
- [x] Two integration tests spin `aimx serve` against the override:
  - `dkim_startup_check_mismatch_logs_warning_and_binds_listeners` — Mismatch emits the multi-line ERROR message and both listeners still bind.
  - `dkim_startup_check_resolve_error_logs_warning_and_binds_listeners` — ResolveError emits a Warning-level message and both listeners still bind.
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all clean.

### [S47-2] Exhaustiveness & defense-in-depth hardening
- [x] `restart_hint_command` replaces the `_` fallback with an explicit `InitSystem::{Systemd, OpenRC, Unknown}` match — adding a new variant now fails to compile until the arm is supplied.
- [x] New `restart_hint_command_is_exhaustive_over_init_system` test destructures every current variant, validating the exhaustive check at compile time.
- [x] `channel::execute_triggers` calls `.env_clear()` before selectively re-adding `PATH`, `HOME`, and the `AIMX_*` vars. Inline comment explains why.
- [x] New `execute_triggers_clears_parent_env` test sets a sentinel env var on the parent and asserts it does NOT appear in the trigger subshell (while `PATH` is preserved).
- [x] `cargo test`, `cargo clippy`, `cargo fmt -- --check` all clean.

### [S47-3] Validation tightening + TOML-rewrite preservation + stronger rename-failure test
- [x] `validate_mailbox_name` consolidated to a single canonical helper in `src/mailbox.rs`; the duplicate in `src/mailbox_handler.rs` now calls through. Tightened to `[a-z0-9._-]+` (case-folded, no whitespace, no leading/trailing `.`, no `..`). Closes the hole where `"hello world"` made it past validation and produced an invalid `<name>@<domain>` address.
- [x] New unit tests cover `"hello world"`, `"a b"`, `"..foo"`, `""`, `"Alice"`, `"alice+bob"`, `".leading"`, `"trailing."` (all rejected) and `"good-mailbox.1"`, `"catchall"`, `"alice"`, `"a.b_c-1"` (all accepted).
- [x] `write_config_atomic`: chose the document-current-behaviour path (lower-risk for v1). The doc comment now spells out that unknown TOML stanzas and comments are erased on rewrite; `unknown_stanza_is_dropped_on_rewrite` pins the behaviour. `toml_edit` adoption deferred to v2.
- [x] `create_failure_at_disk_write_leaves_handle_and_disk_unchanged` rewritten: the temp write now succeeds and the `rename(2)` fails because the target path is a non-empty directory (forces `EISDIR`/`ENOTEMPTY`). Previous form tripped `File::create` before the temp write started, so the rollback branch was never exercised.
- [x] `cargo test`, `cargo clippy`, `cargo fmt -- --check` all clean.

### [S47-4] Unify ingest + MARK writers under one per-mailbox lock
- [x] New `src/mailbox_locks.rs` owns `MailboxLocks` — a shared per-mailbox `Arc<tokio::sync::Mutex<()>>` map keyed by mailbox name, exposed via the `Arc<Mutex<HashMap<String, Arc<AsyncMutex<()>>>>>` accessor pattern.
- [x] `ingest_email` now takes `&MailboxLocks` and acquires the per-mailbox lock via `blocking_lock()` for its filename-allocation + bundle-write + `.md`-write critical section. Replaces the process-wide `INGEST_WRITE_LOCK`.
- [x] `StateContext` holds `Arc<MailboxLocks>` instead of its own `RwLock` map; `handle_mark` takes the same per-mailbox lock. `MailboxContext`'s handler already routed through `state_ctx.lock_for(...)` — unchanged at the call site.
- [x] `aimx serve` constructs one `MailboxLocks` and threads it to the SMTP server (via `with_mailbox_locks`), `StateContext`, and `MailboxContext` so inbound ingest, MARK-*, and MAILBOX-* all serialize on the same lock per mailbox.
- [x] Module-level comment in `src/state_handler.rs` updated; new full explanation lives in `src/mailbox_locks.rs` covering the lock hierarchy (**outer**: per-mailbox tokio Mutex; **inner**: process-wide `CONFIG_WRITE_LOCK`; always outer → inner).
- [x] New integration test `concurrent_ingest_burst_and_mark_same_mailbox_no_torn_writes`: 6 concurrent SMTP ingests + 2 concurrent MARK-READ calls on a seeded mailbox; verifies every `.md` file has intact `+++ … +++` frontmatter that parses as TOML.
- [x] New integration test `concurrent_mailbox_create_and_ingest_does_not_deadlock`: races `MAILBOX-CREATE` + ingest addressed to the same name under a 20-second watchdog; verifies no deadlock, the message lands consistently (either in the new mailbox or catchall, never lost or torn), and a follow-up message still delivers after the race (daemon remains responsive).
- [x] `cargo test`, `cargo clippy`, `cargo fmt -- --check` all clean.

## Technical Decisions

- **`debug_assert!` over async-ifying `DkimTxtResolver`.** Async-ification would cascade into `run_dkim_startup_check` and every mock resolver in the test suite. The debug-assert documents the contract in one place, fires in test builds, and leaves the zero-cost production path unchanged. If a future caller needs the current-thread flavor we can flip to async then.
- **`tokio::sync::Mutex` (not `RwLock`) for the unified per-mailbox lock.** The spec prescribed `Mutex`, and writes dominate every user on this lock (ingest always writes; MARK always rewrites; MAILBOX-* always rewrites). A reader/writer split would buy nothing and complicate the API surface (`.write().await` vs `.lock().await`).
- **`blocking_lock()` for the ingest call path.** `ingest_email` runs from a sync context (manual stdin or SMTP session's `spawn_blocking` worker). `tokio::sync::Mutex::blocking_lock()` works from a blocking thread and never needs an async context. This avoids making `ingest_email` `async`, which would require refactoring the stdin path and every test (30+ call sites).
- **Lock hierarchy: per-mailbox outer, `CONFIG_WRITE_LOCK` inner.** Inverting would deadlock: two `MAILBOX-CREATE` calls on different names would both acquire the config lock, then race for their per-mailbox locks. The config write is short and bounded, so holding it while a longer ingest critical section waits on the outer lock is also bad (priority inversion).
- **`write_config_atomic`: document-not-fix for v1.** Adopting `toml_edit` needs a Cargo.toml dep + `Config` serialization rewrite + round-trip testing. For v1 (machine-authored `config.toml`, only `aimx setup` / `aimx mailbox *` edit it), the simpler path is to pin the current behaviour with a test and flag it for v2.
- **Rename-failure test uses a non-empty directory as the rename target.** Simplest way to force `rename(2)` to fail after `File::create` succeeds, cross-platform between Linux and macOS (`EISDIR`/`ENOTEMPTY`).

## Deferred Items
- **`toml_edit` adoption** — documented-behaviour path taken for v1; pinning test guards the current behaviour; revisit in v2 alongside `Config::save` which has the same limitation.

## Review Focus Areas
- `src/mailbox_locks.rs` — the new shared-lock type and its documented hierarchy. The soundness argument for moving from "disjoint-file invariant" to "unified lock" lives here.
- `src/ingest.rs` — `ingest_email` signature change ripples through session.rs and ~30 test call sites. Please confirm the `blocking_lock()` placement is correct for all three callers (stdin, SMTP `spawn_blocking`, tests).
- `src/state_handler.rs` module doc — the updated lock-hierarchy explanation. Flag any ambiguity that could mislead a future reader.
- `tests/integration.rs` — the two new concurrency tests plus the DKIM startup tests. The watchdog in the deadlock test is 20s (tests normally complete in <1s); flag if you want that tighter.
- `src/serve.rs` — the `AIMX_TEST_DKIM_RESOLVER_OVERRIDE` test-only hatch. Please confirm the env-var gating is tight enough (it only triggers when the var is set, and the production daemon never sets it).

### DIFF
--- a/docs/sprint.md
+++ b/docs/sprint.md
@@ -433,7 +433,7 @@ or
 
 - [ ] Consolidate `validate_mailbox_name` into a single canonical helper (one of `src/mailbox.rs`, `src/mailbox_handler.rs`, or `src/config.rs` — whichever already owns mailbox-name invariants), tighten to reject whitespace and any character outside a safe local-part class (`[a-z0-9._-]`, case-folded), and thread both call sites through it
 - [ ] New unit tests cover: `"hello world"` rejected, `"a b"` rejected, `"..foo"` rejected, `""` rejected, `"good-mailbox.1"` accepted
-- [ ] `write_config_atomic`: either adopt a TOML-editing crate (e.g. `toml_edit`) that preserves comments and unknown stanzas, OR document the current behavior with a comment + add a test asserting a known-unknown stanza survives (whichever the implementer judges less risky for v1)
+- [ ] `write_config_atomic`: either adopt a TOML-editing crate (e.g. `toml_edit`) that preserves comments and unknown stanzas, OR document the current behavior with a comment + add a test asserting a known-unknown stanza survives (whichever the implementer judges less risky for v1) <!-- Chose the document-current-behaviour path. The "stanza survives" wording conflates the two paths; on the documented path the stanza is dropped (v1 contract). Test `unknown_stanza_is_dropped_on_rewrite` pins this so any future regression toward a preserving editor without updating the doc comment will trip. -->
 - [ ] `create_failure_at_disk_write_leaves_handle_and_disk_unchanged`: rewrite to make the temp write succeed and the rename fail (read-only target dir on the parent, or equivalent), so the test genuinely exercises the rollback-on-rename-failure branch; assertion remains: disk + handle unchanged
 - [ ] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean
 
--- a/docs/sprint.md
+++ b/docs/sprint.md
@@ -168,7 +168,7 @@ Fix: `sanitize_attachment_filename(raw: &str, index: usize) -> String` — NFC-n
 
 ---
 
-## Sprint 44 — Post-launch Security + Quick Fixes (Days 123–125.5) [NOT STARTED]
+## Sprint 44 — Post-launch Security + Quick Fixes (Days 123–125.5) [DONE]
 
 **Goal:** Close the four highest-priority findings from the 2026-04-17 manual test run with small, targeted patches: shell-injection fix in channel triggers (security), operator-visible DKIM sanity check at daemon startup, corrected Claude Code plugin hint, and a restart-hint on `aimx mailbox create`. Also fix the docs nit that caused forwarded-message noise in the test log. Finding #2 (SPF envelope MAIL FROM) already shipped in commit `cd22428` and is excluded. Finding #10 is mostly an operator-side DNS republish; only its two small code add-ons (startup DKIM sanity check + louder setup warning) are in scope here.
 
@@ -184,14 +184,15 @@ Fix: `sanitize_attachment_filename(raw: &str, index: usize) -> String` — NFC-n
 
 **Priority:** P0 (security)
 
-- [ ] `src/channel.rs`: `substitute_template` rewritten to only expand `{id}` and `{date}`; `shell_escape_value` deleted
-- [ ] Command spawn point uses `Command::new("sh").arg("-c").arg(&script).env("AIMX_FROM", …).env("AIMX_SUBJECT", …).env("AIMX_TO", …).env("AIMX_MAILBOX", …).env("AIMX_FILEPATH", …)` — every user-controlled field goes via env
-- [ ] Config loader rejects any `on_receive.cmd` containing `{from}`/`{subject}`/`{to}`/`{mailbox}`/`{filepath}` with an error naming the offending mailbox + the env-var migration
-- [ ] `book/channel-recipes.md` rewritten to use `"$AIMX_FROM"`, `"$AIMX_SUBJECT"`, etc. for every recipe (all agents + the shell-log example)
-- [ ] `docs/manual-test.md` T8 recipe updated to the env-var pattern
-- [ ] New unit tests covering injection attempts: `U-Zyn Chua <chua@uzyn.com>` (angle-bracket redirect, the T8 repro), `` `whoami` ``, `$(rm -rf /)`, `foo; ls`, `foo\nbar`, subject with embedded single/double quotes — all must run the intended command with the payload safely landing in the env var
-- [ ] New unit test: config with a legacy placeholder in `on_receive.cmd` fails to load with the migration error
-- [ ] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean
+- [x] `src/channel.rs`: `substitute_template` rewritten to only expand `{id}` and `{date}`; `shell_escape_value` deleted
+- [x] Command spawn point uses `Command::new("sh").arg("-c").arg(&script).env("AIMX_FROM", …).env("AIMX_SUBJECT", …).env("AIMX_TO", …).env("AIMX_MAILBOX", …).env("AIMX_FILEPATH", …)` — every user-controlled field goes via env
+- [x] Config loader rejects any `on_receive.cmd` containing `{from}`/`{subject}`/`{to}`/`{mailbox}`/`{filepath}` with an error naming the offending mailbox + the env-var migration
+- [x] `book/channel-recipes.md` rewritten to use `"$AIMX_FROM"`, `"$AIMX_SUBJECT"`, etc. for every recipe (all agents + the shell-log example)
+- [x] `docs/manual-test.md` T8 recipe updated to the env-var pattern
+- [x] New unit tests covering injection attempts: `U-Zyn Chua <chua@uzyn.com>` (angle-bracket redirect, the T8 repro), `` `whoami` ``, `$(rm -rf /)`, `foo; ls`, `foo\nbar`, subject with embedded single/double quotes — all must run the intended command with the payload safely landing in the env var
+- [x] New unit test: config with a legacy placeholder in `on_receive.cmd` fails to load with the migration error
+- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean
+- [x] Extra: `shell-escape` crate removed from `Cargo.toml`; `book/channels.md`, `book/configuration.md`, `docs/manual-setup.md`, `docs/prd.md` FR-30, `docs/idea.md` all swept; integration test `ingest_rejects_legacy_placeholder_config_at_cli` added
 
 #### S44-2: DKIM DNS sanity check at daemon startup + louder setup warning
 
@@ -199,13 +200,13 @@ Fix: `sanitize_attachment_filename(raw: &str, index: usize) -> String` — NFC-n
 
 **Priority:** P0
 
-- [ ] Helper `public_key_spki_base64(path: &Path) -> Result<String>` in `src/dkim.rs` (extract from existing setup code if already derived there; otherwise new); unit-tested against a fixture key
-- [ ] `src/serve.rs` startup: after DKIM key load and before binding listeners, resolve TXT `dkim._domainkey.{config.domain}` via the existing resolver; if DNS resolution fails, log at `warn` and continue (transient, non-fatal); if DNS `p=` differs from on-disk SPKI, log a multi-line warning to stderr + journal stating mismatch detected, receiver DKIM will fail, and suggesting `aimx setup` to republish DNS
-- [ ] Startup never blocks or exits on mismatch — daemon proceeds to bind SMTP + UDS listeners normally
-- [ ] `src/setup.rs verify_dkim` mismatch branch: render the FAIL line via `term::error_red` (or existing semantic helper) and append a second line: "⚠ Outbound DKIM signatures will FAIL verification at receivers until DNS matches."
-- [ ] Integration test: spin `aimx serve` with a mocked resolver returning a mismatched `p=`; assert the startup log contains the mismatch warning; assert the daemon still binds both listeners and accepts mail
-- [ ] Integration test: spin `aimx serve` with a resolver that fails the DKIM TXT lookup; assert startup logs a `warn` and continues
-- [ ] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean
+- [x] Helper `public_key_spki_base64(path: &Path) -> Result<String>` in `src/dkim.rs` (extract from existing setup code if already derived there; otherwise new); unit-tested against a fixture key
+- [x] `src/serve.rs` startup: after DKIM key load and before binding listeners, resolve TXT `dkim._domainkey.{config.domain}` via the existing resolver; if DNS resolution fails, log at `warn` and continue (transient, non-fatal); if DNS `p=` differs from on-disk SPKI, log a multi-line warning to stderr + journal stating mismatch detected, receiver DKIM will fail, and suggesting `aimx setup` to republish DNS
+- [x] Startup never blocks or exits on mismatch — daemon proceeds to bind SMTP + UDS listeners normally
+- [x] `src/setup.rs verify_dkim` mismatch branch: render the FAIL line via `term::error_red` (or existing semantic helper) and append a second line: "⚠ Outbound DKIM signatures will FAIL verification at receivers until DNS matches."
+- [x] Integration test: spin `aimx serve` with a mocked resolver returning a mismatched `p=`; assert the startup log contains the mismatch warning; assert the daemon still binds both listeners and accepts mail <!-- Partial: substituted with unit-test coverage via `DkimTxtResolver` trait + mock resolver exercising `run_dkim_startup_check` / `log_dkim_startup_check` across Match/Mismatch/NoRecord/NoPTag/ResolveError branches + pure `evaluate_dkim_startup` branch tests. Reviewer accepted deferral. Full end-to-end on live `run_serve` deferred as non-blocker. -->
+- [x] Integration test: spin `aimx serve` with a resolver that fails the DKIM TXT lookup; assert startup logs a `warn` and continues <!-- Partial: ResolveError branch exercised via fake resolver returning error; end-to-end daemon-level integration deferred. -->
+- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean
 
 #### S44-3: `aimx agent-setup claude-code` hint fix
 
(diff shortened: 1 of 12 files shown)


uzyn (commented):
## Sprint 47 Review — Post-v1 Non-blocking Cleanup

### Sprint Goal Assessment
The sprint goal is fully met. All four thematic stories close the eight accumulated non-blocking items from Sprints 44–46, and the implementation is a clean defense-in-depth pass with no user-visible churn. The most substantive change — unifying inbound ingest and MARK writers under a single per-mailbox `tokio::sync::Mutex` (S47-4) — is architecturally sound, well-documented, and backed by two new concurrency integration tests. I ran `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check`, unit tests (734 passed), and the new integration tests (`dkim_startup_*`, `concurrent_*`) locally — all clean.

**[S47-1] DKIM startup check — end-to-end integration test + runtime-flavor contract**
- [x] Integration test covers `Mismatch` + `ResolveError` branches, asserts both listeners bind — met. `dkim_startup_check_mismatch_logs_warning_and_binds_listeners` and `dkim_startup_check_resolve_error_logs_warning_and_binds_listeners` exercise the full `run_serve` path with a clean `AIMX_TEST_DKIM_RESOLVER_OVERRIDE` hatch.
- [x] `HickoryDkimResolver::resolve_dkim_txt` debug-asserts `RuntimeFlavor::MultiThread` with a comment — met. The choice to debug-assert over async-ify is the right call (zero-cost in production, loud in test builds).
- [x] Clippy / fmt / tests clean — met.

**[S47-2] Exhaustiveness & defense-in-depth hardening**
- [x] `restart_hint_command` uses explicit `InitSystem::{Systemd, OpenRC, Unknown}` arms — met.
- [x] `restart_hint_command_is_exhaustive_over_init_system` test destructures every variant — met. Adding a new variant will break compilation of this test arm, which is exactly the intent.
- [x] `execute_triggers` calls `.env_clear()` then selectively re-adds `PATH`/`HOME`/`AIMX_*` — met.
- [x] `execute_triggers_clears_parent_env` asserts `AIMX_LEAK_TEST` does not leak and `PATH` is preserved — met.
- [x] Clippy / fmt / tests clean — met.

**[S47-3] Validation tightening + TOML-rewrite preservation + stronger rename-failure test**
- [x] `validate_mailbox_name` consolidated in `src/mailbox.rs`, tightened to `[a-z0-9._-]+` — met. Duplicate in `mailbox_handler.rs` correctly threads through.
- [x] New unit tests for rejected (`hello world`, `..foo`, `""`, `.leading`, `trailing.`, `Alice`, `alice+bob`, etc.) and accepted names — met, with stronger coverage than the AC required.
- [~] `write_config_atomic` — "document-current-behaviour" path chosen. See detail below: the pinning test asserts the stanza is **dropped**, while the AC literally reads "…a test asserting a known-unknown stanza survives". Non-blocker — the AC wording conflates the `toml_edit` and document-it paths; the intent of pinning current behaviour is met.
- [x] `create_failure_at_disk_write_leaves_handle_and_disk_unchanged` rewritten to force a `rename(2)` failure via a non-empty target directory — met. Also adds a stray-temp-file check which is a nice bonus.
- [x] Clippy / fmt / tests clean — met.

**[S47-4] Unify ingest + MARK writers under one per-mailbox lock**
- [x] New `src/mailbox_locks.rs` with `MailboxLocks` type — met. Clean API, good doc header, lock hierarchy spelled out.
- [x] `ingest_email` takes `&MailboxLocks` and acquires via `blocking_lock()` — met. The `blocking_lock()` placement is correct for all three callers (stdin path creates a fresh map; SMTP session inherits from the daemon; tests use the helper).
- [x] `StateContext` holds `Arc<MailboxLocks>`; `handle_mark` uses `.lock().await` — met.
- [x] `aimx serve` threads one `MailboxLocks` into SMTP server, StateContext, MailboxContext — met via the new `with_mailbox_locks` builder.
- [x] Module-level comment in `state_handler.rs` updated; full rationale in `mailbox_locks.rs` with explicit outer → inner hierarchy — met.
- [x] `concurrent_ingest_burst_and_mark_same_mailbox_no_torn_writes` — met. Fires 6 SMTP bursts + 2 MARK-READs, asserts `+++`-frontmatter integrity and TOML-parseability on every `.md` file.
- [x] `concurrent_mailbox_create_and_ingest_does_not_deadlock` — met. 20s watchdog is generous (tests complete in <1s) but acceptable.
- [x] Clippy / fmt / tests clean — met.

### Test Coverage
Strong overall. Specific observations:

- `blocking_lock()` soundness is indirectly validated by the two new concurrency integration tests — they exercise the actual SMTP → `spawn_blocking` → `ingest_email` → `blocking_lock()` path against concurrent async MARK/MAILBOX-CREATE handlers. No bug surfaced on my local run. Good.
- The `MailboxLocks::async_lock_serializes_two_holders` unit test is a smart use of `tokio::time::timeout` to assert waiter behavior without relying on thread-sleep timing.
- Integration tests correctly use `setup_test_env` which seeds `agent.example.com` with `alice` and `catchall`, matching the From/To addresses used in the bursts.

### Potential Bugs

**Non-blocker: `concurrent_mailbox_create_and_ingest_does_not_deadlock` watchdog thread leaks after test completion.**

`tests/integration.rs` — the watchdog thread is spawned then `drop(watchdog)`-ed after the joins return. `drop` on a `JoinHandle` only detaches the thread; the inner closure continues sleeping 20s and then panics. With the default `panic = "unwind"` the panic is contained to that thread, but:

1. The test "completes successfully" while a detached thread is still sleeping — up to 20s of unreaped work per run.
2. If the cargo test binary exits before the thread's 20s timer, the thread is reaped cleanly; if it's still alive (another test still running), the panic prints a scary "Sprint 47 deadlock watchdog fired" banner to stderr that is a false alarm.

A cleaner pattern is an `Arc<AtomicBool>` the watchdog polls every ~100ms and exits early when the test sets it before `drop`. Not shipping-blocking, but worth fixing when touched again.

**Non-blocker: S47-3 pinning-test wording diverges from the literal AC.**

The S47-3 AC reads: *"OR document the current behavior with a comment + add a test asserting a known-unknown stanza survives"*. The implementer correctly chose the document-current-behaviour path and added `unknown_stanza_is_dropped_on_rewrite` — but this test asserts the stanza is **dropped**, not that it survives. The AC wording seems to conflate the two paths (survives only makes sense on the `toml_edit` path). The *spirit* of "document current behaviour with a pinning test" is met: the doc comment on `write_config_atomic` clearly explains the v1 limitation, and the test pins the current (dropping) behavior so any accidental future change (e.g. adopting `toml_edit` without updating the doc) will trip this test. I consider this fully discharged — flagging only because the PR says to confirm the documented-behaviour path is "genuinely complete, not partial". It is.

**Non-blocker (low): `execute_triggers_clears_parent_env` mutates process-global env.**

`std::env::set_var("AIMX_LEAK_TEST", ...)` + `remove_var` is process-global (the test correctly wraps these in `unsafe`). The sentinel name is unique to this test so there's no correctness race with other tests reading `AIMX_LEAK_TEST`. However, if the assertion between `set_var` and `remove_var` panics, the sentinel leaks. The code correctly removes the var *before* the final assertions so this is minimal-risk, but a `scopeguard`-style RAII pattern would be more robust.

### Code Quality

Generally good. Two small notes:

- `mailbox_locks::MailboxLocks::new()` + `Default` impl + the public `with_mailbox_locks` builder on `SmtpServer` with `Arc::new(MailboxLocks::new())` as default gives three initialization paths. Works, but slightly redundant. Not worth changing.
- The `AIMX_TEST_DKIM_RESOLVER_OVERRIDE` test-only env-var hatch is well-gated: it only activates when the env var is set, and `aimx serve` in production never sets it. The namespaced `AIMX_TEST_` prefix is appropriate. The format parser (`ok:...||...` / `err:...` / `no-record`) falls through to the real resolver on unrecognized input, which is safer than panicking — good call.

### Alignment with PRD
No divergence. All four stories are hardening / test-coverage / architectural-cleanup; no PRD changes touched. Sprint goal explicitly disclaims user-visible features.

- **Overall verdict:** Needs minor fixes — but only minor (all flagged items are non-blockers). Can be merged as-is if the author accepts the trade-offs; each finding is cheap to address in a follow-up if preferred.
- **Blockers:** none.
- **Non-blockers (should fix but not blocking):**
  1. Replace `drop(watchdog)` with an `AtomicBool` cancel flag in `concurrent_mailbox_create_and_ingest_does_not_deadlock` so the watchdog thread exits promptly on success rather than lingering for 20s and panicking in the background.
  2. Reconcile S47-3 AC wording with the chosen implementation — either update the sprint doc to reflect the "dropped" pinning direction or, if reviewers prefer, add an explanatory comment next to `unknown_stanza_is_dropped_on_rewrite`.
  3. Consider a `scopeguard`-style RAII for `AIMX_LEAK_TEST` cleanup in `execute_triggers_clears_parent_env`.
- **Nice-to-haves:** none material.

Recommended path: merge as-is (all findings are legitimately non-blocking and the sprint goal is met), then track the three items in the Non-blocking Review Backlog.

uzyn:
## Cycle 2: addressed all three non-blockers from the Sprint 47 review

All three flagged items are now fixed. `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and the full test suite (734 unit + 63 integration) are clean.

### 1. Watchdog thread no longer lingers (`tests/integration.rs`)

Replaced `drop(watchdog)` with an `Arc<AtomicBool>` cancel flag. The watchdog now polls the flag every 100ms and exits promptly on success rather than detaching for the full 20s. On the deadlock path it still fires. Success path also `.join().unwrap()`s the watchdog so any future misuse surfaces immediately.

### 2. RAII env guard for `AIMX_LEAK_TEST` (`src/channel.rs`)

Replaced manual `set_var`/`remove_var` with an `EnvVarGuard` struct whose `Drop` impl calls `remove_var`. The sentinel is now cleaned up even if an assertion panics between set and remove. Test body is simpler too — no separate cleanup block required.

### 3. S47-3 AC wording reconciled (`docs/sprint.md`)

Added an inline `<!-- ... -->` note on the `write_config_atomic` AC explaining that the document-current-behaviour path was chosen, that the "stanza survives" wording only applies to the alternate `toml_edit` path, and that `unknown_stanza_is_dropped_on_rewrite` pins the documented v1 contract (dropped, not preserved). The test itself already carried a clarifying comment; no code change was needed.

### Verification

- `cargo test` — 734 unit + 63 integration pass, 0 failures

Pushed as 5c86c3a.

uzyn (commented):
## Sprint 47 Re-Review (Cycle 2) — Post-v1 Non-blocking Cleanup

All three cycle-1 non-blockers are resolved cleanly in 5c86c3a. No regressions, no new findings. Verified locally: `cargo fmt -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, the two tests directly touched by this commit (`concurrent_mailbox_create_and_ingest_does_not_deadlock` and `channel::tests::execute_triggers_clears_parent_env`) pass.

### Resolved issues

1. **Watchdog thread no longer lingers** — `tests/integration.rs:3301-3332`. `drop(watchdog)` replaced with `Arc<AtomicBool>` cancel flag; watchdog polls every 100ms and returns promptly on cancel; the main body now `.join().unwrap()`s the watchdog instead of detaching. Correctly re-checks the flag after the deadline loop exits to avoid a spurious panic on the race where cancel is set right at the deadline. Clean.

2. **RAII env guard for `AIMX_LEAK_TEST`** — `src/channel.rs:717-738`. `EnvVarGuard::set()` / `impl Drop` pattern, bound to `_sentinel` (leading underscore, not bare `_`, so Rust actually holds the binding to end of scope — the important detail is correct). Sentinel is now removed even if an assertion panics between `set` and the final assertions. Simpler test body as a bonus.

3. **S47-3 AC wording reconciled** — `docs/sprint.md:436`. Inline `<!-- ... -->` annotation preserves the original AC text (good for audit trail) and explains that the document-current-behaviour path was chosen, that the "stanza survives" phrasing only applies to the alternate `toml_edit` path, and names `unknown_stanza_is_dropped_on_rewrite` as the pinning test. Fine.

### Still unresolved

None.

### New issues introduced

None. The diff is tightly scoped (3 files, +57/-22 lines) and every hunk is in service of one of the three flagged items.

### Verdict

**Ready to merge.** All cycle-1 non-blockers addressed, nothing new surfaced, sprint goal remains fully met. Sprint 47 closes the eight hardening items from Sprints 44–46 with no user-visible changes and a measurable improvement in concurrency correctness (unified per-mailbox lock) and test hygiene.

```
[Sprint 47] Post-v1 Non-blocking Cleanup

Closes the eight hardening items carried over from Sprints 44-46 reviews,
grouped into four thematic stories:

- S47-1: DKIM startup check — integration test covering Mismatch and
  ResolveError branches via AIMX_TEST_DKIM_RESOLVER_OVERRIDE; resolver
  debug_asserts MultiThread runtime flavor.
- S47-2: InitSystem match exhaustiveness (compile-time); channel triggers
  now env_clear() + selectively re-add PATH/HOME/AIMX_*.
- S47-3: validate_mailbox_name tightened to [a-z0-9._-]+ and consolidated;
  write_config_atomic v1 drop-unknown-stanzas behaviour pinned by test;
  rename-failure test rewritten to genuinely exercise the rollback branch.
- S47-4: inbound ingest + MARK + MAILBOX-* handlers unified under one
  per-mailbox tokio::sync::Mutex via MailboxLocks; outer per-mailbox
  lock -> inner process-wide CONFIG_WRITE_LOCK hierarchy documented;
  two new concurrency integration tests.

Defense-in-depth, test-coverage, and architectural-cleanup only — no
user-visible changes.

Tests: 734 unit + 63 integration passing; clippy and fmt clean.
```

PR #88, opened 2026-04-18
feat(cli): port-25 preflight first in setup; DNS check in status

uzyn:
## Summary

Two independent ergonomic improvements:

1. **`aimx setup` runs the port-25 preflight first.** Before prompting for a domain or doing any install work, setup now checks for a foreign process on :25 and probes outbound/inbound port 25 against the verify service. A VPS that blocks SMTP fails fast without making the operator fill in details or leaving behind TLS certs, config, or DKIM keys.
2. **`aimx status` gains a `DNS` section.** Runs the same `verify_all_dns` checks `aimx setup` runs (MX, A, AAAA, SPF, DKIM, DMARC) so operators can spot DNS drift without re-running setup. Uses the on-disk DKIM public key when present, honours `config.enable_ipv6`, and stays non-root with no change to exit-code semantics.

### Setup preflight-first
- [x] Run port-25 occupancy + outbound/inbound preflight after the root check and before the domain prompt
- [x] On preflight failure: exit with the existing error message, no domain prompt, no TLS cert, no config or DKIM write
- [x] `Port25Status::OtherProcess(...)` → same foreign-process error as before, just raised earlier
- [x] `Port25Status::Aimx` (re-entry) → preflight runs against the live daemon, no temp listener
- [x] Duplicate preflight block at the old step-5 location removed

### Status DNS
- [x] `aimx status` renders a `DNS` section using `setup::verify_all_dns` + `setup::dns_verification_lines`
- [x] AAAA / SPF (IPv6) included iff `config.enable_ipv6 = true`
- [x] DKIM verification uses the local public key via `dkim::dns_record_value` when present
- [x] If server IP cannot be determined, section renders a single `skipped` warn line and the rest of `aimx status` still renders
- [x] Section appears after Mailboxes + Recent activity
- [x] No root required; DNS failures do NOT change exit code

## Out of scope

- Renaming `aimx verify` to `aimx portcheck` (separate in-flight branch)
- `--dns` / `--no-dns` flag on `aimx status` (always-on was the explicit choice)
- Reverse-DNS / PTR checks
- Any changes to `aimx verify`, `DnsVerifyResult`, or individual `verify_*` helpers

## Technical approach

- `src/setup.rs::run_setup`: reordered so root check → port-25 occupancy + preflight (with temp listener when `Free`, direct when `Aimx`) → domain prompt → finalize/install. The old step-5 preflight block is removed.
- `src/status.rs`: new `DnsSection { results, records }` struct, new `dns: Option<DnsSection>` field on `StatusInfo`, new `gather_dns_section` helper, `format_status` renders a `DNS` header + per-record lines (or a `skipped` line when `None`). `gather_status_with_ops` gained a `&dyn NetworkOps` parameter.
- `src/cli.rs`, `book/getting-started.md`, `book/setup.md`, `book/troubleshooting.md`: descriptions updated to mention the new DNS section and the new setup ordering.

## Test plan

- New setup test `fresh_setup_does_not_write_config_when_preflight_fails` — asserts that with `outbound_port25 = false`, `run_setup` returns an error and neither `config.toml`, the DKIM key, nor the service file appear on disk. Replaces the inverted `fresh_setup_writes_config_and_dkim_before_preflight`.
- New status tests:
  - `gather_status_includes_dns_when_server_ip_available`
  - `gather_status_dns_none_when_server_ip_unavailable`
  - `gather_status_dns_none_when_get_server_ips_errors`
  - `gather_dns_includes_aaaa_when_ipv6_enabled`
  - `format_status_renders_dns_section`
  - `format_status_renders_dns_skipped_when_none`
- Full suite green: `cargo test` → 746 unit + 63 integration pass; `cargo fmt -- --check` clean; `cargo clippy -- -D warnings` clean on both the `aimx` root crate and `services/verifier`.


uzyn (commented):
## Changes Overview

Reorders `aimx setup` so the port-25 preflight (outbound + inbound + foreign-process scan) runs immediately after the root check — before the domain prompt, TLS cert, or config writes — so a port-25-blocked VPS fails fast without prompting or leaving artefacts. Also adds a `DNS` section to `aimx status` that runs the same DNS checks setup runs (MX/A/AAAA/SPF/DKIM/DMARC) via the existing `verify_all_dns` helper, displayed after Mailboxes + Recent activity.

## Scope Alignment

Fully met. Every PR-body checkbox corresponds to a concrete code change, and no scope creep — the six changed files are all in the described surface area.

## Potential Bugs

**1. Misleading DKIM "→ Add:" hint when local public key is missing or unreadable** — Non-blocker
- File: `src/status.rs`, `gather_dns_section`
- When `<dkim_dir>/public.key` doesn't exist or can't be parsed, `local_dkim_pubkey` is `None` and `dkim_dns_value` falls back to the literal `\"v=DKIM1; k=rsa; p=\"` (empty `p=`). That string is embedded in `records` and surfaced by `dns_verification_lines` as the "→ Add:" hint on DKIM failures. An operator whose DKIM key never got generated would see a hint recommending they add a DNS record with an empty key value.
- Why non-blocker: `aimx status` loads `config.toml` (requires a completed setup), so `public.key` being absent is a narrow post-setup-corruption state. Misleading, not dangerous.

**2. Fragile `.iter().skip(3)` on `dns_verification_lines`** — Non-blocker
- File: `src/status.rs`, `format_status` (DNS branch)
- Relies on `dns_verification_lines` emitting a specific three-line preamble (`\"\"`, `\"DNS Verification:\"`, `\"\"`) and skips it. If the helper's preamble ever changes shape (trailing blank dropped, extra leading line), `format_status` will silently swallow the first real record or leak a stray "DNS Verification:" line. Existing format-DNS tests only assert `\"DNS\"` and `\"MX:\"` are substrings, which would remain true across most plausible drift.
- Why non-blocker: maintainability risk, not a runtime bug.

**3. `aimx status` now performs synchronous DNS lookups with no explicit budget** — Non-blocker
- File: `src/status.rs`, `gather_dns_section`
- `verify_all_dns` issues ~6 DNS queries (MX, A, SPF TXT, DKIM TXT, DMARC TXT, plus AAAA/SPF(IPv6) when enabled), synchronously. On a machine with broken DNS, `aimx status` could block for several seconds before rendering. Hickory-resolver has its own per-query defaults but there's no overall status-command budget.
- Why non-blocker: status is an operator command run interactively; multi-second latency is tolerable. No correctness impact.

**4. Silent UX regression on re-entry with `Port25Status::Aimx`** — Non-blocker
- File: `src/setup.rs`, `run_setup`
- The pre-PR code printed `` `aimx serve` is already running on port 25. Proceeding with setup.`` when setup detected a live daemon on fresh-install (`!already_configured`). The new code drops this message. The preflight itself runs and its PASS/FAIL badges communicate the same info, but the PR body doesn't flag this change.
- Why non-blocker: narrow edge case, preflight output covers it.

## Test Coverage

Tests cover the new plumbing (StatusInfo field, DnsSection construction, IPv6 gating, missing-IP / erroring-IP fallback, section rendering, skipped rendering) and the setup fail-fast invariant. Gaps worth closing in follow-up:

- `gather_status_includes_dns_when_server_ip_available` only asserts result *names* are present. It doesn't assert that matching-IP records actually produce `DnsVerifyResult::Pass`. A single `assert!(matches!(results.iter().find(...), Some((_, DnsVerifyResult::Pass))))` would prove the IP→result wiring, not just struct shape.
- `fresh_setup_does_not_write_config_when_preflight_fails` pre-supplies `Some(\"example.com\")` as the domain, bypassing the prompt entirely. It proves "no side effects on failure" but not directly "preflight runs before prompt." Direct ordering testing would require stdin mocking; source-order inspection is sufficient.
- No test exercises the `local_dkim_pubkey = None` rendering path (bug #1).

## Security Issues

None. The changes are read-only surface additions (DNS lookups, service-state probe) plus a reorder of existing logic. No new inputs, writes, or privileges.

## Code Quality

- `gather_dns_section` is readable and well-scoped.
- `MockNetworkOps` in `status.rs::tests` duplicates the shape of `MockNetworkOps` in `setup.rs::tests` and `verify.rs::tests`. A shared `#[cfg(test)] pub mod test_support` would reduce drift. Non-blocker — each copy is small and self-contained.

**Overall verdict:** Ready to merge (with non-blockers worth a follow-up)

**Blockers:** None

**Non-blockers:**
1. Misleading DKIM "→ Add:" hint when `public.key` is missing (suppress the hint or guard in `gather_dns_section`).
2. Fragile `.iter().skip(3)` — extract a per-record formatter from `dns_verification_lines`.
3. No explicit DNS-query budget in `aimx status`.
4. Dropped "aimx serve already running" message on re-entry — fold into the new preflight header, or call out in the PR body.

**Nice-to-haves:**
1. Strengthen `gather_status_includes_dns_when_server_ip_available` to assert at least one `DnsVerifyResult::Pass`.
2. Add a test for the `local_dkim_pubkey = None` path in `gather_dns_section`.
3. Consolidate the three near-identical `MockNetworkOps` test doubles.

```
feat(cli): port-25 preflight first in setup; DNS check in status (#88)

- `aimx setup` now runs the port-25 occupancy scan + outbound/inbound
  preflight immediately after the root check, before the domain prompt
  or any filesystem writes. A VPS that blocks SMTP fails fast without
  prompting the operator or leaving TLS/config/DKIM artefacts behind.
- `aimx status` gains a `DNS` section that runs the same MX/A/AAAA/
  SPF/DKIM/DMARC checks as setup via `verify_all_dns`, honouring
  `config.enable_ipv6` and using the on-disk DKIM public key when
  present. Stays non-root; DNS failures don't affect exit code.
- Updates CLI help and book/{getting-started,setup,troubleshooting}.md
  to reflect the new ordering and status surface.
```

uzyn:
Addressed the non-blocker feedback from the review. Pushed as `4e285e4`.

### Fixed

- **#1 Misleading DKIM "→ Add:" hint** — `gather_dns_section` now drops the DKIM TXT record from `DnsSection.records` when no local `public.key` is on disk. The DKIM DNS check still runs (and still reports PASS/FAIL/MISSING); only the bogus Add suggestion with an empty `p=` is suppressed. Guarded by new test `gather_dns_drops_dkim_record_when_local_pubkey_missing`.
- **#2 Brittle `.iter().skip(3)`** — split `dns_verification_lines` into two functions: `dns_verification_record_lines` (body only, used by `format_status`) and `dns_verification_lines` (wraps body with preamble + trailing blank, used by setup). Existing setup callers + tests unchanged.
- **#3 DNS-query budget** — `RealNetworkOps::resolve_*` now shells out via a shared `dig_short_args` helper that adds `+time=2 +tries=1`. dig's own defaults would let a single bad recursive resolver stall the command for ~15s per query (~90s across the six queries `verify_all_dns` issues); the new bounds cap worst case at ~12s. New test `dig_short_args_bounds_per_query_timeout` locks the per-query budget in so a future refactor dropping the args trips a red test.
- **#4 Dropped "aimx serve already running" message** — restored as a one-liner inside the new preflight block: `` `aimx serve` is already running on port 25 — probing the live daemon. `` Only fires when `Port25Status::Aimx`, which is the same narrow case as before.
- **#5 Test strengthening** — `gather_status_includes_dns_when_server_ip_available` now asserts the primed A record produces `DnsVerifyResult::Pass`, not just that the result name appears in the vec.
- **#6 Missing-DKIM-pubkey test** — see #1 above.

### Intentionally deferred (from nice-to-haves)

- **#7 MockNetworkOps consolidation** — the three mocks in `setup.rs::tests`, `verify.rs::tests`, and `status.rs::tests` aren't identical: they track different call counts, and the status one has `unreachable!()` guards on port-25 methods that encode a real "`gather_status` must not touch these" invariant. Consolidation involves real design tradeoffs (union mock vs. trait dance vs. builder) and is scope-worthy of its own PR rather than a drop-in here.

### Test results

- `cargo test`: 748 unit + 63 integration pass
- `cargo clippy -- -D warnings`: clean on root crate and `services/verifier`

PR #94, opened 2026-04-19
fix: DNS verification no longer flaps between PASS and MISSING

uzyn:
## Summary

`aimx setup` and `aimx status` occasionally reported DNS records as `MISSING` on one verify pass and `PASS` on the next (and vice versa) with no actual change to the authoritative records — e.g., `A: PASS` then `A: MISSING` on the very next retry, or `DKIM: MISSING` then `DKIM: PASS`. This PR fixes the root cause.

**Root cause.** The `dig`-based resolvers in `RealNetworkOps` (`resolve_mx` / `resolve_a` / `resolve_aaaa` / `resolve_txt`) ran a single UDP query with `+tries=1 +time=2` and ignored dig's exit status. Any transient UDP loss or slow recursive-resolver response produced an empty stdout that the code treated identically to "no record exists," so a different subset of the 6 checks failed on each retry.

**Fix.** A module-private `DigRunner` trait + `dig_with_cascade` helper:
- Queries explicit public resolvers in order: **1.1.1.1 → 8.8.8.8 → 9.9.9.9**.
- Up to **3 attempts per resolver** with a 300 ms backoff on non-zero exit.
- An exit-0 response (including a NOERROR/NXDOMAIN empty answer) is authoritative and stops the cascade immediately — genuine "no record" paths stay fast.
- Only when every resolver exhausts its attempts does the helper return `io::Error`, which `resolve_*` propagates as `Err` so `verify_*` renders the failure as `Fail` (accurate) rather than `Missing` (misleading).

## Requirements

- [x] Check dig's exit status — non-zero no longer masquerades as "no record."
- [x] Query explicit public resolvers (1.1.1.1 → 8.8.8.8 → 9.9.9.9); exit-0 including empty stops the cascade.
- [x] In-process retry (3 attempts × 300 ms) on non-zero dig exit; exhaustion yields `Err` → `DnsVerifyResult::Fail`.
- [x] Latency budget bounded (old `DIG_TIMEOUT_SECS * DIG_TRIES <= 5` compile-time invariant dropped — no longer captures the right quantity given the cascade).
- [x] `NetworkOps` trait, `MockNetworkOps`, and `verify_*` semantics unchanged (only `Err` now produces `Fail` instead of silently becoming `Missing`).
- [x] Unit tests cover the cascade, retries, the empty-but-ok no-fallback case, `resolve_a` integration, and the `verify_a` → `Fail`-not-`Missing` mapping.
- [x] `cargo build`, `cargo clippy -- -D warnings`, `cargo fmt -- --check`, `cargo test` all green.

## Out of scope

- Authoritative-NS lookups (querying the domain's own NS directly). Deferred — public-resolver retry addresses the immediate flakiness.
- DNS-over-HTTPS / DNS-over-TLS. Plain UDP via dig is fine.
- Replacing dig with a native Rust resolver (`hickory-resolver` is already a dep for MX, but porting these checks is a larger refactor).
- Parallelizing the 6 queries in `verify_all_dns`.
- Operator-visible resolver selection (no new CLI flag, no config). Resolvers are hardcoded.

## Technical approach

All changes in `src/setup.rs`:

1. **`DigRunner` trait + `RealDigRunner`** — injectable shim over `std::process::Command::new("dig").args(..).output()`. Mirrors the existing `NetworkOps` / `SystemOps` / `MailTransport` trait-based testing pattern. `pub(crate)`, no new public surface.
2. **`dig_with_cascade(runner, record_type, domain)`** — iterates `DIG_RESOLVERS`, up to `DIG_RETRY_ATTEMPTS` per resolver, with `dig_retry_delay()` between retries. Exit-0 wins and returns (empty or not). Exhaustion returns `io::Error` with the last resolver's stderr tail.
3. **`dig_short_args(resolver, record_type, domain)`** — signature updated to include `@<resolver>`. Old single-invocation bounds assertion removed; the effective budget now lives in the cascade's `resolvers × attempts × per-attempt-timeout` math.
4. **`RealNetworkOps`** — now carries `dig_runner: Box<dyn DigRunner + Send + Sync>`. `Default` and `from_verify_host` inject `RealDigRunner`; a `#[cfg(test)]`-gated `with_dig_runner` builder swaps in a mock. Manual `Debug` impl skips the trait-object field.
5. **`resolve_mx` / `resolve_a` / `resolve_aaaa` / `resolve_txt`** — each is now a three-line wrapper around `dig_with_cascade` that applies type-specific post-processing (IP parsing, TXT quote-stripping). Error propagation via `?` — no more `Ok(vec![])` on failure.
6. **`dig_retry_delay()`** — production returns 300 ms; `#[cfg(test)]` returns 0 so the all-resolvers-fail test doesn't cost 2.4 s.

## Test plan

10 new unit tests in `src/setup.rs`:

- `dig_short_args_includes_resolver_and_bounds` — argv shape, updated for the new resolver parameter.
- `dig_resolvers_default_cascade_order` — locks in `[1.1.1.1, 8.8.8.8, 9.9.9.9]` order.
- `dig_cascade_first_resolver_succeeds` — happy path, single mock call, targets 1.1.1.1.
- `dig_cascade_falls_through_on_exit_code` — all 3 attempts on 1.1.1.1 fail → 8.8.8.8 succeeds first try; verifies exact call count and resolver-per-call.
- `dig_cascade_all_resolvers_fail_returns_err` — 9 failed attempts → `io::Error`, total call count = 9.
- `dig_cascade_empty_but_ok_is_success_no_fallback` — NOERROR empty answer is authoritative, no fallback.
- `dig_cascade_retries_on_spawn_error` — spawn-errors (not just non-zero exits) also trigger retry.
- `resolve_a_returns_err_when_all_resolvers_fail` — integration through `RealNetworkOps`; old code would have returned `Ok(vec![])`.
- `resolve_a_succeeds_on_second_resolver_after_first_fails` — end-to-end: 1.1.1.1 fails, 8.8.8.8 wins → parsed IP returned.
- `verify_a_maps_resolver_err_to_fail_not_missing` — the core user-visible change: real lookup failure renders as `Fail` with the underlying error, not `Missing`.

All 228 `setup::` tests pass; full `cargo test` run: 790 unit + 64 integration, all green. `cargo clippy -- -D warnings` and `cargo fmt -- --check` clean.

## Notes for reviewers

- `DIG_RESOLVERS`, `DIG_RETRY_ATTEMPTS`, `DIG_RETRY_DELAY_MS` are `pub(crate)` constants — happy to change to fully private if you'd prefer.
- `RealNetworkOps` no longer derives `Debug` (can't — trait-object field); a hand-written `Debug` impl preserves `.unwrap_err()` usage in the existing `from_verify_host` tests.
- The 300 ms retry delay is a judgment call. Shorter feels snappier on a flaky network but risks hitting the same packet-loss window; longer feels sluggish. 300 ms was the user's confirmed choice.

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -342,10 +342,89 @@ pub(crate) fn parse_hostname_i_output(stdout: &str) -> (Option<Ipv4Addr>, Option
     (ipv4, ipv6)
 }
 
-#[derive(Debug)]
+/// Injectable shim over `std::process::Command::new("dig").args(...).output()`.
+/// Tests inject a scripted implementation to exercise the resolver cascade
+/// without shelling out to a real `dig`.
+pub(crate) trait DigRunner {
+    fn run(&self, args: &[String]) -> io::Result<std::process::Output>;
+}
+
+pub(crate) struct RealDigRunner;
+
+impl DigRunner for RealDigRunner {
+    fn run(&self, args: &[String]) -> io::Result<std::process::Output> {
+        std::process::Command::new("dig").args(args).output()
+    }
+}
+
+/// Query `dig` for a record, cascading across `DIG_RESOLVERS` with up to
+/// `DIG_RETRY_ATTEMPTS` per resolver.
+///
+/// Stop conditions, in priority order:
+/// 1. Any invocation returns exit 0 → parse and return the lines (an
+///    empty vec is a valid success; NOERROR/NXDOMAIN is authoritative
+///    so we do NOT fall through on an empty-but-ok response).
+/// 2. All resolvers × all attempts exit non-zero → return `io::Error`
+///    with the last resolver's stderr tail for diagnostics.
+///
+/// The empty-filtered, trimmed lines are returned raw. Type-specific
+/// post-processing (IP parsing for A/AAAA, quote-stripping for TXT)
+/// lives in the `resolve_*` callers.
+pub(crate) fn dig_with_cascade(
+    runner: &dyn DigRunner,
+    record_type: &str,
+    domain: &str,
+) -> io::Result<Vec<String>> {
+    let mut last_err_detail = String::new();
+    for resolver in DIG_RESOLVERS {
+        for attempt in 0..DIG_RETRY_ATTEMPTS {
+            let args = dig_short_args(resolver, record_type, domain);
+            match runner.run(&args) {
+                Ok(output) if output.status.success() => {
+                    let stdout = String::from_utf8_lossy(&output.stdout);
+                    let lines: Vec<String> = stdout
+                        .lines()
+                        .filter(|l| !l.trim().is_empty())
+                        .map(|l| l.trim().to_string())
+                        .collect();
+                    return Ok(lines);
+                }
+                Ok(output) => {
+                    last_err_detail = format!(
+                        "@{resolver} exit={} stderr={}",
+                        output.status,
+                        String::from_utf8_lossy(&output.stderr).trim()
+                    );
+                }
+                Err(e) => {
+                    last_err_detail = format!("@{resolver} spawn error: {e}");
+                }
+            }
+            if attempt + 1 < DIG_RETRY_ATTEMPTS {
+                std::thread::sleep(dig_retry_delay());
+            }
+        }
+    }
+    Err(io::Error::other(format!(
+        "dig {record_type} {domain} failed across all {} resolvers ({} attempts each); last error: {last_err_detail}",
+        DIG_RESOLVERS.len(),
+        DIG_RETRY_ATTEMPTS,
+    )))
+}
+
 pub struct RealNetworkOps {
     pub verify_host: String,
     pub check_service_smtp_addr: String,
+    dig_runner: Box<dyn DigRunner + Send + Sync>,
+}
+
+impl std::fmt::Debug for RealNetworkOps {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RealNetworkOps")
+            .field("verify_host", &self.verify_host)
+            .field("check_service_smtp_addr", &self.check_service_smtp_addr)
+            .finish_non_exhaustive()
+    }
 }
 
 impl Default for RealNetworkOps {
@@ -353,6 +432,7 @@ impl Default for RealNetworkOps {
         Self {
             verify_host: DEFAULT_VERIFY_HOST.to_string(),
             check_service_smtp_addr: DEFAULT_CHECK_SERVICE_SMTP_ADDR.to_string(),
+            dig_runner: Box::new(RealDigRunner),
         }
     }
 }
@@ -365,7 +445,14 @@ impl RealNetworkOps {
         Ok(Self {
             verify_host: trimmed,
             check_service_smtp_addr: smtp_addr,
+            dig_runner: Box::new(RealDigRunner),
         })
+    }
+
+    #[cfg(test)]
+    fn with_dig_runner(mut self, runner: Box<dyn DigRunner + Send + Sync>) -> Self {
+        self.dig_runner = runner;
+        self
     }
 
     fn curl_probe(&self) -> Result<bool, Box<dyn std::error::Error>> {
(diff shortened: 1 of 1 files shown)


uzyn (commented):
## Changes Overview

`RealNetworkOps` previously called `dig +short +tries=1 +time=2 <TYPE> <domain>` once per query against the system's recursive resolver, ignored dig's exit status, and returned `Ok(vec![])` whenever stdout was empty — so a single dropped UDP packet looked identical to a NOERROR/NXDOMAIN "no record" answer. This PR introduces a `DigRunner` trait + `dig_with_cascade` helper that pins each query to an explicit public resolver (`1.1.1.1` -> `8.8.8.8` -> `9.9.9.9`), retries up to 3 times per resolver on non-zero exit (with 300 ms backoff), treats exit-0 as authoritative (stops cascading even on empty output), and only returns `io::Error` when every resolver is exhausted. The four `resolve_*` methods now propagate that error via `?`, so `verify_*` renders genuine lookup failures as `Fail(...)` rather than silently as `Missing(...)`. 10 new unit tests script a mock `DigRunner` to exercise every cascade path without shelling out.

## Scope Alignment

Implementation matches the PR description and acceptance checklist exactly. The fix is fully contained in `src/setup.rs`; the public `NetworkOps` trait and the `verify_*` arms are unchanged at the type level — only the `Err` arm of `resolve_*` is now reachable for transient failures. No scope creep, no unrelated refactors.

## Potential Bugs

None found. Walked the cascade end-to-end:

- `dig_with_cascade` correctly stops on the first `Ok(output)` with `status.success()` (including empty stdout — verified by `dig_cascade_empty_but_ok_is_success_no_fallback`).
- The `attempt + 1 < DIG_RETRY_ATTEMPTS` guard prevents the post-final-attempt sleep, so the 300 ms delay only sits between intra-resolver retries — not after the last attempt of the last resolver.
- `last_err_detail` is correctly threaded; the final error mentions the last failure encountered.
- The `with_dig_runner` builder is `#[cfg(test)]`-gated so it cannot leak into production binaries.
- dig's exit-code semantics (0 for NOERROR/NXDOMAIN, non-zero for SERVFAIL/timeout) are correctly leveraged — SERVFAIL won't be silently swallowed as "no record."

## Security Issues

None. The cascade hardcodes well-known public anycast resolvers (no user-supplied input flowing into the resolver address), and `dig` is invoked with the same fixed-shape argv as before plus the `@<resolver>` prefix — no shell, no injection surface introduced.

## Test Coverage

Strong. The 10 new tests cover:

- argv shape (incl. the new `@<resolver>` token)
- resolver order invariant
- happy path (first resolver wins on first try)
- intra-resolver exhaustion + fallthrough to next resolver
- full exhaustion -> `Err`
- empty-but-OK -> success without fallback (the latency-saving authoritative-empty case)
- spawn-error retry behaviour distinct from non-zero exit
- `resolve_a` integration: end-to-end propagation of cascade `Err` through `RealNetworkOps`
- `resolve_a` integration: end-to-end happy-path with second-resolver fallthrough
- `verify_a` integration: cascade `Err` correctly maps to `Fail(...)` not `Missing(...)`

Verified locally: full `setup::` test suite (228 tests) passes; `cargo build`, `cargo clippy -- -D warnings`, `cargo fmt -- --check` all clean.

## Code Quality

Clean. Trait-based testing pattern matches the existing `SystemOps` / `NetworkOps` / `MailTransport` convention used elsewhere in the codebase. The `Debug` impl for `RealNetworkOps` is hand-written (necessary because of the trait-object field) and preserves the existing `.unwrap_err()` test usages. `dig_retry_delay()` is a tasteful `#[cfg(test)]` shim — the all-resolvers-fail test would otherwise sit on 8 × 300 ms = 2.4 s of real sleep.

The new `ErroringNet` impl inside `verify_a_maps_resolver_err_to_fail_not_missing` duplicates much of `MockNetworkOps`'s shape because `MockNetworkOps`'s `resolve_*` arms can't model `Err`. Acceptable as-is, but one alternative would be to add an optional `force_resolve_err: Option<String>` field to `MockNetworkOps` and reuse it.

- **Blockers**: none.
- **Non-blockers**: none.
- **Nice-to-haves**:
  - With 3 resolvers × 3 attempts × ~2s per dig invocation + intra-resolver sleeps, the worst-case per-query latency rises from ~2 s to ~20 s, and `verify_all_dns` issues 6 queries — so in the pathological all-resolvers-down case `aimx status` could block for ~2 minutes. The previous compile-time `DIG_TIMEOUT_SECS * DIG_TRIES <= 5` guard was intentionally dropped because the cascade changes what "budget" means; mentioning here only as a future enhancement (an explicit overall deadline, or `verify_all_dns` parallelism, both already noted as out of scope).
  - Consider extending `MockNetworkOps` with an opt-in error injection path so the `verify_a_maps_resolver_err_to_fail_not_missing` test doesn't need its own one-off `NetworkOps` impl.

```
fix: DNS verification no longer flaps between PASS and MISSING

Pin dig queries to explicit public resolvers (1.1.1.1 -> 8.8.8.8 ->
9.9.9.9) with up to 3 retries per resolver on non-zero exit. Exit-0
responses (including empty NOERROR/NXDOMAIN answers) stay authoritative
and stop the cascade. resolve_* now propagates Err on full exhaustion
so verify_* renders genuine DNS lookup failures as Fail rather than
Missing. Adds an injectable DigRunner trait and 10 unit tests covering
every cascade path.
```

PR #95, opened 2026-04-19
fix: order DNS verification output to match display table

uzyn:
## Summary
The `[DNS]` section of `aimx setup` prints records in the order A, [AAAA], MX, SPF, DKIM, DMARC. The subsequent "DNS Verification:" block printed MX before A, which makes it harder to scan results against the table. Reorder `verify_all_dns` so the verification output matches the display order.

## Requirements
- [x] `verify_all_dns` returns results in the order `A, [AAAA], MX, SPF, [SPF (IPv6)], DKIM, DMARC` — matching `generate_dns_records`.
- [x] IPv6 pairs are preserved: AAAA immediately follows A, SPF (IPv6) immediately follows SPF.
- [x] `display_dns_verification` / `dns_verification_lines` render correctly with the new order (they iterate in input order).
- [x] New unit test locks the ordering so a future regression is caught.

## Out of scope
- Changing the `[DNS]` display table order in `generate_dns_records` — we're matching TO it, not changing it.
- Refactoring the verify_* functions themselves.
- Any change to `src/status.rs`, which consumes `verify_all_dns` results and iterates in order — it inherits the new order for free.

## Technical approach
Surgical reorder inside `src/setup.rs::verify_all_dns`:

```
before: [MX, A] → [AAAA if ipv6] → SPF → [SPF (IPv6) if ipv6] → [DKIM, DMARC]
after:  [A] → [AAAA if ipv6] → [MX] → SPF → [SPF (IPv6) if ipv6] → [DKIM, DMARC]
```

No public signatures change; no call sites need updates.

## Test plan
- [x] New unit test `verify_all_dns_orders_results_to_match_display` asserts the exact returned name order for both IPv4-only and IPv4+IPv6 variants.
- [x] `cargo test` — 782 unit + 64 integration tests pass.
- [x] `cargo clippy -- -D warnings` clean.

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -858,15 +858,13 @@ pub fn verify_all_dns(
     local_dkim_pubkey: Option<&str>,
 ) -> Vec<(String, DnsVerifyResult)> {
     let ip_str = server_ip.to_string();
-    let mut results = vec![
-        ("MX".into(), verify_mx(net, domain)),
-        ("A".into(), verify_a(net, domain, server_ip)),
-    ];
+    let mut results = vec![("A".into(), verify_a(net, domain, server_ip))];
 
     if let Some(ipv6) = server_ipv6 {
         results.push(("AAAA".into(), verify_aaaa(net, domain, ipv6)));
     }
 
+    results.push(("MX".into(), verify_mx(net, domain)));
     results.push(("SPF".into(), verify_spf(net, domain, &ip_str)));
 
     if let Some(ipv6) = server_ipv6 {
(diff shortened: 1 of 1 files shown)


uzyn (commented):
## Changes Overview

Surgical reorder inside `src/setup.rs::verify_all_dns` so the verification output matches the `[DNS]` display-table order from `generate_dns_records`: `A, [AAAA], MX, SPF, [SPF (IPv6)], DKIM, DMARC`. The previous order emitted MX before A, which made scanning the verification block against the display table harder. One new unit test pins the order for both IPv4-only and IPv4+IPv6 configurations.

## Scope Alignment

Exactly matches the PR description:
- `verify_all_dns` now returns the promised order.
- IPv6 pairing is preserved — `AAAA` immediately after `A`; `SPF (IPv6)` immediately after `SPF`.
- Public signature unchanged; no call sites touched.
- Both consumers (`src/setup.rs::run_setup` retry loop at line 1509 and `src/status.rs::gather_dns` at line 107) iterate in input order via `dns_verification_record_lines`, so they inherit the new order without modification.
- No change to `generate_dns_records`, which is explicitly the fixed point being matched.
- No positional indexing into the results vec anywhere in the crate — the reorder is safe.

## Potential Bugs

None. The reordered calls are independent side-effect-free verify_* functions whose results do not depend on each other's invocation order.

## Test Coverage

- New test `verify_all_dns_orders_results_to_match_display` locks the exact name sequence for both IPv4-only (`A, MX, SPF, DKIM, DMARC`) and IPv4+IPv6 (`A, AAAA, MX, SPF, SPF (IPv6), DKIM, DMARC`) variants. This is the right level of assertion — it fails loudly on any future re-shuffle.
- Existing tests (`verify_all_dns_all_pass`, `verify_all_dns_partial_fail`, `verify_all_dns_with_ipv6`, `verify_all_dns_without_ipv6_has_no_aaaa`, the `display_dns_verification_*` tests) use order-agnostic assertions (`iter().all` / `iter().any` / count), so they remained valid and did not require updates.

Independently verified locally on this worktree:
- `cargo test --bin aimx` — 782 passed, 0 failed.
- `cargo test --test integration` — 64 passed, 0 failed.

## Security Issues

Not applicable — no I/O, auth, network, or data-flow changes; purely a reorder of in-process verification dispatch.

## Code Quality

The refactor from a two-element `vec![MX, A]` into a single-element `vec![A]` followed by a conditional `AAAA` push and then an `MX` push is a mild stylistic increase in line count but justified: it keeps the IPv4/IPv6 pair contiguous, which was the whole point of the reorder. Readable.

Minor nit (not worth changing): the new test's leading comment says the display table includes `TXT(SPF IPv6)` as a separate row, whereas `generate_dns_records` actually emits one SPF TXT record whose value branches on IPv6 presence; `verify_all_dns` is the place where SPF and SPF (IPv6) become two separate check entries. The comment is accurate about verification-output order (which is what the test pins) even if slightly loose about the DNS-records table. Non-blocking.

- **Blockers**: none
- **Non-blockers**: none
- **Nice-to-haves**: none

```
fix: order DNS verification output to match display table

Reorder verify_all_dns to return A, [AAAA], MX, SPF, [SPF (IPv6)], DKIM,
DMARC — matching generate_dns_records — so the "DNS Verification:" block
reads in the same order as the [DNS] display table. Adds a unit test
pinning the order for both IPv4-only and IPv4+IPv6 configurations. No
public signatures change; status.rs and the setup retry loop inherit
the new order via the existing input-order iteration in
dns_verification_record_lines.
```

PR #104, opened 2026-04-20
feat(setup): default DNS-ownership prompt to Yes

uzyn:
## Summary
After the user has entered a domain during `aimx setup`, the follow-up "Do you control this domain and have access to its DNS settings?" is a go/no-go gate on the happy path — declining is the exceptional answer. Flip the default from N to Y so pressing Enter proceeds.

## Requirements
- [x] DNS-ownership prompt text changes from `(y/N)` to `(Y/n)`.
- [x] Empty input (bare Enter) is treated as **yes** and setup proceeds.
- [x] Any input starting with `n` or `N` (case-insensitive) is treated as **no** and setup is cancelled with the existing error message.
- [x] Explicit `y` / `Y` / `yes` continues to be treated as **yes** (unchanged happy path).
- [x] No other `y/N` prompts modified.

## Out of scope
- Flipping destructive confirmations (`aimx hooks delete`, `aimx mailboxes delete`, `aimx mailboxes delete --force`, `aimx agent-setup` overwrite, `aimx uninstall`). Convention across Unix tooling is default-No on destructive prompts; each already has `--yes` / `--force` for scripted use.
- `docs/sprint.2.md:642` — historical sprint doc, marked `[x]` done. Per `CLAUDE.md`, `[DONE]` sprint entries are not modified.
- `book/` docs do not quote the DNS-ownership prompt text, so no doc updates were needed.

## Technical approach
Single edit site: `prompt_domain()` in `src/setup.rs`.

- Prompt text: `(y/N) ` → `(Y/n) `.
- Decline logic: from `!input.eq_ignore_ascii_case("y")` (which rejected empty and anything non-y) to an explicit "first char is n/N" check, so empty input falls through to the accept path.

No changes to callers — `run_setup()` already handles the `Err` returned on decline.

## Test plan
- [x] Existing `prompt_domain_accepts_valid_domain_with_confirmation` (explicit `y`) — passes.
- [x] Existing `prompt_domain_exits_on_declined_confirmation` (explicit `n`) — passes.
- [x] New `prompt_domain_accepts_empty_confirmation_as_yes` — bare Enter now returns `Ok`.
- [x] New `prompt_domain_exits_on_uppercase_n` — uppercase `N` still declines.
- [x] Full `cargo test` (6 unit + 85 integration) — green.

## Notes
During the audit, five other `y/N` prompts were found and intentionally **not** flipped (all destructive): hooks delete, mailbox delete (non-force + `--force`), agent-setup overwrite, uninstall. Reviewer: please confirm you agree destructive prompts should stay default-No.

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -1421,12 +1421,17 @@ pub fn prompt_domain(reader: &mut dyn BufRead) -> Result<String, Box<dyn std::er
 
     print!(
         "You will need to add MX, SPF, and DKIM DNS records for this domain.\n\
-         Do you control this domain and have access to its DNS settings? (y/N) "
+         Do you control this domain and have access to its DNS settings? (Y/n) "
     );
     io::stdout().flush()?;
     let mut confirm = String::new();
     reader.read_line(&mut confirm)?;
-    if !confirm.trim().eq_ignore_ascii_case("y") {
+    if confirm
+        .trim()
+        .chars()
+        .next()
+        .is_some_and(|c| c == 'n' || c == 'N')
+    {
         return Err("Setup cancelled. You need DNS access to proceed.".into());
     }
 
@@ -3153,6 +3158,26 @@ mod tests {
     #[test]
     fn prompt_domain_exits_on_declined_confirmation() {
         let input = b"agent.example.com\nn\n";
+        let mut reader = io::Cursor::new(input);
+        let result = prompt_domain(&mut reader);
+        assert!(result.is_err());
+        assert!(
+            result.unwrap_err().to_string().contains("cancelled"),
+            "should indicate cancellation"
+        );
+    }
+
+    #[test]
+    fn prompt_domain_accepts_empty_confirmation_as_yes() {
+        let input = b"agent.example.com\n\n";
+        let mut reader = io::Cursor::new(input);
+        let domain = prompt_domain(&mut reader).expect("empty confirmation should default to yes");
+        assert_eq!(domain, "agent.example.com");
+    }
+
+    #[test]
+    fn prompt_domain_exits_on_uppercase_n() {
+        let input = b"agent.example.com\nN\n";
         let mut reader = io::Cursor::new(input);
         let result = prompt_domain(&mut reader);
         assert!(result.is_err());


uzyn (commented):
## Changes Overview

Flips the DNS-ownership confirmation in `aimx setup`'s `prompt_domain()` from default-No to default-Yes. Prompt text goes from `(y/N)` to `(Y/n)`, and the decline logic changes from "accept only if trimmed input equals `y`" to "decline only if the first non-whitespace character is `n`/`N`". Adds two unit tests covering bare-Enter-is-yes and uppercase-N-still-declines.

## Scope Alignment

Implementation matches the PR description exactly. Single edit site in `src/setup.rs::prompt_domain()`; no other prompts touched; the five destructive `y/N` prompts (`hooks delete`, `mailboxes delete`, `mailboxes delete --force`, `agent-setup` overwrite, `uninstall`) remain default-No as claimed. I grepped for `y/N` / `Y/n` patterns across `src/` and confirmed only the target prompt was changed. `book/` has no references to the prompt text so no doc updates are warranted.

## Potential Bugs

None. I walked through the new predicate `confirm.trim().chars().next().is_some_and(|c| c == 'n' || c == 'N')` against every reachable input class:

- Empty (`""`) → `chars().next()` is `None` → `is_some_and` returns `false` → accept. Correct.
- Whitespace-only (`"   \n"`) → trims to `""` → accept. Correct.
- `y` / `Y` / `yes` → first char not `n`/`N` → accept. Correct, preserves old happy path.
- `n` / `N` / `no` / `nope` → first char is `n`/`N` → decline. Correct.
- Multi-byte UTF-8 leading char (e.g. a user pasting a smart quote): `chars().next()` returns a `char` (scalar value), not a byte — no panic, the char is not `n`/`N` so it accepts. Reasonable for a default-Yes prompt.
- Arbitrary text (`"maybe"`, `"0"`, random noise) → accept. This is looser than the old `eq_ignore_ascii_case("y")` which accepted only exact `y`/`Y`, but it is consistent with the PR's stated default-Yes intent: anything not clearly a "no" proceeds.

No crash paths, no correctness regression, no data-flow issues.

## Security Issues

None. This is a local interactive prompt with no attacker-controlled input surface.

## Test Coverage

Adequate for the change. The four tests together cover the meaningful equivalence classes:

- Lowercase `y` (existing) — explicit accept
- Lowercase `n` (existing) — explicit decline
- Empty input (new) — new default-Yes behavior
- Uppercase `N` (new) — case-insensitive decline still works

Two minor gaps worth noting, none blocking:

- No test for uppercase `Y` explicitly. The old check used `eq_ignore_ascii_case`, the new one uses explicit `'n' || 'N'` (accept is implicit), so `Y` is accepted by falling through. Given that empty-is-yes is tested, `Y` naturally works by the same path, so this is fine.
- No test for whitespace-only input (e.g. a user who types a space then Enter). It behaves correctly by inspection (trims to empty, accepts), and this is the same behavior as bare Enter, so it's covered in spirit.

## Code Quality

The new predicate is a touch more verbose than strictly necessary — `trim().starts_with(['n', 'N'])` would read slightly more naturally than a `chars().next().is_some_and(...)` chain. Not a change request; the current form is clear enough and handles the empty case explicitly which is arguably more obvious than relying on `starts_with`'s empty-slice behavior.

## Scope Decision: Destructive Prompts

Reviewer confirming the tradeoff the PR asked about: **agree that the destructive prompts should remain default-No.** The distinction is sound:

- The DNS-ownership prompt is a **precondition check** on an otherwise safe, idempotent setup flow — declining just aborts, nothing is modified. Default-Yes reduces friction for the overwhelmingly common case.
- The five left-alone prompts all gate **irreversible data loss** (deleting hooks, mailboxes with emails, overwriting agent configs, uninstalling the service). Unix convention is default-No for destructive operations, each already has a `--yes`/`--force` escape hatch for scripted use, and a mispress of Enter here would be genuinely painful. Keeping those default-No is the correct call.

- Blockers: none
- Non-blockers: none
- Nice-to-haves: none

```
feat(setup): default DNS-ownership prompt to Yes

Flip the post-domain confirmation in `aimx setup` from `(y/N)` to
`(Y/n)` so pressing Enter on the happy path proceeds. Decline logic
changes to "first char is n/N" so empty input now accepts. Destructive
prompts (hooks delete, mailbox delete, agent-setup overwrite,
uninstall) intentionally remain default-No per Unix convention.

Adds two unit tests: bare-Enter accepts, uppercase N declines.
```

PR #108, opened 2026-04-20
refactor(hooks): drop content filters, rename id to name, make name optional

uzyn:
## Summary

Strip the content-based match filters (`from`, `to`, `subject`,
`has_attachment`) from hooks so matching is driven solely by event + the
`trusted` gate. Rename the hook `id` field to `name` everywhere (config,
wire protocol, env var, structured log field, CLI, Rust API). Make `name`
optional: when omitted, the daemon derives a stable 12-char hex `name`
from `sha256(event + cmd + dangerously_support_untrusted)` at use —
no `config.toml` mutation on load.

## Requirements

- [x] Remove match filters from the `Hook` struct and its config schema
  (`from`, `to`, `subject`, `has_attachment` dropped).
- [x] `#[serde(deny_unknown_fields)]` on `Hook` so stale filter fields or
  typos (`subjct = "..."`) fail loudly with the field name called out.
- [x] Rename `id` → `name` everywhere: struct field, TOML key, env var
  (`AIMX_HOOK_NAME`), log field (`hook_name=`), wire header (`Hook-Name:`),
  Rust fns/consts (`is_valid_hook_name`, `HOOK_NAME_MAX_LEN`), and CLI
  (`hooks delete <name>`, `NAME` column in `hooks list`).
- [x] New name charset validator matches `^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$`.
- [x] `name` is optional. `derive_hook_name()` returns the first 12 hex
  chars of `sha256(event || 0x1F || cmd || 0x1F || dangerous_byte)`,
  stable across restarts; no write-back to `config.toml`.
- [x] Uniqueness enforced globally across mailboxes for every effective
  name with distinct error strings for explicit-vs-explicit,
  derived-vs-derived, and mixed collisions.
- [x] `aimx hooks create` grows an optional `--name <name>` flag; the
  old `--from` / `--to` / `--subject` / `--has-attachment` flags are
  gone. When `--name` is omitted, the CLI writes no `name =` line and
  prints the derived name on success.
- [x] `aimx hooks list` shows the effective name (derived names included).
- [x] `aimx hooks delete <name>` resolves against effective names for
  both explicit and anonymous hooks.
- [x] Documentation updated: `book/hooks.md` rewritten (properties table,
  env-var tables, log-format block, examples — at least one sample uses
  an anonymous hook with the `# name is optional — ...` comment); the
  "Match filters" section is gone. `book/hook-recipes.md` rewritten:
  `id = "..."` → `name = "..."`, `\$AIMX_HOOK_ID` → `\$AIMX_HOOK_NAME`,
  the filter-based "notify on delivered marketing mail only" recipe is
  rewritten to do the recipient match inside a `case` shell guard, and
  the Hermes sample now demonstrates the anonymous-hook convention.
  `README.md`, `book/faq.md`, `docs/manual-setup.md`, `docs/prd.md`,
  and project `CLAUDE.md` updated to reflect the new schema.

## Out of scope

- Adopting `toml_edit` so `write_config_atomic` preserves comments and
  unknown fields. (Known v1 limitation; tracked separately.)
- Any change to `dangerously_support_untrusted` — stays as the trust-gate
  escape hatch.
- Any change to the trust model itself (`trust` modes, `trusted_senders`,
  DKIM/SPF/DMARC).
- Adding filter capability back in a different form (regex, body match,
  header match, etc.). Deferred.
- Per-hook `AIMX_HOOK_ID` backward-compat alias. Pre-launch; clean break.

## Technical approach

- `src/hook.rs`: `Hook` shrinks to `{ name: Option<String>, event, type,
  cmd, dangerously_support_untrusted }` with `deny_unknown_fields`.
  `derive_hook_name()` uses `sha2` (already a direct dep) with 0x1F
  separators between the three inputs. `effective_hook_name()` is the
  single call site for "what the operator sees / greps."
- `src/config.rs::validate_hooks`: builds a `HashMap<effective_name,
  (mailbox, is_explicit)>` and emits distinct error strings for the
  three collision classes. New `validate_single_hook()` helper used by
  the daemon handler for pre-disk validation of submitted stanzas.
- Wire: `Hook-Id:` header → `Hook-Name:`; `HookDeleteRequest::id` →
  `name`. Handlers resolve deletes against the effective name.
- CLI: `HookCreateArgs` drops the four filter flags; `Delete { id }` →
  `Delete { name }`.

## Test plan

- New unit coverage in `src/hook.rs` for `is_valid_hook_name` boundaries,
  determinism / divergence / charset-safety of `derive_hook_name`, and
  `effective_hook_name` prefers explicit over derived.
- New unit coverage in `src/config.rs::validate_hooks` for the three
  collision classes (explicit-vs-explicit, anonymous-vs-anonymous with
  the "set an explicit name" hint, explicit-vs-derived), plus
  `deny_unknown_fields` covering both stale filters (`from = "..."`) and
  typos (`subjct = "..."`).
- New unit coverage in `src/hooks.rs` (CLI) for `--name` validation,
  anonymous-hook flow, and `find_hook_by_effective_name` matching both
  explicit and derived names.
- Updated unit coverage in `src/hook_handler.rs` — anonymous submission
  round-trips, delete-by-derived-name succeeds, duplicate-name collision
  is rejected at `VALIDATION`.
- Updated `src/send_protocol.rs` tests cover the `Hook-Name:` header
  rename, missing/empty/case-insensitive handling, and round-trip of
  the renamed `HookDeleteRequest { name }`.
- `tests/integration.rs` switched from `id =` to `name =` in every
  hook-using fixture; dropped `--from` / `--to` / `--has-attachment` /
  `--subject` flags from the `hooks create` tests; added new
  `hooks_create_anonymous_prints_derived_name` covering the "no `name =`
  written" path end-to-end; dropped filter-related hooks from the
  `mailboxes show` fixture.
- Ran: `cargo fmt -- --check` clean, `cargo clippy --tests -- -D warnings`
  clean, `cargo test --bin aimx` → 848 passed / 0 failed, `cargo test
  --test integration` → 83 passed / 0 failed.

## Notes

- `AfterSendContext::has_attachment` was removed (only ever read by the
  now-deleted filter path); `send_handler.rs` no longer populates it.
- `docs/sprint.md` historical DONE entries deliberately left untouched per
  the project convention not to rewrite completed-sprint history.


uzyn (commented):
## Changes Overview

This PR refactors the hook subsystem by (a) removing all content-based match filters (`from`, `to`, `subject`, `has_attachment`) so hook dispatch is driven purely by `event` + the `on_receive` trust gate, (b) renaming the hook identifier field `id` → `name` across the entire surface (TOML schema, UDS wire protocol `Hook-Name:` header, env var `AIMX_HOOK_NAME`, structured log field `hook_name=`, CLI flag, Rust API), and (c) making `name` optional by deriving a stable 12-hex-char name via `sha256(event || 0x1F || cmd || 0x1F || dangerous_byte)` when omitted — with no write-back to `config.toml`. `#[serde(deny_unknown_fields)]` is added to `Hook` so stale filter fields or typos fail loudly at load. Uniqueness is now enforced globally across mailboxes with distinct errors for explicit-vs-explicit, anonymous-vs-anonymous, and mixed collisions.

## Scope Alignment

Implementation matches the stated PR intent on the code side: the `Hook` struct is stripped down, `deny_unknown_fields` is in place, `derive_hook_name` / `effective_hook_name` are implemented and tested, `validate_hooks` emits the three collision classes with hint strings, the wire protocol cleanly renames `Hook-Id:` → `Hook-Name:`, the CLI drops the four filter flags and grows `--name`, and `delete <name>` resolves against effective names for both explicit and anonymous hooks. Legacy-schema and stale-filter paths are deleted — there is no "compat shim," which matches the pre-launch convention.

**Scope gap:** The PR description explicitly claims `book/hooks.md`, `book/hook-recipes.md`, `README.md`, `book/faq.md`, `docs/manual-setup.md`, `docs/prd.md`, and `CLAUDE.md` were updated. Two user-facing guides that describe the current schema were **not** updated and now contradict the implementation:

- `book/configuration.md` lines 103–111: the Hook settings table still lists `id`, `from`, `to`, `subject`, `has_attachment` as valid schema fields, and the "Full config example" (lines 218, 236, 242, 245–246) still uses `id = "notifyall001"`, `id = "supportlog01"`, `id = "supportgmai"`, `from = "*@gmail.com"`, `has_attachment = true`. A user who copy-pastes this example will hit `unknown field` errors from `deny_unknown_fields` before the daemon even starts.
- `book/cli.md` lines 136, 144, 151–154, 157, 159: the `hooks list` table still says it prints `ID` and `FILTERS` columns, `hooks create` still documents `--from`, `--to`, `--subject`, `--has-attachment`, and `hooks delete <id>` is still described with an `id` argument. These flags no longer exist on the binary — a user following this doc will get a clap parse error.

Both files are user-facing guides in `book/` and per the project CLAUDE.md "When making changes that affect CLI behavior, setup output, or MCP tools, update the corresponding guide files too." This is not a minor omission — a brand new user who starts in `book/` will copy a config that fails to parse.

Historical files (`docs/sprint.md`, `docs/sprint.1.md`, `docs/sprint.4.md`, `docs/idea.md`, `docs/manual-test.md`) correctly left untouched per the project convention.

## Potential Bugs

### Non-blocker: `apply_delete_direct` and `handle_hook_delete` silently stop at the first match

**Blocker?** No — uniqueness is enforced at `validate_hooks`, so in a well-formed config there can only be one hook with a given effective name.

`src/hooks.rs:209-225` (`apply_delete_direct`) and `src/hook_handler.rs:160-167` (`handle_hook_delete`) both `break` after finding the first mailbox whose `retain` shrank its `hooks` list. If a corrupt config somehow got past validation and had two hooks sharing an effective name, delete would remove only one and silently leave the other. Validation normally catches that at `Config::load`, so the code path is unreachable in practice — hence non-blocker. Worth a sentence comment noting the reliance on `validate_hooks`.

### Non-blocker: `derive_hook_name` does not mix the mailbox name into the hash

The PR deliberately hashes only `(event, cmd, dangerous_byte)`, which means two mailboxes with identical anonymous hooks collide and must be explicitly named to disambiguate. This is documented in the design (the "set an explicit `name`" error hint), tested (`load_rejects_two_anonymous_hooks_with_same_identity`), and — depending on product intent — may be exactly right or surprising. Flagging here for the author's awareness, not as a defect.

## Security Issues

No new issues introduced. The removal of content filters actually **reduces** attack surface by deleting the glob matcher. Env vars are still passed to subprocesses via `Command::env` (not interpolated into the command string), and `env_clear` + explicit `PATH` / `HOME` passthrough are preserved. Injection regression tests (`env_var_preserves_backtick_injection`, `env_var_preserves_dollar_paren_injection`) are retained. The `dangerously_support_untrusted` gate is unchanged. `deny_unknown_fields` is a defence-in-depth win — a typo like `subjct = ...` that previously would have been silently ignored now refuses to load.

## Test Coverage

Unit coverage is thorough for the scope:
- `is_valid_hook_name` boundary cases (empty, max-length, overflow, leading dot/dash, unicode).
- `derive_hook_name` determinism, divergence by event / cmd / dangerous flag, charset safety.
- `effective_hook_name` prefers explicit, falls back to derived.
- `validate_hooks` covers all three collision classes and `deny_unknown_fields` for both stale filters and typos.
- Daemon handler covers anonymous round-trip, delete-by-derived-name, duplicate-name collision, unknown mailbox, malformed TOML, and concurrent cross-mailbox create (lost-update regression mirror of the existing MAILBOX path).
- Integration test `hooks_create_anonymous_prints_derived_name` confirms the "no `name =` written" invariant end-to-end by parsing the on-disk TOML.

**Gap (non-blocker):** I could not find a test that exercises a config with two mailboxes holding hooks with the same effective name being handed to `apply_delete_direct` / `handle_hook_delete`. Since validation blocks this at `Config::load`, the test is not essential — but pairing the `break`-after-first-match logic with a unit test that asserts the validator prevents the ambiguity would make the invariant explicit.

**Gap (nice-to-have):** No test for `hooks create` / `hooks delete` printing the derived name when invoked anonymously through the daemon UDS path (integration test covers the direct-edit fallback; handler tests cover `AckResponse::Ok` but not the CLI's printout path through `submit_hook_create_via_daemon`). Low priority.

## Code Quality

Clean. The renames are mechanical and consistent, `Hook::name` is `Option<String>` with `skip_serializing_if = Option::is_none` which correctly preserves the "omitted → stays omitted" round-trip, and the daemon handler re-runs the full `validate_hooks` after appending (belt-and-suspenders). The CLI fallback path goes through the same `validate_hooks` before writing. `HookCrudFallback` cleanly distinguishes socket-missing from daemon-rejected errors.

One small nit (non-blocker): the `derive_hook_name` loop uses `DERIVED_HOOK_NAME_LEN.div_ceil(2)` = 6 and then `out.truncate(DERIVED_HOOK_NAME_LEN)`. Because `DERIVED_HOOK_NAME_LEN` is 12 (even), the truncate is a no-op — `.take(6)` already produces exactly 12 hex chars. Not wrong, just slightly obfuscated for an even constant.

1. Update `book/configuration.md` — remove `id` / `from` / `to` / `subject` / `has_attachment` rows from the "Hook settings" table; rewrite the Full config example so it parses under `deny_unknown_fields` (replace `id = ...` with `name = ...`, drop `from`, drop `has_attachment`). A user copy-pasting the current example gets a config that fails to load.
2. Update `book/cli.md` — drop `--from`, `--to`, `--subject`, `--has-attachment` from the `hooks create` flag table; rename `hooks delete <id>` to `hooks delete <name>`; update the `hooks list` description to say the table prints `NAME MAILBOX EVENT CMD` (no `FILTERS` column). These flags no longer exist and the command errors out if anyone follows this doc.

**Non-blockers (nice to address):**
- Add a one-line comment to `apply_delete_direct` and `handle_hook_delete` noting that the `break`-after-first-match is safe because `validate_hooks` guarantees effective-name uniqueness.
- Consider either a test that rides through `submit_hook_create_via_daemon` and confirms the CLI prints the derived name for anonymous hooks, or explicitly note the gap in a follow-up issue.
- Simplify the `derive_hook_name` body — since `DERIVED_HOOK_NAME_LEN` is even, `out.truncate(...)` is dead code.

- Confirm intent on whether anonymous-hook identity should include mailbox name (current design: no — two mailboxes with identical anonymous hooks must set an explicit name to coexist). Documented in error hint; just flagging for future-you.

uzyn:
## Review cycle 1 — fixes pushed (56dcfa4)

### Blockers fixed

- **`book/configuration.md`** — Hook settings table rewritten: dropped `id` / `from` / `to` / `subject` / `has_attachment` rows; added optional `name` with the new charset and derivation note. "Full config example" rewritten so every hook entry parses under `deny_unknown_fields`: one anonymous hook on `catchall` with the canonical `# name is optional — a stable 12-char hex id is derived from event+cmd if omitted` comment, the other hooks renamed to `name =` and stripped of the removed filter fields.
- **`book/cli.md`** — `hooks create` flag table: dropped `--from` / `--to` / `--subject` / `--has-attachment`; added optional `--name`. `hooks delete <id>` → `hooks delete <name>` (both explicit and derived names resolve). `hooks list` description updated to print `NAME MAILBOX EVENT CMD` (no `FILTERS` column).

### Non-blockers addressed

- **`apply_delete_direct` / `handle_hook_delete`** — one-line comment at each `break` tying correctness to the `validate_hooks` global-uniqueness guarantee.
- **`derive_hook_name` cosmetic cleanup** — removed the dead `out.truncate(DERIVED_HOOK_NAME_LEN)` and added a comment noting that `DERIVED_HOOK_NAME_LEN` is even, so `ceil(len/2)` bytes hex-encoded already produces exactly `DERIVED_HOOK_NAME_LEN` chars.
- **Intent comment on anonymous-hook identity** — added a doc comment on `derive_hook_name` stating that mailbox name is deliberately excluded from the hash; two mailboxes with identical `(event, cmd, dangerously_support_untrusted)` collide, forcing the operator to set an explicit `name`. Cross-references the collision error in `validate_hooks`.
- **New UDS-path E2E test** — `hooks_create_anonymous_prints_derived_name_via_daemon` mirrors the existing direct-edit variant but goes through the daemon UDS path (`submit_hook_create_via_daemon`). Spins up `aimx serve`, runs `aimx hooks create` without `--name`, asserts the CLI prints the expected derived name (computed inline via `sha2`), and reads back the daemon-rewritten `config.toml` to confirm no `name =` line was written.

### Verification

- `cargo clippy --tests -- -D warnings` — clean
- `cargo test --bin aimx` — 848 passed / 0 failed / 3 ignored
- `cargo test --test integration` — 84 passed / 0 failed (the new `_via_daemon` test brings the count from 83 → 84)

### Intentionally left as-is

- Nothing. All four non-blockers from the review were addressed in this cycle.

uzyn (commented):
## Re-review cycle 2 — verdict: APPROVE

All cycle-1 blockers and non-blockers are resolved, and no new issues were introduced.

### Resolved

**Blockers**
- `book/configuration.md` — the Hook settings table and full config example now match the new schema. Removed rows for `id` / `from` / `to` / `subject` / `has_attachment`; added optional `name` with the correct charset and derivation note. Verified the full config example parses under `#[serde(deny_unknown_fields)]` + `validate_hooks` by staging it as `/etc/aimx/config.toml` and running `aimx hooks list` — config loads cleanly and the anonymous hook resolves to a derived name (`d76229eff2e9`), with the two `support` hooks printing their explicit names.
- `book/cli.md` — `hooks create` drops `--from` / `--to` / `--subject` / `--has-attachment` and adds `--name`; `hooks delete <id>` → `<name>`; `hooks list` docs describe `NAME MAILBOX EVENT CMD` (no `FILTERS`). Cross-checked against `aimx hooks create --help` and the real `hooks list` output — both agree.

**Non-blockers**
- Uniqueness-invariant comments at each `break` in `apply_delete_direct` (`src/hooks.rs:217`) and `handle_hook_delete` (`src/hook_handler.rs:165`). Tie correctness explicitly to `validate_hooks`.
- Dead `out.truncate(DERIVED_HOOK_NAME_LEN)` removed from `derive_hook_name`; added a one-line comment explaining that `DERIVED_HOOK_NAME_LEN = 12` is even so `div_ceil(2) = 6` bytes hex-encoded produces exactly 12 chars. Verified the math holds.
- Function-level doc on `derive_hook_name` now states the mailbox name is deliberately excluded from the hash (two mailboxes with the same `(event, cmd, dangerous)` collide under `validate_hooks`, forcing explicit `name =`).
- New integration test `hooks_create_anonymous_prints_derived_name_via_daemon` exercises the UDS/daemon path end-to-end, asserts the derived 12-hex-char name is printed, and confirms the daemon does not write `name =` to `config.toml`. Passes locally.

### Verification

- `cargo clippy --tests -- -D warnings` — clean
- `cargo test --bin aimx` — 848 passed, 3 ignored
- `cargo test --test integration` — 84 passed (one flaky failure on the first parallel run in `after_send_hook_fires_with_delivered_status` due to a UDS socket race; reruns clean at `--test-threads=4` and in isolation — unrelated to this PR)
- End-to-end doc-example parse check: `aimx hooks list` against the exact TOML in `book/configuration.md` succeeds and resolves the anonymous hook's derived name.

### Still unresolved / New issues

None.

~~~
refactor(hooks): drop content filters, rename id to name, make name optional (#108)

Strip the content-based match filters (from, to, subject, has_attachment)
from hooks so matching is driven solely by event + the trusted gate.
Rename the hook `id` field to `name` everywhere (config, wire protocol,
env var, structured log field, CLI, Rust API). Make `name` optional:
when omitted, the daemon derives a stable 12-char hex name from
sha256(event + cmd + dangerously_support_untrusted) at use — no
config.toml mutation on load.

- Hook gains #[serde(deny_unknown_fields)]; stale filter fields and
  typos fail loudly at config load.
- derive_hook_name() uses sha2 over the three inputs joined by 0x1F;
  the first 12 hex chars pass is_valid_hook_name. Mailbox is
  deliberately excluded from the hash so collisions force an explicit
  name =.
- validate_hooks enforces global uniqueness across mailboxes and emits
  distinct errors for explicit-vs-explicit, anonymous-vs-anonymous,
  and mixed collisions.
- Wire protocol: Hook-Id: header renamed to Hook-Name:;
  HookDeleteRequest::id renamed to name.
- CLI: `hooks create` gains --name, drops --from / --to / --subject /
  --has-attachment. Anonymous hooks print their derived name on success
  without writing name = to config.
- Docs updated across book/, README.md, docs/, and project CLAUDE.md;
  book/configuration.md full config example uses an anonymous hook with
  the canonical explanatory TOML comment.
~~~

PR #111, opened 2026-04-20
[Sprint 1] Foundation: schema, socket rename, service user

uzyn:
## Sprint Goal
Land low-risk mechanical changes (socket rename, `[[hook_template]]` schema, `aimx-hook` service user) so every downstream hook-templates sprint builds on a clean base.

Tracks [`docs/hook-templates-sprint.md`](docs/hook-templates-sprint.md) Sprint 1 (Days 1-2.5).

### [S1-1] Rename runtime socket `send.sock` -> `aimx.sock`
- [x] `SEND_SOCKET_NAME` / `send_socket_path()` renamed to `AIMX_SOCKET_NAME` / `aimx_socket_path()` in `src/serve.rs`
- [x] All client call sites (`hook_client.rs`, `mcp.rs`, `send.rs`, `serve.rs`) resolve via the shared helper; no `send.sock` literal outside the intentional rename comment in `serve.rs`
- [x] Test fixtures in `src/serve.rs` + `tests/integration.rs` updated
- [x] Systemd / OpenRC `RuntimeDirectory=aimx` unchanged (only the socket filename moves)
- [x] `aimx doctor` surfaces a `stale send.sock detected` warning when the old name is present without the new one

### [S1-2] Add `[[hook_template]]` schema with load-time validation
- [x] New `HookTemplate` struct in `src/config.rs` with `name`, `description`, `cmd`, `params`, `stdin`, `run_as`, `timeout_secs`, `allowed_events`; defaults match PRD section 6.1 (`run_as = "aimx-hook"`, `timeout_secs = 60`, `allowed_events = [on_receive, after_send]`, `stdin = "email"`)
- [x] `Config` gains `#[serde(default, rename = "hook_template")] hook_templates: Vec<HookTemplate>`
- [x] `validate_hook_templates` called from `Config::load`; rejects duplicate names, empty cmd, placeholder in `cmd[0]`, undeclared placeholders, unused params, `timeout_secs` out of `[1, 600]`, `run_as` outside `{aimx-hook, root}`, invalid template names, empty `allowed_events`
- [x] Unit tests (`config::tests`) cover every rejection path plus a golden round-trip test
- [x] Fixture `tests/fixtures/hook_templates_valid.toml` demonstrates two-template load -> save -> load with no diff

### [S1-3] Extend `Hook` struct with `origin` + `template` + `params`
- [x] New `HookOrigin` enum (`Operator` default, `Mcp`) serialized lowercase
- [x] `Hook` gains `origin`, `template: Option<String>`, `params: BTreeMap<String, String>`; all default when absent so legacy hooks load as operator-origin raw-cmd
- [x] Mutual exclusion enforced in `validate_hooks`: `template = Some` requires empty `cmd`; `template = None` requires non-empty `cmd` and empty `params`
- [x] Unknown template refs fail `Config::load` with an actionable error pointing at `[[hook_template]]`
- [x] `Hook::is_template_bound()` convenience method; `derive_template_hook_name(event, template, sorted params)` produces stable 12-hex-char derived names for template-bound hooks
- [x] MCP-origin template hooks cannot set `dangerously_support_untrusted` (PRD section 6.8)
- [x] Unit tests cover round trip, mutual exclusion, missing template ref, legacy-default behavior, and the new derived-name function

### [S1-4] Create `aimx-hook` system user in `aimx setup`
- [x] `SystemOps` gains `user_exists`, `create_system_user`, `chown_group_readable` with `useradd` + busybox `adduser` fallback
- [x] `ensure_hook_user(sys, data_dir)` creates the user idempotently, then chowns `<data_dir>/inbox` and `<data_dir>/sent` to `root:aimx-hook` with `g+rX` recursively
- [x] `run_setup` calls `ensure_hook_user` after `finalize_setup` and before service install
- [x] `MockSystemOps` records user-creation and chown calls so unit tests assert the ordering + target paths
- [x] `[User]` section header matches the existing `[DNS]` / `[MCP]` style and reports `created` vs `already present`

## Technical Decisions
- **Separate derive path for template hooks.** `derive_hook_name(event, cmd, dangerous)` stays unchanged for raw-cmd hooks; template-bound hooks use `derive_template_hook_name(event, template, sorted params)`. Mixing them in one function would mean a stable raw-cmd hash depends on two empty fields when a template is set — and vice versa — which is fragile.
- **`cmd` stays `String` (not `Option<String>`) on the `Hook` struct.** Empty string is the "no cmd" sentinel, skip-serialized so round-trips remain clean. Switching to `Option<String>` would ripple through 60+ call sites for zero correctness benefit.
- **TOML key is `[[hook_template]]` (singular)** via `#[serde(rename = "hook_template")]` on the `hook_templates` field. Matches PRD wording and TOML's array-of-tables convention.
- **Datadir README version bumped from 5 to 6** because the template now references `aimx.sock` instead of `send.sock`. Running daemons refresh the README on their next restart via the existing `refresh_if_outdated` path.
- **`aimx doctor` stale-socket warning** added now (per the S1-1 AC). Lightweight: surfaces a warn line in the Service section if the runtime dir still has `send.sock` but not `aimx.sock`. Non-blocking; operators just need to restart the daemon.

## Deferred Items
- **Book chapters (`book/*.md`)** - deferred to Sprint 6 per `docs/hook-templates-sprint.md` section S6-3. One `book/faq.md`, `book/cli.md`, `book/configuration.md`, etc. sentence still references `send.sock`; sprint plan explicitly covers those rewrites together.
- **`HookOrigin::as_str()` has `#[allow(dead_code)]`** - Sprint 2 wires it into the structured hook-fire log line (`run_as`, `template`, `origin`). Keeping the helper available now avoids a noisy diff when Sprint 2 lands.
- **Sandbox execution** - Sprint 2 replaces `Command::new("sh").arg("-c")` with `spawn_sandboxed` (systemd-run or setuid fallback). This sprint only lays the foundation (user exists, datadir chown'd).
- **MCP tool surface** - Sprint 5 adds `hook_list_templates`, `hook_create`, `hook_list`, `hook_delete`. The schema this sprint ships is the contract those tools will speak.

## Review Focus Areas
- **`src/config.rs` `validate_hook_templates` + `iter_placeholders`** - the placeholder extractor is a hand-rolled parser. Worth scrutinising the edge cases (unclosed braces, non-lowercase, empty name `{}`). See `iter_placeholders_*` tests.
- **`src/config.rs` `validate_hooks`** - mutual-exclusion branches now cross-check `template` -> templates map, `cmd` emptiness, `params` emptiness, and the MCP-origin `dangerously_support_untrusted` rule. A good place to double-check for ordering bugs.
- **`src/setup.rs` `ensure_hook_user` + `chown_datadir_for_hook_user`** - filesystem side effects on a root-only path. Only exercised in unit tests via `MockSystemOps`; manual `sudo aimx setup` verification will happen as part of the Sprint 4 integration test story.
- **`src/hook.rs` `derive_template_hook_name`** - determinism relies on `BTreeMap` iteration order. Covered by a dedicated test; worth eyeballing the hash format vs. the existing raw-cmd `derive_hook_name`.

## Test Plan
- [x] `cargo test --bin aimx` - 885 unit tests pass (+33 from the Sprint 1 baseline of 852).
- [x] `cargo test --test integration` - 84 integration tests pass. One pre-existing test (`serve_e2e_stale_readme_refreshed_at_startup`) was updated to expect README version 6.
- [x] `cargo clippy --all-targets -- -D warnings` - clean.
- [x] `cargo fmt -- --check` - clean.
- [ ] Manual `sudo aimx setup` run on a clean VM to confirm the `[User]` section prints and `id aimx-hook` resolves afterwards (deferred; setup is a privileged-only flow).


uzyn (commented):
## Sprint Goal Assessment

Sprint 1 lands exactly what it promised: the `send.sock` → `aimx.sock` rename propagates cleanly, the `[[hook_template]]` schema + load-time validator go in with a thorough rejection-path test suite, the `Hook` struct gains `origin` / `template` / `params` with mutual-exclusion enforcement, and `aimx setup` now creates the `aimx-hook` system user and chowns the mailbox dirs group-readable. 885 unit tests + 84 integration tests pass locally; `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean. The foundation is solid for Sprints 2-6 to build on.

**[S1-1] Rename runtime socket `send.sock` → `aimx.sock`**
- [x] `SEND_SOCKET_NAME` → `AIMX_SOCKET_NAME`, `send_socket_path()` → `aimx_socket_path()` in `src/serve.rs` — met
- [x] All client call sites (`hook_client.rs`, `mcp.rs`, `send.rs`) resolve via shared helper — met; remaining `send.sock` literals are the intentional historical-note comment in `serve.rs` and the stale-socket detection in `doctor.rs`
- [x] Test fixtures + `tests/integration.rs` updated — met
- [x] `RuntimeDirectory=aimx` unchanged — met (only filename moves)
- [x] `aimx doctor` stale-socket warning — met, covered by `format_status_flags_stale_send_sock` test
- [x] Clean `cargo test` / `clippy` / `fmt` — met

**[S1-2] `[[hook_template]]` schema with load-time validation**
- [x] `HookTemplate` struct with PRD §6.1 fields + `#[serde(deny_unknown_fields)]` — met
- [x] `Config.hook_templates` via `#[serde(default, rename = "hook_template", skip_serializing_if = "Vec::is_empty")]` — met
- [x] `validate_hook_templates` runs from `Config::load` before `validate_hooks` — met
- [x] Every rejection path covered: duplicate name, empty cmd, placeholder in `cmd[0]`, undeclared placeholder, unused param, timeout bounds, bad `run_as`, invalid name, empty `allowed_events` — met
- [x] Fixture `tests/fixtures/hook_templates_valid.toml` exercises load → save → load round-trip — met

**[S1-3] Extend `Hook` with `origin` + `template` + `params`**
- [x] `HookOrigin` enum, `#[serde(rename_all = "lowercase")]`, `Operator` as `#[default]` — met
- [x] `Hook` gains `origin`, `template`, `params` with `#[serde(default)]` so legacy hooks load unchanged — met, covered by `legacy_hook_toml_defaults_to_operator_origin`
- [x] Mutual exclusion in `validate_hooks` + `validate_single_hook` — met
- [x] Unknown template ref fails load — met
- [x] `Hook::is_template_bound()` — met
- [ ] `Hook::resolve_argv(&self, &[HookTemplate], &Builtins) -> Result<Vec<String>>` — **NOT MET**: the sprint plan AC explicitly names this method, but it's absent from the PR. The PR description replaces it with `derive_template_hook_name`, which is a different thing (name derivation vs. argv resolution). Placeholder substitution is deferred to Sprint 2's S2-1 (`src/hook_substitute.rs::substitute_argv`), which is the right architectural call — but the S1-3 AC as written is unmet. See Non-blocker #1 below; easiest fix is to either add a thin `Hook::resolve_argv` stub that defers to S2-1 or amend the sprint plan to move this AC into Sprint 2.
- [x] Unit tests cover round trip, mutual exclusion, missing template ref, legacy defaults, derived name — met
- [x] MCP-origin template hooks cannot set `dangerously_support_untrusted` — met for template-bound case. See Non-blocker #2 for the raw-cmd gap.

**[S1-4] Create `aimx-hook` system user in `aimx setup`**
- [x] `SystemOps::{user_exists, create_system_user, chown_group_readable}` with `useradd` + `adduser` fallback — met
- [x] `ensure_hook_user(sys, data_dir)` idempotent + chowns `inbox/` and `sent/` `root:aimx-hook` with `g+rX` recursive — met
- [x] `run_setup` calls `ensure_hook_user` after `finalize_setup` and before `install_service_file` — met (line 1707, service install at 1811)
- [x] `MockSystemOps` records calls + tests assert order and target paths — met
- [x] `[User]` section matches `[DNS]` / `[MCP]` style, reports created vs. already-present — met
- [x] Clean `cargo test` / `clippy` / `fmt` — met

## Test Coverage

Test coverage on the new surface is thorough. Notable positives:

- Every named rejection path in `validate_hook_templates` has a dedicated unit test with a minimal failing TOML input.
- `hook_templates_fixture_round_trips` confirms save→load idempotence (important for the hot-reload sprints downstream).
- `derive_template_hook_name_is_deterministic` + `derive_template_hook_name_differs_by_params` + `effective_hook_name_prefers_template_hash_over_cmd_hash` pin the hash-invariance contract.
- `MockSystemOps` correctly simulates user-creation and chown ordering; both fresh-install and already-present paths are exercised.

Gaps worth noting:

- **`iter_placeholders` edge cases are under-tested.** The parser is a hand-rolled scanner (reasonable given the small regex). Three tests exist (simple names, unclosed braces, non-alphanum). Missing coverage: empty `{}`, adjacent placeholders `{a}{b}`, placeholders at end-of-string `"hello {x}"`, nested braces `{{name}}`, very long names. Sprint 2's fuzz story (S2-2 / S6-2) will catch most of these, but a few targeted cases now would be cheap insurance on a function that's load-bearing for the injection-resistance guarantee.
- **No test asserts that `origin = "mcp"` + raw-cmd (no template) + `dangerously_support_untrusted = true` is rejected.** See Non-blocker #2.

## Potential Bugs

**Non-blocker #1: `Hook::resolve_argv` missing — S1-3 acceptance criterion unmet**  
`src/hook.rs`. The sprint plan's S1-3 acceptance criteria explicitly name `Hook::resolve_argv(&self, &[HookTemplate], &Builtins) -> Result<Vec<String>>`. The PR adds `is_template_bound()` and `derive_template_hook_name` but not `resolve_argv`. The PR description's S1-3 AC list silently drops it. This isn't a correctness bug per se — no current call site needs the method — but it means a reviewer checking the sprint plan literally will find one AC unmet. Recommended fix: either (a) add a thin `resolve_argv` that errors with `"not yet implemented — Sprint 2 wires this"` so the API surface exists, or (b) amend the sprint plan to move the method into Sprint 2 where the real substitution lives. Deciding this now avoids confusion when Sprint 2 lands.

**Non-blocker #2: `dangerously_support_untrusted` + MCP-origin guard only fires for template-bound hooks**  
`src/config.rs`, `validate_hooks`, lines ~354-365. The check reads:

```rust
if hook.is_template_bound()
    && hook.origin == HookOrigin::Mcp
    && hook.dangerously_support_untrusted
```

PRD §6.8 says MCP-origin hooks must always fire only on trusted mail; that invariant should hold for all MCP-origin hooks, not just template-bound ones. Today the UDS `HOOK-CREATE` path accepts any `Hook` TOML (Sprint 3's S3-1 tightens the body schema), so a crafted request could land an `origin = "mcp"`, raw-cmd, `dangerously_support_untrusted = true` hook on disk. Sprint 3 will close the client-side path, but the validator itself should reject MCP-origin + `dangerously_support_untrusted = true` regardless of whether the hook is template-bound. Drop the `is_template_bound()` guard.

**Non-blocker #3: `create_system_user` falls back silently on any `useradd` failure**  
`src/setup.rs`, lines ~97-123. The `match status { Ok(s) if s.success() => Ok(true), _ => { <fall back to adduser> } }` pattern treats every non-success `useradd` exit (syntax error, permission denied, UID pool exhausted, `useradd` missing) as a signal to try `adduser`. On Debian/Ubuntu where both are on PATH, the `adduser` fallback will usually fail with confusing diagnostics because Debian's `adduser` is a different program with non-compatible flags. This isn't a correctness bug — setup still fails loudly eventually — but the error propagated to the operator is suboptimal. Consider distinguishing "command not found" (errno `ENOENT`) from "command ran and failed" and only falling back on the former.

**Non-blocker #4: Placeholder parser accepts `_` but rejects uppercase, with misleading error paths**  
`src/config.rs`, `iter_placeholders` + `validate_hook_templates`. Placeholder syntax accepts `[a-z0-9_]+`; `params` entries are never validated against any charset. If an operator writes `params = ["FOO"]` and `cmd = ["/bin/echo", "{FOO}"]`, the placeholder extractor doesn't match `{FOO}`, so `FOO` is reported as "declares unused param" — misleading. Two fixes possible: validate `params` entries against a regex at load time, or make `iter_placeholders` match a wider charset and reject disallowed names with a precise error. Not urgent because the operator does get an error; the message is just confusing.

## Security Issues

No new security risks introduced. The security story is:

1. Socket rename is mechanical; permissions (`0o666`) unchanged — same authorization boundary as today.
2. `aimx-hook` user creation + chown is conservative (`root:aimx-hook`, `g+rX`, recursive). The DKIM private key at `/etc/aimx/dkim/private.key` remains `0600` root-only — the chown is only on `<data_dir>/inbox` and `<data_dir>/sent`, never on the config dir.
3. Template schema validation rejects placeholders in `cmd[0]`, which is the injection-resistance load-bearing check. Sprint 2's `substitute_argv` will do the actual runtime enforcement; Sprint 1 correctly lays the ground rule ("binary path is always a literal") at schema-parse time.

One thing to keep in mind for Sprint 2: the stale-`send.sock`-detected warning in `doctor` is informational and doesn't prevent operation, which is correct (a mid-restart state where both sockets exist briefly should not trip it, and the check guards against that with `&& !aimx.sock.exists()`).

## Code Quality

- **Good**: `HookTemplate`, `HookOrigin`, and the new `Hook` fields are documented with purpose + rejection rules + the cross-cutting PRD section references. Makes the downstream sprints much easier to read.
- **Good**: `#[serde(deny_unknown_fields)]` on `HookTemplate` (and preserved on `Hook`) means typos in `config.toml` fail loudly rather than silently dropping fields.
- **Good**: `derive_template_hook_name` uses the same `DERIVED_HOOK_NAME_LEN` constant and `0x1F` separator convention as `derive_hook_name`. Consistent.
- **Minor**: `HookOrigin::as_str()` carries `#[allow(dead_code)]` — flagged as intentional in the PR description (Sprint 2 wires it into the log line). Acceptable for now but should be removed in Sprint 2.
- **Minor**: The `validate_hook_templates` function builds a `HashSet<&str>` of `HOOK_TEMPLATE_BUILTIN_PLACEHOLDERS` inline each iteration via `.contains(&placeholder)` on a `&'static [&str]` (linear scan, 5 entries — fine in practice). No action needed; just noting.

## Alignment with PRD

Implementation tracks PRD §§6.1, 6.5, 6.8, and 8.2 faithfully for the schema, socket rename, service user, and the `origin` tag. One minor divergence from PRD §6.8 flagged as Non-blocker #2 above.

No P0 PRD requirement that this sprint was supposed to address is missing. Deferrals (sandbox execution, MCP surface, book docs) are all explicitly scoped to later sprints in the sprint plan and the PR description.

**Overall verdict:** Needs minor fixes (one unmet AC to reconcile + three non-blocker improvements that would sharpen the validator). The code itself is solid — everything builds, all tests pass, and the foundation is clean — but the sprint plan explicitly promised `Hook::resolve_argv`, and we should either add it or formally move it to Sprint 2 before marking this done.

**Blockers:** None.

**Non-blockers:**
1. **S1-3 AC gap**: `Hook::resolve_argv` is listed in the sprint plan's S1-3 acceptance criteria but not implemented. Add a stub or amend the sprint plan.
2. **Validator asymmetry**: MCP-origin `dangerously_support_untrusted = true` is only rejected for template-bound hooks. Drop the `is_template_bound()` guard in `validate_hooks` so the check applies to raw-cmd MCP-origin hooks too (belt-and-braces; the UDS body-schema tightening lands in Sprint 3).
3. **`create_system_user` fallback classification**: Silently falling back to `adduser` on any `useradd` failure produces confusing errors on Debian where `adduser` has different flag semantics. Only fall back when `useradd` is `ENOENT`-not-found.
4. **Placeholder parser vs. `params` charset**: If an operator declares a param containing `-` or uppercase, the extractor silently skips it and the error surfaces as "unused param" rather than "invalid param name". Either validate `params` entries at load time or widen the extractor + reject.

**Nice-to-haves:**
- A few more targeted `iter_placeholders` unit tests (empty `{}`, adjacent `{a}{b}`, nested `{{x}}`) before Sprint 2's fuzz story.
- When Sprint 2 lands, remove `#[allow(dead_code)]` on `HookOrigin::as_str()`.

uzyn:
## Review fixes (commit 5d82dc6)

Addressed all four non-blockers from the review. Summary:

### Fixed

1. **S1-3 `Hook::resolve_argv` AC gap** — formally re-scoped to Sprint 2's S2-1 by amending `docs/hook-templates-sprint.md`. The method delegates to `substitute_argv` in the new `src/hook_substitute.rs` module, which only comes into existence in S2-1; adding a stub now would just be a forward reference to that module. The S1-3 bullet now carries a back-reference so the history is visible to future readers.

2. **MCP-origin `dangerously_support_untrusted` guard too narrow** — dropped the `is_template_bound()` qualifier in `validate_hooks`. Added the same guard to `validate_single_hook` so the UDS `HOOK-CREATE` path (Sprint 3) also rejects the shape. Per PRD §6.8 this invariant applies to every MCP-origin hook regardless of template binding. Two new tests cover it (`load_rejects_mcp_origin_raw_cmd_with_dangerously_flag`, `validate_single_hook_rejects_mcp_origin_raw_cmd_with_dangerously_flag`).

3. **`create_system_user` silent fallback** — now only falls back from `useradd` to `adduser` on `io::ErrorKind::NotFound`. Any other failure (useradd ran and exited non-zero, permission denied spawning it) is propagated as-is with the real error message. Debian boxes where `useradd` exists but fails for reasons unrelated to "not on PATH" now surface the real cause.

4. **Placeholder charset vs. params charset mismatch** — picked option (a) from the review: `validate_hook_templates` now validates every `params` entry against `[a-z0-9_]+` at load time, and also rejects params that collide with the built-in placeholder names (`event`, `mailbox`, `message_id`, `from`, `subject`). Operators writing `params = [\"FOO\"]` now get a precise `invalid param name 'FOO'` error instead of the misleading "unused param" downstream. Three new tests cover uppercase / dashed / built-in-colliding cases.

### Intentionally left as-is

- **Nice-to-have: `iter_placeholders` edge-case tests** (empty `{}`, adjacent `{a}{b}`, nested `{{x}}`, end-of-string). The reviewer explicitly flagged these as \"fuzz story in Sprint 2's S2-2 / S6-2 will catch most of these\"; leaving them for that sprint where the fuzz infrastructure lands rather than hand-writing a few more cases now.
- **Nice-to-have: remove `#[allow(dead_code)]` on `HookOrigin::as_str`** — explicitly scoped to Sprint 2 by the reviewer (\"should be removed in Sprint 2\"), since Sprint 2 is what wires the field into the fire-log line. Same applies to the new `#[allow(dead_code)]` on `Hook::is_template_bound()`, which Sprint 2's `resolve_argv` will consume.

### Verification

- `cargo test` — 890 unit + 84 integration tests pass

uzyn (commented):
## Re-review of PR #111 (Sprint 1)

Reviewed fix commit `5d82dc6` against the four non-blockers flagged in the initial review. All four are resolved cleanly, no regressions introduced.

## Resolution of Previously Flagged Issues

### 1. `Hook::resolve_argv` missing from S1-3 — RESOLVED

- `docs/hook-templates-sprint.md` S1-3 bullet amended to explicitly note the method was moved to S2-1 with a back-reference.
- S2-1 bullet gains a `(moved from S1-3)` note citing the rationale (`hook_substitute.rs` doesn't exist until S2-1, so resolve_argv would be a forward reference).
- `Hook::is_template_bound()` marked `#[allow(dead_code)]` with a comment noting the Sprint 2 consumer.
- History is preserved and the re-scope is discoverable. Acceptable.

### 2. MCP-origin `dangerously_support_untrusted` guard — RESOLVED

- `validate_hooks` (src/config.rs:354) now reads `hook.origin == HookOrigin::Mcp && hook.dangerously_support_untrusted` — the `is_template_bound()` qualifier was correctly dropped so the guard covers every MCP-origin hook regardless of template binding. PRD §6.8 now enforced uniformly.
- Equivalent guard added to `validate_single_hook` (src/config.rs:441) so the UDS `HOOK-CREATE` path (Sprint 3) fails fast with the same rejection. Defense in depth is maintained: `hook_handler.rs` calls `validate_single_hook` first then re-runs `validate_hooks` on the merged config.
- Two new tests cover both paths: `load_rejects_mcp_origin_raw_cmd_with_dangerously_flag` and `validate_single_hook_rejects_mcp_origin_raw_cmd_with_dangerously_flag`. Both pass.

**Minor nit (non-blocker, noting here only for the implementer's awareness):** in `validate_hooks`, the on_receive event check fires first, so an MCP-origin raw-cmd hook with `event = after_send` + `dangerously = true` would surface the event-mismatch error rather than the MCP-origin error. Both are correct rejections, just different messages. Fine to leave.

### 3. `create_system_user` silent `adduser` fallback — RESOLVED

- `setup.rs:113-131` now explicitly matches:
  - `Ok(s) if s.success()` → user created (happy path)
  - `Ok(s)` → `useradd exited with {s}` propagated (new precise error)
  - `Err(e) if e.kind() == ErrorKind::NotFound` → fall back to BusyBox `adduser`
  - `Err(e)` → `failed to spawn useradd: {e}` propagated (new precise error)
- The fallback is now only triggered when `useradd` is genuinely missing from PATH. Non-zero exits and spawn failures for other reasons are surfaced verbatim. Matches the reviewer's requested fix.
- No test was added for the error paths, but the logic is simple and testing it would require mocking `Command::status()` — acceptable scope boundary for Sprint 1.

### 4. Params charset mismatch — RESOLVED

- New `is_valid_placeholder_name` helper (src/config.rs:610) using the exact charset `[a-z0-9_]+` that `iter_placeholders` accepts. Charsets are now in lockstep with a comment explicitly calling out the invariant.
- `validate_hook_templates` (src/config.rs:539-554) loops over `tmpl.params` and rejects:
  - Invalid charset with "hook template '...' declares invalid param name 'FOO': param names must match [a-z0-9_]+"
  - Collision with built-in placeholders (`event`, `mailbox`, `message_id`, `from`, `subject`) with a dedicated error message listing the full built-in set.
- The new param-charset validation runs **before** the unused-param check, so operators get the precise error instead of the misleading "unused param" downstream. Exactly the fix requested.
- Three new tests cover the fix:
  - `load_rejects_param_name_with_uppercase_letters` (FOO)
  - `load_rejects_param_name_with_dash` (foo-bar)
  - `load_rejects_param_name_colliding_with_builtin` (param named `event`)
- All three pass.

## New Issues Found

None. The fix commit is tightly scoped, all new code paths are covered by tests, and no code outside the review scope was touched.

## Verification

- `cargo test --bin aimx` → 890 passed (+5 from baseline 885, matching the three charset tests + two MCP-origin tests)
- `cargo test --test integration` → 84 passed
- `cargo clippy --all-targets -- -D warnings` → clean
- `cargo fmt -- --check` → clean

## Still Open (intentionally deferred)

The implementer's PR comment correctly calls these out and they match the original reviewer's guidance:

- `iter_placeholders` edge-case tests (empty `{}`, adjacent `{a}{b}`, nested `{{x}}`) → covered by S2-2 / S6-2 fuzz work
- `#[allow(dead_code)]` on `HookOrigin::as_str` and `Hook::is_template_bound` → removed in Sprint 2 when they're wired into the fire-log line and `resolve_argv`

Both are nice-to-haves, already tracked in the sprint plan, not worth blocking a re-review over.

## Verdict

**Ready to merge.** All four flagged non-blockers are resolved, the fixes are narrowly scoped, test coverage is solid, and the deferrals are appropriate.

```
[Sprint 1] Foundation: schema, socket rename, service user

- Rename runtime socket `send.sock` → `aimx.sock`; doctor surfaces a stale-socket warning
- Add `[[hook_template]]` schema with load-time validation (duplicates, empty cmd, placeholder in cmd[0], undeclared/unused params, timeout / run_as / allowed_events, param charset, built-in collisions)
- Extend `Hook` with `origin` (operator|mcp), `template`, `params`; mutual exclusion enforced; MCP-origin hooks cannot set `dangerously_support_untrusted` (both config load + UDS validate paths)
- Create `aimx-hook` system user in `aimx setup` with useradd / BusyBox adduser fallback; chown mailbox dirs group-readable
- Derived name function `derive_template_hook_name(event, template, sorted params)` for template-bound hooks

Bumps datadir README to v6. 890 unit + 84 integration tests passing; clippy + fmt clean.
```

PR #112, opened 2026-04-20
[Sprint 2] Sandboxed hook executor

uzyn:
## Sprint Goal

Replace today's `sh -c` executor with an argv-based, privilege-dropping, timeout-enforced sandboxed runner. This is the highest-risk sprint in the hook-templates track — every downstream sprint depends on it.

Sprint plan: [`docs/hook-templates-sprint.md`](docs/hook-templates-sprint.md) — Sprint 2.

### [S2-1] Placeholder substitution with argv-safety guarantees
- [x] New `src/hook_substitute.rs` module with `substitute_argv`, `BuiltinContext`, `SubstitutionError`
- [x] `Hook::resolve_argv(&self, templates, builtins)` delegates to `substitute_argv` for template-bound hooks and returns `["/bin/sh","-c",cmd]` for raw-cmd hooks
- [x] `SubstitutionError` variants: `UnknownPlaceholder`, `ParamContainsWhitespace` (via control-char rejection), `ParamContainsNul`, `ParamContainsControl`, `ParamTooLong` (>8 KiB), `PlaceholderInBinaryPath`, `EmptyTemplate`
- [x] Substitution never expands argv count; placeholder filling the whole slot still yields exactly one entry
- [x] Unit tests cover every rejection path + unicode values + shell-metacharacter payloads
- [x] Inline fuzz-style loop runs 15,624 iterations of metacharacter permutations and asserts `substituted.len() == template.len()` (exceeds the 10K target)

### [S2-2] `spawn_sandboxed` platform helper (systemd-run + setuid fallback)
- [x] `src/platform.rs::spawn_sandboxed` picks path via `detect_init_system()`
- [x] systemd path shells out to `systemd-run --quiet --pipe --wait --collect --uid --gid --property=ProtectSystem=strict --property=PrivateDevices=yes --property=NoNewPrivileges=yes --property=MemoryMax=256M --property=RuntimeMaxSec=<timeout>`; env vars forwarded via `--setenv=KEY=VAL`
- [x] Fallback path uses `Command` + `pre_exec { setsid; setgroups(0); setgid; setuid }`, then `execvp`; timeout enforced via parent-side poll + SIGTERM + SIGKILL at `timeout + SIGKILL_GRACE` (5s)
- [x] Signals go to the whole process group (`-pid`) so grandchildren (e.g. a shell's `sleep`) are killed too — prevents stdout-drain-thread wedging
- [x] `SandboxStdin::{Email, EmailJson, None}` written on a thread, stdin closed before timeout clock starts
- [x] stdout/stderr drained on threads and truncated at `MAX_OUTPUT_BYTES` (64 KiB)
- [x] `SandboxError::{UserNotFound, SpawnFailed, IoFailed}` variants with clear Display
- [x] Unknown-user fallback for non-root callers runs as current UID with a warning (keeps dev + CI usable before `aimx setup` creates `aimx-hook`)
- [x] Unit tests cover exit-code propagation, stdout+stderr capture, 64 KiB tail truncation, SIGTERM timeout (200 ms), SIGKILL-after-grace when SIGTERM ignored, stdin piping, env-var propagation, unknown-user lookup

### [S2-3] Structured hook-fire log with new fields
- [x] Log line now carries `template=<name|->`, `run_as=<user>`, `sandbox=<systemd-run|setuid|->`, `timed_out=<bool>`, `stderr_tail=<json-escaped>`
- [x] `stderr_tail` is JSON-escaped so multi-line output doesn't break structured parsing; empty stderr renders as `""`, not `null`
- [x] Long stderr rendered as `"<head>...<tail>"` at a stable UTF-8 char boundary, capped at ~1 KiB inline
- [x] Unit tests verify empty / escape / truncation / template=name / template=`-` cases

### [S2-4] Wire sandboxed executor into on_receive + after_send
- [x] `run_and_log` refactored: resolves argv via `Hook::resolve_argv`, builds `BuiltinContext` from the event, prepares stdin per `template.stdin` (or `Email` default for raw-cmd), calls `spawn_sandboxed` with `hook.run_as` (defaults to `"aimx-hook"`)
- [x] Raw-cmd hooks gain an optional `run_as: Option<String>` on the `Hook` struct — `None` → template/default `aimx-hook`, `Some("root")` → no privilege drop (operator-only; not settable over UDS per PRD §6.7)
- [x] `execute_on_receive` / `execute_after_send` now take `&Config` so they can resolve `hook_templates` at fire time; call sites in `src/ingest.rs` and `src/send_handler.rs` updated
- [x] Env vars propagated through both sandbox paths (`AIMX_*` including new `AIMX_MESSAGE_ID`, `AIMX_EVENT`, `AIMX_ID`, `AIMX_DATE`); on the systemd path they go through `--setenv=K=V`
- [x] Dead code removed: old `Command::new("sh").arg("-c")` path gone; `substitute_template` helper and its unit test removed; `SandboxKind` / `SandboxOutcome` / `SandboxError` wired into the log line and caller

## Technical Decisions

- **Process-group signalling.** The fallback path calls `setsid` in `pre_exec` so the child is its own session leader; timeout signals go to `-pid` (the pgid). This is the fix for a real bug found while writing the SIGKILL test: a shell that `exec`s `sleep 30` leaves the sleep reparented to init, which keeps stdout pipes open and wedges the drain threads after the shell exits.
- **No new dependency.** Sprint plan suggested `nix::unistd`; stdlib `Command::pre_exec` + `libc::{setsid, setgroups, setgid, setuid}` covers it without growing the dep graph.
- **`AIMX_SANDBOX_FORCE_FALLBACK=1` test hook.** `systemd-run` on a systemd host requires PolicyKit auth for non-root users, so every hook-firing test would fail on a dev / CI systemd box. The env-var override lets integration tests (and `tests::` modules in `src/`) force the setuid fallback, which exercises the same observable surface (exit codes, stderr capture, env var propagation, timeouts).
- **Unknown-user fallback on non-root callers.** `spawn_sandboxed("aimx-hook", …)` on a CI host without that user would blow up. Instead, when `run_as` doesn't resolve AND the caller isn't root, we log a warning and run as the current UID. Production is unaffected — `aimx setup` creates `aimx-hook` before the daemon starts. Test coverage: `non_root_unknown_user_falls_back_to_current_uid`.
- **`Hook.run_as: Option<String>`.** Per PRD §6.7, raw-cmd hooks must be able to pin `run_as = "root"` via `config.toml`. The field is deliberately `Option<String>` so the default (`aimx-hook` or the template's value) is invisible in serialized hooks. The UDS path will reject this field in Sprint 3's S3-1.
- **stdin=`Email` default for raw-cmd.** Sprint plan calls for it; I kept ingest/after_send tests working by having `build_stdin` tolerate a missing `.md` file (empty payload instead of error).

## Deferred Items

None. All four stories' ACs are met. The `systemd-run` tests ran only via the fallback path on this CI (the systemd path is exercised in production); PRD §7.2 already calls this out as the stance.

## Review Focus Areas

- **`src/platform.rs`** — the sandbox core. The signal-pgroup behavior + stdout/stderr drain threads + stdin thread are the load-bearing bits; they took several iterations to get right under the tests. Particular attention to `run_child_with_timeout` (lines that spawn the drain threads and the poll loop) and the early-return user-not-found path (stdio pipes have to be configured BEFORE the early return).
- **`src/hook.rs` `run_and_log`** — replaced end to end. `build_stdin` is best-effort on missing files per PRD §9 out-of-scope; `format_stderr_tail` is JSON-escaped and truncated at a char boundary.
- **`src/hook_substitute.rs`** — the argv-injection defense. The substitution never grows argv count; the fuzz-like loop is inline and runs in the default `cargo test` profile (not just `--release`).
- **Integration test env threading.** `aimx_cmd`, `aimx_cmd_isolated`, `start_serve`, `start_serve_with_mail_drop`, `start_serve_with_env` all now set `AIMX_SANDBOX_FORCE_FALLBACK=1`. A prod daemon does not set this.
- **`Hook::run_as` schema.** Option field added with `#[serde(default, skip_serializing_if = "Option::is_none")]` — legacy hooks missing the field load as `None`, and defaults aren't written to disk.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2258,6 +2258,7 @@ dangerously_support_untrusted = true
             origin: HookOrigin::Mcp,
             template: None,
             params: std::collections::BTreeMap::new(),
+            run_as: None,
         };
         let err = validate_single_hook(&hook).unwrap_err();
         assert!(
--- a/src/doctor.rs
+++ b/src/doctor.rs
@@ -1543,6 +1543,7 @@ mod tests {
                     origin: crate::hook::HookOrigin::Operator,
                     template: None,
                     params: std::collections::BTreeMap::new(),
+                    run_as: None,
                 }],
                 trust: Some("verified".to_string()),
                 trusted_senders: Some(vec!["alice@example.com".to_string()]),
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -79,10 +79,9 @@ pub enum HookOrigin {
 }
 
 impl HookOrigin {
-    /// String form used in the structured log line and any future
-    /// human-facing doctor / diagnostic output. Unused in Sprint 1 (no
-    /// call site emits the field yet); Sprint 2 wires it into the hook
-    /// fire log.
+    /// String form used by `doctor` output and tests. Production log
+    /// lines do not currently surface origin (see PRD §7.3); kept as an
+    /// API so Sprint 3's `hooks list` / doctor summary can format it.
     #[allow(dead_code)]
     pub fn as_str(&self) -> &'static str {
         match self {
@@ -165,9 +164,9 @@ pub struct Hook {
 
 impl Hook {
     /// True iff this hook references a template (rather than carrying its
-    /// own raw `cmd`). Consumed by [`Hook::resolve_argv`] and by the
-    /// `doctor` summary in a later sprint.
-    #[allow(dead_code)] // Wired into run_and_log in S2-4; kept for doctor in S6.
+    /// own raw `cmd`). Consumed by tests today; wired into `doctor` and
+    /// `hooks list` summary in Sprint 6.
+    #[allow(dead_code)]
     pub fn is_template_bound(&self) -> bool {
         self.template.is_some()
     }
@@ -184,7 +183,6 @@ impl Hook {
     /// template hooks at the [`crate::platform::spawn_sandboxed`] call
     /// site (both produce a `Vec<String>` argv); shell interpretation is
     /// intentional for operator-authored hooks.
-    #[allow(dead_code)] // Wired into run_and_log in S2-4.
     pub fn resolve_argv(
         &self,
         templates: &[HookTemplate],
@@ -218,7 +216,6 @@ impl Hook {
 /// All variants indicate a configuration-level bug that validation should
 /// normally catch at load time. They are kept as distinct variants so the
 /// caller can emit a precise `tracing::warn!` without swallowing context.
-#[allow(dead_code)] // Consumed in S2-4 via run_and_log.
 #[derive(Debug)]
 pub enum ResolveArgvError {
     UnknownTemplate(String),
@@ -665,26 +662,45 @@ fn run_and_log(
 /// Build the stdin payload for a hook fire. For `Email` / `EmailJson`
 /// modes we load the associated `.md` file; if the file doesn't exist
 /// (e.g. TEMP failures on `after_send` where persistence was skipped)
-/// we pipe an empty payload rather than failing the hook.
+/// we pipe an empty payload rather than failing the hook. A real
+/// read error (EACCES etc.) is logged at WARN so operators can tell a
+/// missing-file (intentional empty payload) apart from a permissions
+/// bug (which would otherwise silently present as empty stdin).
 fn build_stdin(mode: HookTemplateStdin, source: Option<&Path>) -> SandboxStdin {
     match mode {
         HookTemplateStdin::None => SandboxStdin::None,
-        HookTemplateStdin::Email => {
-            let bytes = source
-                .and_then(|p| std::fs::read(p).ok())
-                .unwrap_or_default();
-            SandboxStdin::Email(bytes)
-        }
+        HookTemplateStdin::Email => SandboxStdin::Email(read_stdin_source(source)),
         HookTemplateStdin::EmailJson => {
             // Best-effort per PRD §9 out-of-scope: wrap raw `.md` bytes
             // in a JSON object keyed by `raw`. Stabilization is post-v1.
-            let bytes = source
-                .and_then(|p| std::fs::read(p).ok())
-                .unwrap_or_default();
+            let bytes = read_stdin_source(source);
             let escaped = serde_json::to_string(&String::from_utf8_lossy(&bytes).into_owned())
                 .unwrap_or_else(|_| "\"\"".into());
             let json = format!("{{\"raw\":{escaped}}}");
             SandboxStdin::EmailJson(json.into_bytes())
+        }
+    }
+}
+
+/// Read the backing `.md` file for a hook's stdin payload. `None` path
+/// and `NotFound` both yield empty bytes silently (the TEMP-failure
+/// `after_send` case where no file was ever persisted — see PRD §9).
+/// Any other `io::Error` is surfaced as a WARN so an EACCES on a real
+/// file doesn't hide behind the empty-payload code path.
+fn read_stdin_source(source: Option<&Path>) -> Vec<u8> {
+    let Some(p) = source else {
+        return Vec::new();
+    };
+    match std::fs::read(p) {
+        Ok(bytes) => bytes,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Vec::new(),
+        Err(e) => {
+            tracing::warn!(
+                target: "aimx::hook",
+                "hook stdin: read {path} failed: {e}; piping empty payload",
+                path = p.display(),
+            );
+            Vec::new()
         }
     }
 }
@@ -1621,6 +1637,41 @@ cmd = "echo legacy"
         assert!(rendered.contains("..."), "long stderr must be truncated");
         assert!(rendered.starts_with('"'));
         assert!(rendered.ends_with('"'));
+    }
+
+    /// Regression: the tail-start index inside `format_stderr_tail` is
+    /// a byte offset computed from `tail_n`. With ASCII-only input it
+    /// always lands on a char boundary, but multi-byte UTF-8 scalars
+    /// can straddle the boundary. The implementation walks forward to
+    /// the next `is_char_boundary`; this test locks in that the walk
+    /// doesn't panic on either a 3-byte-BMP or a 4-byte-astral char
+    /// straddling the truncation point, AND that the final output is
+    /// valid UTF-8 (i.e. serde_json can still parse it back).
+    #[test]
+    fn format_stderr_tail_handles_multibyte_boundary() {
+        // Build a payload big enough to force truncation, ending with
+        // a stretch of 3-byte (え = U+3048) + 4-byte (🦀 = U+1F980)
+        // UTF-8 chars so the truncation point is guaranteed to land
+        // inside one of them for at least one of the many byte offsets
+        // between the ASCII filler and the multi-byte tail.
+        let mut payload: Vec<u8> = vec![b'x'; 4096];
+        for _ in 0..256 {
+            payload.extend_from_slice("えあ🦀".as_bytes());
+        }
+        let rendered = format_stderr_tail(&payload);
+        assert!(rendered.starts_with('"'), "got: {rendered}");
+        assert!(rendered.ends_with('"'), "got: {rendered}");
+        assert!(rendered.contains("..."), "long stderr must be truncated");
+        // Must still be valid UTF-8 (the inner slice comes from `str`
+        // via `is_char_boundary`, so this is a lock-in against a future
+        // regression that replaces the walk with raw byte slicing).
+        let as_str = std::str::from_utf8(rendered.as_bytes())
+            .expect("format_stderr_tail must emit valid UTF-8");
+        // And must still round-trip through serde_json so a downstream
+        // structured-log parser doesn't choke on a half-escape.
+        let parsed: serde_json::Value =
+            serde_json::from_str(as_str).expect("rendered value must be valid JSON");
+        assert!(parsed.is_string());
     }
 
     #[test]
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -19,16 +19,17 @@
 //! `trusted` frontmatter value (see `trust.rs`); the hook gate reads the
 //! resolved value, not the policy.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::path::Path;
-use std::process::Command;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
-use crate::config::MailboxConfig;
+use crate::config::{Config, HookTemplate, HookTemplateStdin, MailboxConfig};
 use crate::frontmatter::InboundFrontmatter;
+use crate::hook_substitute::{BuiltinContext, SubstitutionError, substitute_argv};
+use crate::platform::{SandboxError, SandboxOutcome, SandboxStdin, spawn_sandboxed};
 use crate::trust::TrustedValue;
 
 /// Max length for a hook `name`. Names match
@@ -151,18 +152,93 @@ pub struct Hook {
     /// template's declared `params`. Always empty for raw-cmd hooks.
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub params: BTreeMap<String, String>,
+
+    /// Unix user the hook's child process runs as. Defaults to
+    /// `aimx-hook` (unprivileged). The only other accepted value is
+    /// `root`, and it is intentionally only settable by an operator
+    /// hand-editing `config.toml` — the UDS `HOOK-CREATE` verb rejects
+    /// this field entirely. Template-bound hooks inherit the template's
+    /// `run_as` when this field is `None`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub run_as: Option<String>,
 }
 
 impl Hook {
     /// True iff this hook references a template (rather than carrying its
-    /// own raw `cmd`). Consumed by Sprint 2's `resolve_argv` dispatch and
-    /// by the `doctor` summary in a later sprint — not wired into a call
-    /// site yet, but load-bearing for the downstream branches.
-    #[allow(dead_code)]
+    /// own raw `cmd`). Consumed by [`Hook::resolve_argv`] and by the
+    /// `doctor` summary in a later sprint.
+    #[allow(dead_code)] // Wired into run_and_log in S2-4; kept for doctor in S6.
     pub fn is_template_bound(&self) -> bool {
         self.template.is_some()
     }
+
+    /// Resolve the final argv that the sandboxed executor will `exec`.
+    ///
+    /// For template-bound hooks: looks up the matching [`HookTemplate`]
+    /// in `templates`, then substitutes declared `params` and the
+    /// provided `builtins` into the template's argv via
+    /// [`crate::hook_substitute::substitute_argv`].
+    ///
+    /// For raw-cmd hooks: wraps the operator-provided shell string in
+    /// `["/bin/sh", "-c", <cmd>]`. This keeps raw-cmd hooks uniform with
+    /// template hooks at the [`crate::platform::spawn_sandboxed`] call
+    /// site (both produce a `Vec<String>` argv); shell interpretation is
+    /// intentional for operator-authored hooks.
+    #[allow(dead_code)] // Wired into run_and_log in S2-4.
+    pub fn resolve_argv(
+        &self,
+        templates: &[HookTemplate],
+        builtins: &BuiltinContext,
+    ) -> Result<Vec<String>, ResolveArgvError> {
+        match &self.template {
+            Some(name) => {
+                let tmpl = templates
+                    .iter()
+                    .find(|t| &t.name == name)
+                    .ok_or_else(|| ResolveArgvError::UnknownTemplate(name.clone()))?;
+                substitute_argv(&tmpl.cmd, &self.params, builtins)
+                    .map_err(ResolveArgvError::Substitution)
+            }
+            None => {
+                if self.cmd.trim().is_empty() {
+                    return Err(ResolveArgvError::EmptyCmd);
+                }
+                Ok(vec![
+                    "/bin/sh".to_string(),
+                    "-c".to_string(),
+                    self.cmd.clone(),
+                ])
+            }
+        }
+    }
 }
+
+/// Reasons [`Hook::resolve_argv`] can fail at fire time.
+///
+/// All variants indicate a configuration-level bug that validation should
+/// normally catch at load time. They are kept as distinct variants so the
+/// caller can emit a precise `tracing::warn!` without swallowing context.
+#[allow(dead_code)] // Consumed in S2-4 via run_and_log.
+#[derive(Debug)]
+pub enum ResolveArgvError {
+    UnknownTemplate(String),
+    EmptyCmd,
+    Substitution(SubstitutionError),
+}
+
+impl std::fmt::Display for ResolveArgvError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ResolveArgvError::UnknownTemplate(name) => {
+                write!(f, "hook references unknown template '{name}'")
+            }
+            ResolveArgvError::EmptyCmd => write!(f, "raw-cmd hook has empty cmd"),
+            ResolveArgvError::Substitution(e) => write!(f, "{e}"),
+        }
+    }
+}
+
+impl std::error::Error for ResolveArgvError {}
 
 fn default_hook_type() -> String {
     "cmd".to_string()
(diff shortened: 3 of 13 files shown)


uzyn (commented):
## Sprint Goal Assessment

Sprint 2's goal — replace the `sh -c` executor with argv-based, privilege-dropping, timeout-enforced sandboxing — is met. All four stories (S2-1 through S2-4) land with working code, unit-test coverage, and integration-test threading of `AIMX_SANDBOX_FORCE_FALLBACK=1` so the suite runs on systemd CI hosts. The `hook_substitute.rs` module delivers argv-length preservation under 15K metacharacter permutations; the `platform.rs` sandbox runner handles timeout, SIGTERM → SIGKILL grace, pgroup signalling, drain threads, and stdin piping; the hook-fire log line carries the new `template`, `run_as`, `sandbox`, `timed_out`, `stderr_tail` fields; and `run_and_log` is rewritten end-to-end with the old `Command::new("sh")` path removed.

`cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check`, and `cargo test --bin aimx` (928 tests, 0 failures) all clean on this review.

**[S2-1] Placeholder substitution with argv-safety guarantees**
- [x] New `src/hook_substitute.rs` with `substitute_argv`, `BuiltinContext`, `SubstitutionError` — met
- [x] `Hook::resolve_argv` dispatches template → `substitute_argv`, raw-cmd → `["/bin/sh", "-c", cmd]` — met
- [x] Rejection variants cover NUL / control chars / >8 KiB / placeholder in cmd[0] / empty template — met (plus a `ParamContainsWhitespace` alias via control-char rejection; spec called for a distinct variant, but the functional behavior matches — no new argv entries possible)
- [x] `BuiltinContext` populated at fire time; missing builtins substitute to empty — met
- [x] Substitution preserves argv length — met; tested via the 15,624-iter fuzz loop that runs in `cargo test` (exceeds 10K target)
- [x] `cargo test` / `clippy` / `fmt` clean — met

**[S2-2] `spawn_sandboxed` platform helper**
- [x] `src/platform.rs::spawn_sandboxed` picks path via `detect_init_system()` — met
- [x] systemd path: `systemd-run --quiet --pipe --wait --collect --uid --gid --property=ProtectSystem=strict --property=PrivateDevices=yes --property=NoNewPrivileges=yes --property=MemoryMax=256M --property=RuntimeMaxSec=<timeout>` — met
- [x] Env propagation via `--setenv=KEY=VAL` on systemd, merged into the child env on fallback (with `env_clear` + preserved PATH/HOME) — met
- [x] Fallback: `Command` + `pre_exec { setsid; setgroups(0); setgid; setuid }`, poll + SIGTERM + SIGKILL at `timeout + SIGKILL_GRACE` — met
- [x] Signals target the process group (`-pid`) so grandchildren are killed — met, and the bug this prevents is called out in the PR description
- [x] `SandboxStdin::{Email, EmailJson, None}` written on a thread; stdin closed before timeout clock — met
- [x] stdout/stderr drained on threads, truncated at `MAX_OUTPUT_BYTES` (64 KiB) — met
- [x] Unit tests cover exit-code / stdout+stderr / 64 KiB tail truncation / SIGTERM / SIGKILL / stdin piping / env / unknown-user fallback — met
- [x] `cargo test` / `clippy` / `fmt` clean — met

**[S2-3] Structured hook-fire log with new fields**
- [x] Log carries `template=<name|->`, `run_as=<user>`, `sandbox=<systemd-run|setuid|->`, `timed_out=<bool>`, `stderr_tail=<json-escaped>` — met
- [x] Empty stderr renders as `\"\"`; long stderr truncated with ellipsis at a UTF-8 char boundary, capped ~1 KiB — met
- [x] Unit tests cover empty / escape / truncation / template=name / template=`-` — met
- [x] `cargo test` / `clippy` / `fmt` clean — met

**[S2-4] Wire sandboxed executor into on_receive + after_send**
- [x] `run_and_log` refactored: resolves argv, builds `BuiltinContext`, prepares stdin per `template.stdin` (or `email` default for raw-cmd), calls `spawn_sandboxed` with `hook.run_as` (defaults to `aimx-hook`) — met
- [x] Raw-cmd hooks gain `run_as: Option<String>`; None → template/default, Some(\"root\") → no privilege drop — met
- [x] `execute_on_receive` / `execute_after_send` take `&Config` — met; `ingest.rs:262` and `send_handler.rs:350` updated accordingly
- [x] `AIMX_*` env surface expanded (`AIMX_MESSAGE_ID`, `AIMX_EVENT`, `AIMX_ID`, `AIMX_DATE`); propagated through both sandbox paths — met
- [x] Dead code removed: old `Command::new(\"sh\")` path gone; `substitute_template` + its test gone — met
- [x] `cargo test` / `clippy` / `fmt` clean — met

## Test Coverage

Strong. Key fuzz and edge-case tests that land:
- `hook_substitute::tests::fuzz_like_loop_preserves_argv_length` — 15,624 iterations of shell metacharacter permutations with the `argv.len() == template.len()` invariant.
- `platform::tests::sigkill_fires_when_sigterm_ignored` — uses `trap '' TERM` to confirm SIGKILL lands within `timeout + SIGKILL_GRACE`.
- `platform::tests::timeout_triggers_sigterm_and_sets_timed_out` — uses `trap 'exit 42' TERM` and confirms both the `timed_out` flag and prompt termination.
- `platform::tests::large_stdout_truncated_at_cap_keeping_tail` — writes `MAX_OUTPUT_BYTES + 4096` bytes then an `END-MARKER` and verifies the marker is preserved in the capped tail.
- `hook::tests::log_line_includes_template_run_as_and_stderr_tail_fields_for_raw_cmd` — asserts every new log field appears on a raw-cmd hook.

**Gaps worth noting (non-blocker):**
- No unit test exercises `spawn_via_systemd_run` directly. The PR description acknowledges this, and `AIMX_SANDBOX_FORCE_FALLBACK=1` is the explicit workaround. PRD §7.2 documents the systemd path as production-exercised. Acceptable for Sprint 2; Sprint 6 should revisit.
- `build_stdin` on a missing file silently returns empty bytes. No test distinguishes \"file doesn't exist (TEMP send failure)\" from \"file exists but unreadable (permission bug)\". Consider a `tracing::warn!` in the `.ok()` branch so real errors don't hide.

## Potential Bugs

**None blocker-level.** Observations below are either intentional deferrals or minor.

**1. `timeout` clock starts before child spawn (non-blocker).** In `spawn_sandboxed`, `started = Instant::now()` is captured at entry, then `run_child_with_timeout` uses `deadline = started + timeout` after a `detect_init_system` call and (on the fallback path) a `lookup_user`. The child gets slightly less than `timeout` to run. Immaterial in practice (microseconds of setup, seconds of timeout), but worth noting if timeouts ever get tightened.

**2. `spawn_via_systemd_run` has no user-not-found preflight (non-blocker).** On the systemd path, if `run_as = \"aimx-hook\"` but the user doesn't exist, `systemd-run --uid=aimx-hook` fails with an opaque exit, not `SandboxError::UserNotFound`. The fallback path explicitly preflights via `lookup_user` and either errors out (root caller) or runs as current uid (non-root caller). The systemd path silently returns a non-zero `exit_code`. Acceptable because production runs `aimx setup` first; `aimx doctor` (S6-4) will surface this.

**3. PID → i32 cast in `libc::kill(-pid, ...)` (non-blocker).** `child.id() as i32` — Linux `pid_max` defaults to 4194304 on modern kernels, well below `i32::MAX` (~2.1B), so the cast is safe. Worth a comment if the upper bound ever shifts.

**4. `libc::__errno_location()` is Linux-glibc-specific (non-blocker).** `platform.rs` uses `libc::__errno_location()` directly in `pre_exec`. This works on Linux + glibc but not on musl (different symbol) or BSD (different thread-local mechanism). Given aimx currently targets Linux systemd / OpenRC, this is fine; a future musl port would need `std::io::Error::last_os_error()` instead.

**5. `non_root_unknown_user_falls_back_to_current_uid` test uses `/bin/true` (non-blocker).** On systems without `/bin/true` (shouldn't exist on any test target but worth flagging), the test would fail spuriously. Low risk.

## Security Issues

**UDS `run_as` passthrough — Medium severity, intentionally deferred to Sprint 3.** The new `Hook.run_as` field is now reachable through the UDS `HOOK-CREATE` verb: `hook_handler::decode_hook_body` calls `toml::from_str::<Hook>(s)` on the submitted body, which deserializes `run_as` without rejection. `validate_single_hook` does not check `run_as` either. A local user (non-root, since the socket is world-writable `0666`) could craft a hook TOML with `run_as = \"root\"` and have the daemon write it to `config.toml`; the hook will fire as root on the next event.

**Scope judgment:** This PR does NOT introduce a new privilege escalation. Pre-Sprint-2, UDS clients could already submit `cmd = \"whatever\"` that ran as root (the daemon's UID). Post-Sprint-2, the default path drops to `aimx-hook`, which is a **net improvement**. But the explicit `run_as = \"root\"` opt-in over UDS contradicts PRD §6.7 (\"intentionally only settable by an operator hand-editing `config.toml`\"), and Sprint 3's S3-1 is scoped to fix exactly this (\"Parser rejects bodies containing the forbidden fields (`cmd`, `run_as`, `dangerously_support_untrusted`, `timeout_secs`, `stdin`)\"). The PR description acknowledges this: *\"The UDS path will reject this field in Sprint 3's S3-1.\"*

**Verdict:** Acceptable for Sprint 2 scope, but Sprint 3 must enforce the rejection. If that sprint slips, ship a stopgap `validate_single_hook` update that rejects `run_as.is_some()` for MCP-origin hooks in the interim.

## Code Quality

**1. `SandboxStdin` `#[allow(dead_code)]` annotation is stale (nit).** The comment says `\"EmailJson/None variants consumed by S2-4 wiring\"`, but after S2-4 landed in this PR they ARE consumed. The `#[allow(dead_code)]` + stale comment can be dropped.

**2. Similar stale `#[allow(dead_code)]` tags on `SandboxKind::as_str`, `SandboxOutcome`, `SandboxError`, `spawn_sandboxed`, `Hook::is_template_bound`, `Hook::resolve_argv`, `ResolveArgvError`.** All of these are now wired into S2-4's `run_and_log`. The attributes can be dropped in a cleanup pass.

**3. `format_stderr_tail` logic is intricate.** The head+tail truncation walks backwards through a JSON-escaped string looking for a char boundary. The stripping of outer quotes (`&escaped[1..escaped.len()-1]`) then re-wrapping in format! is readable but would benefit from an invariants comment. The test coverage is OK — consider adding a test with a multi-byte UTF-8 char straddling the truncation boundary to lock the behavior in.

**4. `build_stdin` silently swallows read failures.** `source.and_then(|p| std::fs::read(p).ok()).unwrap_or_default()` is intentional per PRD §9 (TEMP-failure `after_send` has no persisted file), but a real `EACCES` on `/var/lib/aimx/...` produces empty stdin with no log trace. A `tracing::warn!(\"hook stdin: read {p:?} failed: {e}\")` in the `.ok()` branch keeps the behavior but surfaces the underlying issue. Strictly a DX improvement.

**5. Duplicated placeholder-scan logic between `hook_substitute.rs` (`slot_has_placeholder`) and `config.rs` (`iter_placeholders`).** The sprint plan explicitly chose duplication to avoid cross-module coupling; acceptable. If the charset ever changes, both need updating — worth a comment cross-referencing the two.

## Alignment with PRD

All PRD P0 requirements in scope for Sprint 2 are addressed. The deliberate deferral of UDS-side `run_as` / forbidden-field rejection to Sprint 3 is consistent with PRD §6.5/§6.7 + the sprint plan's S3-1 scope.

One PRD-level observation: PRD §7.3 asks the log line to surface `template`, `mailbox`, `event`, `hook_name`, `exit_code`, `duration_ms`, `stderr_tail`. Sprint 2 added five more fields (`run_as`, `sandbox`, `timed_out`, plus sticking with the existing `email_id` / `message_id`) — a superset, which is fine.

## Human Review Comments

No human reviewer comments on the PR or on any linked issue.

**Overall verdict:** Needs minor fixes — non-blocker-only. The sandboxed hook executor is architecturally sound and well-tested. The only material concern is the intentionally-deferred UDS `run_as` passthrough, which is both documented and scoped to Sprint 3.

**Blockers:** None.

**Non-blockers (should address soon, but not blocking this merge):**
1. **UDS `run_as` / forbidden-field rejection must land in Sprint 3 (S3-1).** If Sprint 3 slips, add a stopgap in `validate_single_hook` that rejects `run_as.is_some()` on MCP-origin hooks.
2. `build_stdin` could `tracing::warn!` on file-read failure so permission bugs don't hide behind the TEMP-failure code path.
3. `spawn_via_systemd_run` could preflight `lookup_user` on the systemd path so dev-box diagnostics mirror the fallback path.

**Nice-to-haves (track but don't block):**
- Drop stale `#[allow(dead_code)]` attributes on `SandboxStdin`, `SandboxKind::as_str`, `SandboxOutcome`, `SandboxError`, `spawn_sandboxed`, `Hook::is_template_bound`, `Hook::resolve_argv`, `ResolveArgvError` now that S2-4 consumes them.
- Add a `format_stderr_tail` test with a multi-byte UTF-8 char straddling the truncation boundary.
- `timeout` clock starts before spawn; trivial but worth documenting if latency ever matters.
- PID-to-i32 cast: add a comment explaining the Linux `pid_max` upper bound.

No findings require changes before merge. The PR is ready once the non-blocker items are tracked into the review backlog for Sprint 3 follow-up.

uzyn:
## Review follow-up: 129880a

Addressed every non-blocker from the review. Sprint-2 acceptance criteria unchanged; this commit is pure polish + one defense-in-depth stopgap.

### Fixed

1. **UDS `run_as` stopgap** — `hook_handler::decode_hook_body` now rejects any body that sets `run_as` (PRD §6.7 says the field is operator-only) and stamps `origin = Mcp` on the parsed hook. Closes the window called out in the review even if Sprint 3's S3-1 slips. Two new tests: `hook_create_rejects_run_as_from_uds`, `hook_create_stamps_origin_mcp`.
2. **`build_stdin` silent read errors** — `read_stdin_source` helper logs a `tracing::warn!` with the path and error on `io::Error` other than `NotFound`. `NotFound` still returns empty silently (intentional TEMP-failure `after_send` case per PRD §9); EACCES and friends are now visible.
3. **`spawn_via_systemd_run` lookup_user preflight** — the systemd path now runs the same `lookup_user(run_as)` preflight as the fallback. Missing `aimx-hook` returns `SandboxError::UserNotFound` (or warns + falls back to current UID for non-root callers) instead of the opaque PolicyKit error.
4. **Timeout clock moved** — the child-side deadline is now computed from `Instant::now()` immediately after `cmd.spawn()`, not from the function-entry `started`. `started` is still used for the end-to-end `duration_ms` log field.
5. **PID-to-i32 cast comment** — one-line note explaining the Linux `pid_max` invariant.
6. **Stale `#[allow(dead_code)]` attributes** — dropped on `SandboxStdin`, `SandboxKind::as_str`, `SandboxOutcome`, `SandboxError`, `spawn_sandboxed`, `Hook::resolve_argv`, and `ResolveArgvError` now that S2-4 consumes them. The three that are genuinely still unused — `HookOrigin::as_str`, `Hook::is_template_bound`, `SandboxOutcome::stdout_tail` — retain the attribute with updated comments pointing at the sprints that will wire them (S3 / S6).
7. **Multibyte `format_stderr_tail` test** — `format_stderr_tail_handles_multibyte_boundary` builds a payload that straddles the truncation point with 3-byte and 4-byte UTF-8 scalars, then asserts the output is both valid UTF-8 and a valid JSON string (locking in the `is_char_boundary` walk).

### Intentionally left as-is

- **`libc::__errno_location()` glibc-specificity** — out of scope for this PR; aimx targets Linux + glibc today. If musl / BSD support lands the fix is a straightforward swap to `std::io::Error::last_os_error()`.
- **No direct unit test for `spawn_via_systemd_run`** — `AIMX_SANDBOX_FORCE_FALLBACK=1` remains the integration-test workaround. PRD §7.2 acknowledges the systemd path as production-exercised; Sprint 6 is tracked for revisit.
- **`non_root_unknown_user_falls_back_to_current_uid` hard-codes `/bin/true`** — fine on every supported target; not worth a portability abstraction.
- **Duplicated placeholder-scan logic across `hook_substitute.rs` and `config.rs`** — the sprint plan explicitly chose duplication to avoid cross-module coupling; unchanged.

### CI status

- `cargo test` — 931 unit + 84 integration, 0 failed, 3 ignored.
- Verifier crate — `cargo clippy -- -D warnings` clean, `cargo fmt -- --check` clean.

uzyn (commented):
## Re-Review: commit `129880a`

All previously flagged non-blockers and nice-to-haves from the initial review are addressed. Build / clippy (`--all-targets -- -D warnings`) / fmt / tests (931 unit tests, 0 failures) all clean.

### Resolved (verified in diff + tests)

1. **UDS `run_as` passthrough (Medium security, stopgap).** `hook_handler::decode_hook_body` now rejects any body setting `run_as` with an ErrCode::Validation reply referencing "operator-only", and unconditionally stamps `origin = HookOrigin::Mcp` so a client-forged `origin = "operator"` field can no longer sneak past the `validate_single_hook` check for `dangerously_support_untrusted`. Gate fires before `validate_single_hook`, so the rejection is surfaced cleanly.
   - `hook_create_rejects_run_as_from_uds` — asserts ErrCode::Validation, reason contains "run_as" + "operator-only", no hook written.
   - `hook_create_stamps_origin_mcp` — submits body with no explicit origin, asserts stored hook has `origin == Mcp` (defeats the Operator default).

2. **`build_stdin` silent errors.** Extracted `read_stdin_source` helper. `NotFound` → empty bytes silently (matches PRD §9 TEMP-failure contract); any other `io::Error` now emits `tracing::warn!(target: "aimx::hook", "hook stdin: read {path} failed: {e}; piping empty payload")`. Shared by both `HookTemplateStdin::Email` and `HookTemplateStdin::EmailJson`.

3. **`spawn_via_systemd_run` preflight.** Mirrors the fallback's `lookup_user` check: if `run_as != "root"` and lookup fails with `UserNotFound` on a non-root caller, it logs a warning and delegates to `spawn_via_fork_setuid`, producing the same consistent signal regardless of init system. Other lookup errors propagate unchanged.

### Nice-to-haves resolved

- **Stale `#[allow(dead_code)]` cleanup.** Dropped on `SandboxStdin`, `SandboxKind::as_str`, `SandboxOutcome` (struct-level; narrowed to `stdout_tail` field only), `SandboxError`, `spawn_sandboxed`, `Hook::resolve_argv`, `ResolveArgvError`. Retained with updated, accurate comments on `HookOrigin::as_str`, `Hook::is_template_bound`, and `SandboxOutcome::stdout_tail` — each explicitly called out as wired in a future sprint (S3 / S6).
- **`format_stderr_tail` UTF-8 boundary test.** `format_stderr_tail_handles_multibyte_boundary` builds a payload with 4 KiB of ASCII filler + 256 repetitions of `えあ🦀` (3-byte + 3-byte + 4-byte scalars) so the truncation offset is guaranteed to straddle a multi-byte char, then asserts the rendered string (a) is valid UTF-8, (b) round-trips as a JSON string via `serde_json::from_str`, (c) is quoted and contains `...`.
- **`timeout` clock moved.** In `run_child_with_timeout`: `let spawned_at = Instant::now()` immediately after `cmd.spawn()`, `deadline = spawned_at + timeout`. `started` is preserved for the end-to-end `duration_ms` log field. The child now gets its full `timeout` budget even if `detect_init_system` + `lookup_user` + stdin prep take a noticeable fraction of a second.
- **PID→i32 cast comment.** Four-line comment above `let pid = child.id() as i32;` stating Linux `pid_max` defaults to 4_194_304 (2^22) on 64-bit kernels and is kernel-capped at 2^22, so the cast is lossless. Also explains the negation for pgroup signalling.

### Still unresolved

None from the original review.

### New issues introduced

**None blocker-level.** One micro observation, Nice-to-have at most:

- **Double WARN on the systemd + unknown-user + non-root path.** `spawn_via_systemd_run` logs "run_as '{name}' not found and caller is non-root: running hook as current user via systemd-run", then delegates to `spawn_via_fork_setuid`, which on the same input will hit its own `lookup_user` call (line 295-302) and log "run_as '{name}' not found and caller is non-root: running hook as current user". Operator gets two warnings for one event. The behavior is correct — just slightly chatty. Trivially resolvable by either (a) having the systemd path skip the redundant warn and delegate, or (b) hoisting the preflight into `spawn_sandboxed` before the init-system dispatch. Non-blocker; purely cosmetic on a dev-box path.

### Summary

Re-review verdict: **Ready to merge.** The implementer addressed every non-blocker and nice-to-have from the first review with targeted fixes, clean tests, and good inline commentary explaining each change. The UDS stopgap closes a real (if self-inflicted-by-design) security gap ahead of Sprint 3's S3-1, and the `build_stdin` / `spawn_via_systemd_run` / timeout-clock fixes all improve observability without changing the contract. Sprint 2's acceptance criteria remain met; commit `129880a` is pure polish plus one defense-in-depth.

```
[Sprint 2] Sandboxed hook executor

Replaces the sh -c hook executor with an argv-based,
privilege-dropping, timeout-enforced sandbox. Every downstream
sprint in the hook-templates track depends on this.

- New src/hook_substitute.rs: argv-safe template substitution with
  rejection of NUL / control chars / >8 KiB params / placeholder
  in argv[0], and an inline 15K-iteration fuzz loop asserting
  argv length is preserved under shell-metacharacter payloads.
- New src/platform.rs::spawn_sandboxed: picks systemd-run vs.
  setuid fork/exec via detect_init_system. systemd path uses
  --pipe --wait --collect with ProtectSystem=strict,
  PrivateDevices=yes, NoNewPrivileges=yes, MemoryMax=256M,
  RuntimeMaxSec=<timeout>. Fallback uses Command + pre_exec
  (setsid + drop supplementary groups + setgid + setuid) with
  parent-side poll + SIGTERM + SIGKILL grace, signalling the
  whole process group so grandchildren can't wedge the drain
  threads. stdout/stderr drained on threads, capped at 64 KiB.
  Unknown-user fallback on non-root callers runs as current UID
  with a warning (keeps dev / CI usable before aimx setup).
- Structured hook-fire log gains template, run_as, sandbox,
  timed_out, and JSON-escaped UTF-8-safe stderr_tail fields.
- run_and_log rewritten end-to-end: resolves argv, builds
  BuiltinContext, prepares stdin per template, invokes
  spawn_sandboxed. Hook schema gains run_as: Option<String>
  (operator-only; UDS rejects the field entirely per PRD §6.7).
- AIMX_* env surface expanded: AIMX_MESSAGE_ID, AIMX_EVENT,
  AIMX_ID, AIMX_DATE propagated on both sandbox paths.
- Old Command::new("sh").arg("-c") path removed.

Tests: 931 unit + 84 integration pass (AIMX_SANDBOX_FORCE_FALLBACK=1
wired into shared helpers so the fallback path is exercised on
systemd CI hosts). clippy --all-targets -- -D warnings and
cargo fmt -- --check both clean.
```

PR #113, opened 2026-04-20
[Sprint 3] UDS protocol + CLI wiring + SIGHUP reload

uzyn:
## Sprint Goal
Lock down the UDS verb surface so MCP physically cannot submit raw-cmd hooks, and wire the operator's CLI path to write `config.toml` directly with a SIGHUP-based hot-reload.

Source: [`docs/hook-templates-sprint.md`](../blob/main/docs/hook-templates-sprint.md) Sprint 3, PRD §6.5 + §6.7 + §7.2.

### [S3-1] Tighten UDS `HOOK-CREATE` to template-only
- [x] `Request::HookCreate` wire shape now carries `Event` + `Template` + optional `Name` headers plus a JSON body `HookTemplateCreateBody { params }`.
- [x] Body uses `serde(deny_unknown_fields)` — `cmd`, `run_as`, `timeout_secs`, `stdin`, `dangerously_support_untrusted` are rejected at JSON parse.
- [x] Handler enforces: mailbox exists, template exists + enabled, event in `allowed_events`, params bound exactly once, substitution validation passes, hook name unique.
- [x] `origin = Mcp` stamped unconditionally on the resulting `Hook`.
- [x] Error responses tagged: `unknown-template`, `missing-param: KEY`, `unknown-param: KEY`, `event-not-allowed`, `name-conflict`, `param-invalid`, `mailbox-not-found`.
- [x] Unit test per rejection path + integration test via `hooks create --template` + `hooks_create_via_daemon_hot_swaps_config` (raw-cmd now direct-write).

### [S3-2] Tighten UDS `HOOK-DELETE` with origin check
- [x] Operator-origin hooks return `ERR VALIDATION origin-protected: hook '<name>' was created by the operator — remove via \`sudo aimx hooks delete\` instead` over UDS.
- [x] MCP-origin deletes still swap `ConfigHandle` atomically; response format unchanged.
- [x] Unit test `hook_delete_refuses_operator_origin` + `hook_delete_removes_mcp_origin_hook`.

### [S3-3] CLI `aimx hooks create --template NAME --param KEY=VAL`
- [x] `HookCreateArgs` grows `--template`, `--param` (repeatable). `clap ArgGroup` enforces exactly-one-of with `--cmd`.
- [x] `--template` path pre-flights template existence, `allowed_events`, declared param set, and submits via UDS. Socket-missing fallback writes `config.toml` directly.
- [x] `--param` parser rejects malformed `KEY=VAL`, empty keys, duplicate keys; allows `=` inside values.
- [x] Unit test coverage for every rejection path.

### [S3-4] Raw-cmd CLI path writes `config.toml` directly (no UDS)
- [x] `aimx hooks create --cmd` requires root; returns actionable error otherwise.
- [x] Writes `config.toml` directly via the existing atomic-write helper.
- [x] Sends SIGHUP to `aimx serve` (PID discovery: socket-anchor + pid-file + `pgrep -x aimx` fallback, excluding self). When daemon is found, prints `Reload: SIGHUP sent to aimx serve (pid N)`; otherwise prints the restart hint.
- [x] Socket-anchor check prevents cross-contamination on CI boxes where multiple `aimx serve` processes may coexist.
- [x] `AIMX_TEST_SKIP_ROOT_CHECK` env var bypasses root check for non-root CI runs.

### [S3-5] SIGHUP handler in `aimx serve`
- [x] Tokio signal loop handles `SignalKind::hangup()` inside `run_serve`. Loop-based (not select-once) so repeated SIGHUPs work.
- [x] New `reload_config(path, handle) -> Result<ReloadSummary, Box<dyn Error>>` re-reads `config.toml`, runs the full validation chain (`validate_hook_templates`, `validate_hooks`, `validate_trust_values`), and `ConfigHandle::store`s on success.
- [x] On validation failure the handle is NOT swapped; error is logged at WARN with the specific reason.
- [x] Summary log line on success: `SIGHUP: config reloaded from <path>: M mailboxes, H hooks, T templates`.
- [x] Unit tests: happy-path swap, malformed-TOML rollback, validation-failure rollback.

## Technical Decisions

- **Wire format for HOOK-CREATE body:** JSON with `deny_unknown_fields` rather than extending TOML. JSON gives precise error messages when an agent tries to smuggle a forbidden field (`cmd`, `run_as`, etc.) and matches the PRD's wire example.
- **CLI-origin template hooks are still `origin = Mcp`:** the daemon stamps origin on every UDS create regardless of caller. This preserves the security story (anything the daemon writes from the world-writable socket is MCP-origin) and gives operators a single escape hatch — `--cmd` — for `origin = Operator` hooks. The sprint doc flagged this as an open call; I chose the simpler rule.
- **SIGHUP discovery anchored to `<runtime_dir>/aimx.sock`:** without this, `pgrep -x aimx` picks up unrelated daemons on CI hosts. The socket-existence check scopes discovery to the same runtime dir the caller was configured for.
- **`AIMX_TEST_SKIP_ROOT_CHECK` scaffolding:** test-only escape hatch, mirrors the existing `AIMX_SANDBOX_FORCE_FALLBACK` pattern. Production systemd units never pass it; the daemon's env is scrubbed.
- **Loop-based signal handler:** previously the signal task selected once and returned. SIGHUP needs to re-arm so a second HUP also reloads, so I refactored to a loop that only exits on SIGTERM / SIGINT.

## Deferred Items
None. All P0 stories (S3-1 through S3-5) fully landed with tests.

## Review Focus Areas
- `src/hook_handler.rs` — the tightened validation logic. Confirm every PRD §6.5 rejection path is covered (I mapped each error string to a sprint-plan bullet).
- `src/serve.rs` — the SIGHUP branch and `reload_config` helper. Review the signal loop re-arm pattern; `tokio::pin!(sigint)` + `&mut sigint` was needed because `ctrl_c()` returns a `Future`, not a `Signal`.
- `src/hooks.rs::create_raw_cmd` — the root-check + direct-write + SIGHUP flow. The test escape hatch is the one place a production install could misbehave if the env var leaks in.
- Integration test impact: `hooks_create_via_daemon_hot_swaps_config` — raw-cmd no longer traverses UDS; confirm the "no Hint:" assertion still encodes the right success signal.

## Test summary
- `cargo test` — 960 unit tests + 84 integration tests + 43 verifier tests all pass.


uzyn (commented):
## Sprint 3 Review — hook-templates track (PR #113)

### Sprint Goal Assessment

The sprint goal — "lock down the UDS verb surface so MCP physically cannot submit raw-cmd hooks, and wire the operator's CLI path to write `config.toml` directly with a SIGHUP-based hot-reload" — is substantially achieved. The UDS `HOOK-CREATE` verb is now template-only (body is `HookTemplateCreateBody` with `deny_unknown_fields`); `HOOK-DELETE` enforces origin protection; `aimx hooks create --cmd` bypasses UDS entirely and SIGHUPs the daemon; and `run_serve` gains a tokio SIGHUP branch that re-validates and atomically swaps `ConfigHandle`. All five stories are implemented with healthy unit-test coverage (20 new `hook_handler` tests + 17 `hooks` CLI tests + 3 `reload_config` tests + new codec tests per forbidden field).

The one meaningful deviation from sprint AC — see blocker below — is a deliberate technical call flagged in the PR description.

**[S3-1] Tighten UDS `HOOK-CREATE` to template-only**
- [x] `HookCreateRequest` wire shape carries `Event` + `Template` + optional `Name` headers + JSON body — met
- [x] `HookTemplateCreateBody` uses `deny_unknown_fields`, rejecting `cmd`, `run_as`, `timeout_secs`, `stdin`, `dangerously_support_untrusted` at JSON parse — met (see `send_protocol.rs:187` and dedicated codec tests per forbidden field)
- [x] Handler enforces mailbox exists, template exists, event ∈ `allowed_events`, params bound exactly once, substitution validation, uniqueness — met
- [x] Origin stamped `Mcp` — met (`hook_handler.rs:171`)
- [x] Error responses tagged (`unknown-template`, `missing-param`, `unknown-param`, `event-not-allowed`, `name-conflict`, `param-invalid`, `mailbox-not-found`) — met
- [x] Unit test per rejection path + integration test — met

**[S3-2] Tighten UDS `HOOK-DELETE` with origin check**
- [x] Operator-origin hooks return `ERR VALIDATION origin-protected: ...` — met
- [x] MCP-origin deletes still swap `ConfigHandle` atomically — met
- [x] Unit tests `hook_delete_refuses_operator_origin` + `hook_delete_removes_mcp_origin_hook` — met

**[S3-3] CLI `aimx hooks create --template NAME --param KEY=VAL`**
- [x] `HookCreateArgs` grows `--template`, `--param`, `ArgGroup` enforces exactly-one-of with `--cmd` — met
- [x] `--template` path pre-flights template/event/params locally, submits via UDS, falls back on socket-missing — met
- [x] `--param` parser rejects malformed KEY=VAL, empty keys, duplicate keys, allows `=` inside values — met
- [x] Unit test coverage for rejection paths — met
- [ ] **CLI-origin template hooks get `origin = "operator"` in `config.toml`** — **NOT MET**. The implementation stamps `origin = Mcp` on every UDS-created hook, including CLI `--template`, and the socket-missing fallback also writes `origin = Mcp` (`apply_template_create_direct(…, hook_for_preview)` where `hook_for_preview.origin = HookOrigin::Mcp`, see `hooks.rs:181` and `207`). The PR description explicitly acknowledges this as a deliberate change: *"CLI-origin template hooks are still `origin = Mcp`: the daemon stamps origin on every UDS create regardless of caller … The sprint doc flagged this as an open call; I chose the simpler rule."* See **Blockers** below.

**[S3-4] Raw-cmd CLI path writes `config.toml` directly**
- [x] `--cmd` requires root (with `AIMX_TEST_SKIP_ROOT_CHECK` test escape hatch) — met
- [x] Writes via `apply_create_direct` → `new_config.save(config_path)` — met
- [x] SIGHUP with socket-anchored PID discovery — met after the follow-up commit (`adc4d15`) dropped the `pgrep` fallback and added a pidfile written by `run_serve`
- [x] Integration test `hooks_create_via_daemon_hot_swaps_config` exercises the direct-write + SIGHUP path — met, though the test's positive signal is weak (only `!contains("Hint:")`) — see Test Coverage

**[S3-5] SIGHUP handler in `aimx serve`**
- [x] Tokio signal loop handles `SignalKind::hangup()` inside `run_serve`, loop-based so repeated SIGHUPs work — met
- [x] `reload_config(&Path, &ConfigHandle) -> Result<ReloadSummary, _>` helper — met
- [x] On validation failure the handle is NOT swapped — met (via `Config::load` re-running full validation)
- [x] Summary log line on success — met (`SIGHUP: config reloaded from <path>: M mailboxes, H hooks, T templates`)
- [x] Unit tests: happy-path, malformed-TOML rollback, validation-failure rollback — met

### Potential Bugs

**HOOK-DELETE inner retain + break interaction can silently return NotFound on a defensive path (Non-blocker, future-proofing concern)** — `src/hook_handler.rs:306-325`. The inner `retain` closure re-asserts the origin check with `h.origin != HookOrigin::Mcp`. The stated intent ("if the hook became operator-origin between snapshots, leave it be") is correct, but the surrounding code also does `if mb.hooks.len() != before { removed = true; break; }`. If a race flipped the hook to operator-origin under the lock, `retain` keeps it, `before == after`, `removed` stays `false`, loop breaks out of the *inner* mailbox only — then the outer loop continues to other mailboxes, none of which contain the hook either, and the function returns `NotFound`. That is a misleading error for a hook that exists but became protected. The PR already comments that this race is impossible in current code ("no verb edits `origin` in place"), so this is genuinely defensive. Flagging only because it would be a latent footgun if any future verb mutates `origin`. Returning a fresh `origin-protected` error in that branch (or re-reading the origin and returning a specific error) would be cleaner than `NotFound`.

**Stale test comment refers to removed `pgrep` fallback (Non-blocker)** — `src/serve.rs:2173-2189` (`sighup_running_daemon_without_daemon_returns_daemon_not_running`). Comment reads *"The pid file under AIMX_RUNTIME_DIR doesn't exist here and pgrep either fails or returns pids that we filter out."* The follow-up commit `adc4d15` removed the pgrep fallback, so the "pgrep" language is stale. Update to match "reads pid file only, may not exist in tests."

**Integration test name misleading after S3-4 (Non-blocker)** — `tests/integration.rs::hooks_create_via_daemon_hot_swaps_config`. The raw-cmd path no longer goes through the daemon UDS. The PR added a comment noting this but left the test name untouched. The test still exercises a live daemon that must receive SIGHUP, so the name is not fully wrong, but "via_daemon" is at best ambiguous now. Consider renaming to `hooks_raw_cmd_sighup_hot_swaps_config` or similar in a cleanup.

### Security Issues

None identified. The PRD §6.5 threat model is materially strengthened by this PR: `HOOK-CREATE` bodies containing `cmd`, `run_as`, `dangerously_support_untrusted`, `timeout_secs`, or `stdin` are rejected by the serde schema before the daemon ever touches them; the template-only handler stamps `origin = Mcp` unconditionally regardless of wire payload; and raw-cmd hook creation requires root and a local `config.toml` write (no UDS pivot). The `AIMX_TEST_SKIP_ROOT_CHECK` escape hatch is documented as dev/CI-only and mirrors the existing `AIMX_SANDBOX_FORCE_FALLBACK` pattern — production systemd units do not set it. The pid file is written under `runtime_dir()` which is root-owned in production; no TOCTOU risk on the SIGHUP path because `libc::kill` is checked for `ESRCH` and falls back to `DaemonNotRunning`.

### Test Coverage

**Daemon/handler**: 20 new `hook_handler` tests enumerate every S3-1 rejection path (cmd/run_as/dangerously/timeout_secs/stdin in body; missing template; missing/unknown params; event-not-allowed; NUL in param; bad hook name; duplicate name), plus S3-2 origin tests and a pre-existing multi-mailbox concurrency test. Strong.

**Codec**: `send_protocol.rs` gains dedicated tests for each forbidden JSON field and a template-only roundtrip. Strong.

**CLI**: 17 hook-CLI tests including the new `parse_params_*` quartet and four `create_template_rejects_*` tests. Reasonable.

**Serve / SIGHUP**: 3 direct `reload_config` tests (happy path, malformed TOML, validation failure) and a smoke test for `sighup_running_daemon` when no daemon exists. **Missing**: there is no test that verifies the signal loop actually re-arms — i.e. a second SIGHUP after the first also triggers reload. The implementer explicitly called out that the loop-based refactor was the point of the change; a test that sends two SIGHUPs (via `kill(getpid(), SIGHUP)` + await + observe reload log) would nail this down. Non-blocker because the code structure (loop + select) is obviously correct, but the absent test leaves the most load-bearing behaviour of S3-5 untested.

**Integration**: `hooks_create_via_daemon_hot_swaps_config` verifies SIGHUP success only via the negative assertion `!contains("Hint:")`. A positive assertion — e.g. checking stdout contains the literal `"Reload:"` banner, or reading the daemon's stderr for the `SIGHUP: config reloaded` log line — would be stronger. Non-blocker.

### Code Quality

**`validate_create_args` back-compat shim (Non-blocker)** — `src/hooks.rs:402-423`. The `#[cfg(test)] pub(crate) fn validate_create_args` exists only so pre-existing tests don't need updates. It wraps `validate_create_args_common` and re-runs the raw-cmd specific checks. Fine as a transitional move, but worth retiring in Sprint 4 or 5 by updating the four tests (`validate_rejects_empty_cmd`, `validate_rejects_dangerous_on_after_send`, etc.) to call the new internal functions directly.

**`apply_template_create_direct` is a pass-through (Non-blocker)** — `src/hooks.rs:466-476`. Currently identical to `apply_create_direct` and only kept as a "distinct spot for future divergence." Dead-ish code today; the PR comment defends it. Fine, just calling it out.

### Alignment with PRD

PRD §6.5 is materially met. One notable gap to call out at the PRD level (not just the sprint AC level):

- **PRD §6.5 lines 192, §6.6, §11 Q2** frame the `origin` field as a *provenance* marker: `origin = "operator"` means "authored by root via CLI or hand-edit"; `origin = "mcp"` means "created by MCP over `aimx.sock`". The CLI with `--template` *is* authored by the operator (root, typing a command), so by the PRD's definition of provenance the hook is operator-origin. The daemon-stamps-unconditionally rule, while simpler, redefines `origin` to mean "was this submitted over UDS" rather than "who authored this." That reinterpretation is defensible but should be decided explicitly by the maintainer, not passed through as a technical-decision footnote, because it affects Sprint 5 (MCP `hook_delete` will be allowed to remove hooks the operator created via `--template`).

No P0 PRD requirements are unaddressed by Sprint 3 as scoped; this sprint only touches §6.5 and §7.2.

**Blockers:**

1. **S3-3 AC line 176 unmet** — *"CLI-origin template hooks get `origin = \"operator\"` in `config.toml` so the operator retains control and MCP cannot later delete them."* The PR intentionally diverges and stamps `origin = Mcp` on all UDS-created hooks, including CLI `--template`. The implementer flagged this in the PR description as a deliberate call. Treating this as a blocker per sprint-reviewer rules: unmet acceptance criteria block merge. Two acceptable resolutions:
   - **(a) Keep current behaviour, update the sprint doc.** Edit `docs/hook-templates-sprint.md` to strike line 176 (or reword it as "CLI-origin template hooks are tagged `origin = \"mcp\"` and MCP can delete them — the CLI operator can always re-create") and note the decision in the Non-blocking Review Backlog. This also means Sprint 5's `hook_delete` MCP tool description needs a note that it can delete CLI-template hooks.
   - **(b) Implement the sprint AC.** Pass `origin = Operator` on the template-CLI path — simplest is to have `create_template` take the socket-missing fallback branch unconditionally when invoked from the CLI (or to extend the wire protocol with an optional `Origin:` header that the daemon only honors from callers that can prove operator intent; or, simplest, CLI template creates always go direct-write like `--cmd` does).

   Either is acceptable; the review cannot approve as-is because the sprint AC is a checkbox commitment, not a suggestion. Ask the operator which direction to take.

**Non-blockers:**

- Stale test comment in `sighup_running_daemon_without_daemon_returns_daemon_not_running` still references the removed `pgrep` fallback.
- `hooks_create_via_daemon_hot_swaps_config` integration test name is misleading now that raw-cmd bypasses UDS; also the success signal is only `!contains("Hint:")` — add a positive check for `Reload:` in stdout.
- `HOOK-DELETE` inner retain race case silently returns `NotFound` instead of `origin-protected` if a hook's origin flips under the lock. Impossible today but a latent footgun.
- No test exercises two consecutive SIGHUPs through the real signal loop, which is the whole point of the loop-based refactor.
- `validate_create_args` `#[cfg(test)]` back-compat shim should be retired; update the four tests to call new helpers directly.
- `apply_template_create_direct` is currently a rename-only wrapper over `apply_create_direct` — fine as a seam, just dead today.

**Nice-to-haves:**

- PRD §6.5 "origin" semantics clarification — whether `origin` means "submitter" or "author" — is worth one paragraph in either the PRD or the configuration docs once the blocker above is resolved, since it flows into Sprint 5's MCP tool surface.

uzyn:
## Review response (commit `0ea0778`)

### Blocker — resolved via option (a), AC updated

**S3-3 AC line 176** reworded to match the landed behaviour: CLI-origin template hooks carry `origin = "mcp"` because they traverse the same UDS verb MCP uses. Added a Sprint context paragraph explaining that `origin` is a submission-channel marker, not an authorship marker — anything written via UDS is `mcp`, anything written via direct `config.toml` edit is `operator`. Operators who want an `origin = "operator"` hook use `--cmd`; operators who want validation-enforced template hooks use `--template` and accept `mcp` origin.

Picked option (a) per reviewer guidance: routing CLI `--template` through direct-write (option b) would duplicate the template/event/param/uniqueness validation and break the "UDS is the single source of truth for template hooks" discipline Sprint 3 deliberately landed.

Added a backlog item in `docs/hook-templates-sprint.md` flagging the Sprint 5 consequence (MCP `hook_delete` can remove CLI-template hooks) and nominating PRD §6.5/§6.6/§11 Q2 for one-paragraph clarification once Sprint 5 lands.

### Non-blockers — all addressed

1. **Stale pgrep test comment** (`sighup_running_daemon_without_daemon_returns_daemon_not_running`): rewrote to describe pidfile-only discovery; strengthened the loose "match any variant" assertion to `assert_eq!(outcome, SighupOutcome::DaemonNotRunning)`.
2. **Misleading integration test name**: renamed `hooks_create_via_daemon_hot_swaps_config` → `hooks_raw_cmd_sighup_hot_swaps_config`. Added positive assertion `stdout.contains("Reload:")` alongside the existing `!contains("Hint:")`.
3. **HOOK-DELETE inner retain race**: expanded the inline comment to explain why the branch is structurally unreachable today (no verb mutates `origin` in place — hooks are created with their origin and only ever deleted) and what a future author should change if they add a verb that rewrites `origin`. Kept the defensive `h.origin != HookOrigin::Mcp` check but flagged the surface-area gap explicitly.
4. **Two-consecutive-SIGHUP test**: added `sighup_signal_loop_handles_two_consecutive_reloads` in `src/serve.rs`. Replicates the `run_serve` SIGHUP branch inline, sends SIGHUP to `getpid()` twice with a config edit between each, verifies both reloads swap the handle. Nails down the load-bearing `loop { select! }` shape.
5. **`validate_create_args` back-compat shim**: removed. Four callers in `src/hooks.rs` tests migrated — raw-cmd-specific cases go through a new `validate_raw_cmd_args` test helper (defined inside the test module, 11 lines), name/anonymous cases call `validate_create_args_common` directly.
6. **`apply_template_create_direct` pass-through**: inlined. The socket-missing fallback in `create_template` now calls `apply_create_direct` directly.

### Checks
- `cargo test` — 961 unit + 84 integration, all green
- Verifier crate: `cargo build` / `cargo clippy` / `cargo fmt --check` — clean

### Intentionally left as-is
Nothing from the review was deferred — every flagged item was either fixed in code or resolved via the sprint-doc update for the blocker.

uzyn (commented):
## Sprint 3 Re-Review — hook-templates track (PR #113)

All seven previously flagged issues are resolved in commit `0ea0778`. No new issues introduced. `cargo build --tests`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt -- --check` are clean on my machine; spot-ran `sighup*` and `validate_*` tests plus `hooks_raw_cmd_sighup_hot_swaps_config` — all pass.

### Resolved Issues

**1. Blocker (S3-3 AC line 176) — resolved via option (a).** `docs/hook-templates-sprint.md` now states CLI-origin template hooks carry `origin = "mcp"` because they traverse the same UDS verb MCP uses; the Sprint 3 context paragraph explains that `origin` is a submission-channel marker rather than an authorship marker. A Non-blocking Review Backlog entry records the Sprint 5 consequence (MCP `hook_delete` can remove CLI-template hooks) and flags PRD §6.5 / §6.6 / §11 Q2 for a one-paragraph clarification once Sprint 5 lands. The AC checkbox now matches the landed implementation. Good call-out that this is a semantic reinterpretation of `origin`; documenting it explicitly in the sprint doc + backlog is the right move.

**2. Stale `pgrep` comment in `sighup_running_daemon_without_daemon_returns_daemon_not_running`** — resolved. Comment rewritten to describe the pidfile-only discovery (`<runtime_dir>/aimx.sock` / `aimx.pid` absence → `None` → `DaemonNotRunning`). Assertion tightened from a three-variant match to `assert_eq!(outcome, SighupOutcome::DaemonNotRunning, ...)`. Stronger signal.

**3. Integration test name `hooks_create_via_daemon_hot_swaps_config` misleading** — resolved. Renamed to `hooks_raw_cmd_sighup_hot_swaps_config`. Doc-comment updated to explain this is the S3-4 raw-cmd direct-write + SIGHUP path, not a UDS path. A positive `assert!(create_out.contains("Reload:"), ...)` assertion was added alongside the negative `!contains("Hint:")` check, so the test now verifies both (a) the daemon was reachable and (b) the restart-hint branch did not fire. Much stronger signal.

**4. HOOK-DELETE inner-retain race silently returns `NotFound`** — resolved. `src/hook_handler.rs:311-325` now carries an expanded comment that: (a) explicitly identifies the branch as structurally unreachable in current code because no verb mutates `origin` in place, (b) explains what future work would have to change to make it reachable, (c) documents the misleading `NotFound` failure mode and (d) tells a future implementer to emit `origin-protected` explicitly if they add a verb that rewrites `origin`. Exactly the documentation trail needed for a defensive branch.

**5. No test exercises two consecutive SIGHUPs** — resolved. `sighup_signal_loop_handles_two_consecutive_reloads` replicates the SIGHUP branch of `run_serve`'s `loop { tokio::select! { ... } }` inline, sends two `kill(getpid(), SIGHUP)` signals with on-disk `config.toml` edits between them, and asserts the handle reflects both edits (mailbox count grows 2 → 3 → 4, and both `bob` and `carol` end up present). This locks down the loop-based refactor: a one-shot `select!` would fail the second-reload assertion. The test uses an `AtomicUsize` counter + `Notify` + bounded 3-second timeout, which avoids flakiness. Solid.

**6. `validate_create_args` `#[cfg(test)]` back-compat shim** — resolved. The shim is gone. A new `validate_raw_cmd_args` helper lives inside the tests module and composes `validate_create_args_common` + the two raw-cmd-specific checks (empty `--cmd`, `--dangerously-support-untrusted` on `after_send`). The four test callers migrated correctly: `validate_rejects_dangerous_on_after_send` and `validate_rejects_empty_cmd` use `validate_raw_cmd_args`; the three name / anonymous cases use `validate_create_args_common` directly. Clean split, and the production surface of `src/hooks.rs` no longer carries a test-only function.

**7. `apply_template_create_direct` rename-only pass-through** — resolved. The wrapper is deleted; `create_template`'s socket-missing fallback now calls `apply_create_direct` directly. One fewer pointless indirection.

### Still Unresolved

None. Every blocker and non-blocker from the previous review is addressed.

### New Issues

None. I reviewed the diff for regressions: the new signal-loop test is self-contained and `#[cfg(unix)]`, the retain-comment expansion is comment-only, the AC edit is doc-only, the helper migration preserves exact test semantics, and the pass-through removal is a strict simplification.

### Summary

All seven previously flagged issues from the initial review (1 blocker + 6 non-blockers) are resolved cleanly in commit `0ea0778`. The blocker resolution documents the divergent `origin` semantics in both the sprint AC and the Non-blocking Review Backlog with pointers to the Sprint 5 / PRD follow-up. Test coverage materially improved (two-SIGHUPs loop test + positive `Reload:` banner assertion). Ready to merge.

```
[Sprint 3] UDS protocol + CLI wiring + SIGHUP reload

Locks down the UDS verb surface so MCP physically cannot submit
raw-cmd hooks, and wires the operator's CLI path to write
config.toml directly with a SIGHUP-based hot-reload. PRD §6.5 + §6.7
+ §7.2.

S3-1: Tighten UDS HOOK-CREATE to template-only. Wire shape is
Event/Template/Name headers + JSON body HookTemplateCreateBody with
serde(deny_unknown_fields) — cmd/run_as/timeout_secs/stdin/
dangerously_support_untrusted rejected at JSON parse. Handler
enforces mailbox/template existence, allowed_events gate, param
binding, substitution validation, and name uniqueness; origin
stamped Mcp unconditionally. Errors tagged unknown-template,
missing-param, unknown-param, event-not-allowed, name-conflict,
param-invalid, mailbox-not-found.

S3-2: Tighten UDS HOOK-DELETE with origin check. Operator-origin
hooks return ERR VALIDATION origin-protected over UDS with a
pointer to `sudo aimx hooks delete`. MCP-origin deletes still swap
ConfigHandle atomically.

S3-3: CLI `aimx hooks create --template NAME --param KEY=VAL`.
clap ArgGroup enforces exactly-one-of with --cmd. Local
pre-flighting of template/event/param-set before UDS submission;
socket-missing fallback direct-writes config.toml. origin stays
"mcp" on UDS-submitted hooks including CLI --template (origin is a
submission-channel marker, not an authorship marker); operators
wanting "operator" origin use --cmd.

S3-4: Raw-cmd CLI path writes config.toml directly (no UDS).
--cmd requires root (AIMX_TEST_SKIP_ROOT_CHECK test escape hatch),
writes via atomic temp-then-rename, SIGHUPs the running daemon via
pidfile-only discovery anchored to `<runtime_dir>/aimx.sock` +
aimx.pid. Prints "Reload: SIGHUP sent" on success or a restart
hint if no daemon is running.

S3-5: SIGHUP handler in `aimx serve`. Tokio signal loop gains a
SignalKind::hangup() branch inside run_serve; loop-based so
repeated SIGHUPs work. New reload_config(path, handle) re-reads
config.toml, runs the full validation chain, and atomically swaps
ConfigHandle on success. Validation failure logs WARN without
swapping the handle. Summary log "SIGHUP: config reloaded from
<path>: M mailboxes, H hooks, T templates" on success.

Tests: 20 new hook_handler tests per rejection path, codec tests
per forbidden JSON field + template-only roundtrip, 3
reload_config tests (happy/malformed/validation-failure), a
two-consecutive-SIGHUPs test that nails down the loop-based signal
handler, and integration coverage for the raw-cmd direct-write +
SIGHUP path with positive "Reload:" + negative "Hint:" banner
assertions.
```

PR #116, opened 2026-04-20
[Sprint 6] Integration tests + book docs + doctor

uzyn:
## Sprint Goal
Close out the hook-templates track (final sprint) with end-to-end integration tests, placeholder-substitution fuzz coverage, book-chapter refreshes, and operator visibility via `aimx doctor`.

### [S6-1] End-to-end integration test (MCP → hook fire → sandbox verify)
- [x] `tests/integration.rs::hook_templates_end_to_end_mcp_to_sandbox` runs in a tempdir with `AIMX_DATA_DIR` override
- [x] Fixture config carries a `webhook` template whose `cmd[0]` is a mock-curl shell script in the test tempdir
- [x] Spins up `aimx serve`, connects MCP, calls `hook_list_templates` then `hook_create`
- [x] Asserts substituted argv matches expectations: `cmd[0]` verbatim, unsplit `Content-Type: application/json` header, URL in the last slot
- [x] Asserts stdin is valid JSON with a `raw` key carrying the persisted `.md` payload
- [x] Asserts the daemon's structured log line carries `template=webhook run_as=aimx-hook`
- [x] Root-gated assertion (skipped on non-root CI) verifies subprocess UID via `id -u` capture
- [ ] **on_receive path** not exercised directly — the DKIM-pass trust requirement can't be satisfied by fixture email. `after_send` uses the same `run_and_log` / `spawn_sandboxed` call path; on_receive trust-gate tests already exist in Sprints 1–3. Documented inline.

### [S6-2] Placeholder substitution fuzz test
- [x] 10K-iter fuzz loop with seeded xorshift64* PRNG (reproducible across toolchains)
- [x] Input generator produces empty strings, 8 KiB strings at the cap, shell metacharacters, ASCII control bytes (NUL, `\r`, `\x01`, `\x1b`, `\x7f`), high-bit Unicode (emoji, RTL, combining marks, max scalar `U+10FFFF`), near-surrogate boundary scalars
- [x] Asserts `substituted.len() == template.len()` OR deterministic rejection via typed `SubstitutionError`
- [x] Generator balance sanity check: >1000 OKs AND >1000 rejections, so a future regression that stops exercising reject paths trips the test
- [x] Two companion directed tests: NUL / control / oversize always reject; Unicode extremes round-trip cleanly
- [x] Runs under default `cargo test` (not gated on `--release`) — hand-rolled loop, no new deps

### [S6-3] Update book chapters for hook templates
All 9 chapters in PRD §8.4 refreshed:
- [x] `book/hooks.md` — template-first lead, raw-cmd moved under "power user", new origin + run_as + placeholders sections
- [x] `book/hook-recipes.md` — new template-first preamble with the 2-step pattern
- [x] `book/mcp.md` — four new tool entries (`hook_list_templates`, `hook_create`, `hook_list`, `hook_delete`) with examples
- [x] `book/setup.md` — new Hook Templates subsection with the checkbox UI copy and `aimx-hook` user creation
- [x] `book/agent-integration.md` — new per-agent template hint section with the agent → template mapping table
- [x] `book/configuration.md` — new `[[hook_template]]` schema reference + default template catalog table
- [x] `book/cli.md` — `hooks create --template` / `--param` documented, new `hooks templates` subcommand, socket path updated
- [x] `book/troubleshooting.md` — six new entries (user missing, unknown template, missing/unknown param, untrusted blocking MCP-origin, sandbox EACCES, SIGHUP reload failed, `cmd[0]` missing)
- [x] `book/faq.md` — new Q "Why can my agent create hooks but not run arbitrary shell commands?" explaining the template safety model
- [x] Socket path updated to `/run/aimx/aimx.sock` throughout the book
- [ ] **`mdbook build`** not applicable — the book/ tree is plain markdown (no `book.toml`). Chapters render via whatever static-site tooling the aimx docs site uses.

### [S6-4] `aimx doctor` extensions for hook templates
- [x] `src/doctor.rs` gains a "Hook templates" section listing every enabled template with name, truncated description, `cmd[0]`, 24h fire count, 24h failure count
- [x] Section verifies the `aimx-hook` user exists via new `SystemOps::lookup_user_uid_gid` trait method
- [x] Checks read access on `<datadir>/inbox` and `<datadir>/sent` (warns when chown step was skipped)
- [x] Fire-count parser tolerates journalctl unavailability (OpenRC fallback path already in `SystemOps::tail_service_logs`); counts render as `-` when no log source is reachable
- [x] Unit tests with fixture log lines (well-formed, malformed, empty, mixed-origin) verify the parser handles all paths without panic
- [x] 12 new tests in total; all existing doctor tests preserved

## Technical Decisions

- **Tracing subscriber wired into `run_serve`.** Discovered during S6-1 that no subscriber was ever initialised, so the structured `tracing::info!` hook-fire line emitted by `run_and_log` never reached stderr or journalctl. Without this, S6-4's fire-count parser has nothing to parse in production. Added `tracing-subscriber 0.3` (default-features-off, `fmt` + `env-filter`) as a runtime dep and initialise it at the top of `run_serve` with a `try_init` call that's safe under integration-test re-entry. RUST_LOG respected; default level is `info`.
- **S6-1 fires `after_send`, not `on_receive`.** MCP-origin hooks can never set `dangerously_support_untrusted`, so their `on_receive` gate requires `trusted == "true"` which requires DKIM=pass which fixture email can't produce. The `after_send` path traverses the same `run_and_log` → `spawn_sandboxed` call stack, proving the template → substitute → sandbox → fire → log pipeline end-to-end.
- **Doctor's `lookup_user_uid_gid` lives on `SystemOps`** rather than as a free function so the new tests can inject a mock without needing `aimx-hook` provisioned on the CI host.
- **Fuzz loop uses hand-rolled xorshift64*** rather than adding a proptest dep. Deterministic + seeded + no new deps; plain `cargo test` runs it every time.

## Deferred Items
- `mdbook build` is listed in the S6-3 AC but the repo has no `book.toml` and no mdbook entry point today. The book/ tree is consumed as plain markdown. If mdbook integration is a goal, it's a separate task.
- The on_receive→DKIM-pass fixture gap flagged under S6-1 is a known trust-model limitation (you need a real DKIM-signed fixture for `trusted = "true"` without a test override). A follow-up could add `AIMX_TEST_FORCE_TRUSTED=1` but that's intentionally not in scope for this sprint.

## Review Focus Areas
- `tests/integration.rs::hook_templates_end_to_end_mcp_to_sandbox` — the new e2e test. Worth checking the argv + stdin assertions and the root-gated UID branch.
- `src/doctor.rs::parse_hook_fire_counts` — lenient parser for structured log lines; edge-case heavy.
- `src/serve.rs::init_tracing_subscriber` — small but load-bearing. Makes sure nothing depends on a subscriber NOT being installed.
- `src/hook_substitute.rs::random_param_value` — the fuzz generator. If drift makes one branch unreachable, the balance assertion will catch it.

## Status
Sprint 6 wraps the hook-templates track. After merge, the full PRD (`docs/hook-templates-prd.md`) §9 "In Scope (v1)" is delivered.


uzyn (commented):
## Sprint Goal Assessment

Sprint 6 closes out the hook-templates track cleanly. S6-1 (e2e MCP→hook→sandbox), S6-2 (10k-iter fuzz + directed rejection/unicode/8KiB tests), S6-3 (all 9 book chapters plus index/getting-started/mailboxes/faq), and S6-4 (doctor extensions + `aimx-hook` user check + fire-count parser) are all delivered. The in-flight discovery that no `tracing_subscriber` was installed in production and the subsequent fix in `run_serve` turns the structured hook-fire log line from a silent no-op into the real foundation S6-4 parses against — without this fix S6-4 would have read empty logs in production. Good catch, correctly scoped.

**[S6-1] End-to-end integration test**
- [x] New `tests/integration.rs::hook_templates_end_to_end_mcp_to_sandbox` runs in a tempdir with `AIMX_DATA_DIR` override
- [x] Fixture config with one-template `webhook`, single `alice` mailbox, trust set implicitly via `after_send`
- [x] Mock-curl shell script records argv + stdin + `id -u` to sibling files; `webhook` `cmd[0]` points at it
- [x] Assertion on substituted argv (URL last slot, header unsplit, `cmd[0]` verbatim) and JSON stdin with `raw` key
- [x] Daemon stderr captured and asserted to contain `template=webhook` + `run_as=aimx-hook`
- [x] Root-gated UID assertion (skipped on non-root CI); non-root path verifies subprocess runs as current test user
- [~] `on_receive` path not exercised directly. Same `run_and_log` → `spawn_sandboxed` call stack is hit via `after_send`, which is documented inline and in the PR description. This is a reasonable trade-off for the DKIM-pass-trust fixture gap; surfacing as a follow-up for a `v2` story (e.g. `AIMX_TEST_FORCE_TRUSTED=1` escape hatch) is the right call, not a blocker for this sprint.

**[S6-2] Placeholder substitution fuzz test**
- [x] 10,000-iter loop with seeded xorshift64* (fixed seed `0xDEAD_BEEF_CAFE_F00D`, bit-for-bit reproducible)
- [x] Input generator produces all classes flagged in AC: empty, 8 KiB at cap, shell metachars, ASCII control (NUL/0x01/0x07/0x0D/0x1B/0x7F), high-bit Unicode (emoji/RTL/combining/PUA/max scalar `U+10FFFF`), near-surrogate boundary scalars
- [x] Invariant: `argv.len() == template.len()` on OK OR typed `SubstitutionError::{ParamContainsNul, ParamContainsControl, ParamTooLong}` on reject; any other outcome panics
- [x] Balance check asserts >1000 OK AND >1000 reject — a regression that stops hitting one path trips the test
- [x] Two companion directed tests pin deterministic rejection (NUL positions, oversize, ctrl including `\r`) and Unicode round-trip
- [x] Runs under default `cargo test`, no new deps

**[S6-3] Update book chapters for hook templates**
- [x] `book/hooks.md` — template-first lead, raw-cmd relabeled "power user", new sections for origin/run_as/stdin/placeholders
- [x] `book/hook-recipes.md` — template-first preamble with the 2-step pattern
- [x] `book/mcp.md` — four new tool entries (`hook_list_templates`, `hook_create`, `hook_list`, `hook_delete`) with full request/response shapes
- [x] `book/setup.md` — Hook Templates subsection with verbatim checkbox copy + `aimx-hook` user creation
- [x] `book/agent-integration.md` — per-agent hint section with agent → template mapping table
- [x] `book/configuration.md` — `[[hook_template]]` schema table + default catalog table
- [x] `book/cli.md` — `hooks create --template` / `--param` documented, `hooks templates` subcommand, socket path updated
- [x] `book/troubleshooting.md` — six new entries (user missing, unknown/missing/unknown-param, untrusted blocking MCP-origin, EACCES, SIGHUP reload failed, `cmd[0]` missing)
- [x] `book/faq.md` — new "Why can my agent create hooks but not run arbitrary shell" Q&A explaining the template safety model concretely
- [x] Socket path `send.sock` → `aimx.sock` updated in `index.md`, `getting-started.md`, `mailboxes.md`, `faq.md` too — no stale references left in the book
- [~] `mdbook build` AC is unsatisfiable on this repo (no `book.toml`). Correctly called out in the PR description as not-applicable; the book is plain markdown rendered by whatever static-site generator the aimx docs site uses. I re-ran `grep -r "send.sock" book/` — clean.

**[S6-4] `aimx doctor` extensions for hook templates**
- [x] `src/doctor.rs::HookTemplatesSection` with per-template `name`, truncated description, `cmd[0]`, 24h fire count, 24h failure count
- [x] `aimx-hook` user existence via new `SystemOps::lookup_user_uid_gid` trait method (backed by `libc::getpwnam` on unix)
- [x] `is_datadir_readable_by` checks read access on `<datadir>/{inbox,sent}` (world-read, uid-match, gid-match all covered); `false` renders as a warn with remediation hint
- [x] Fire-count parser short-circuits when `hook_templates` is empty (doesn't shell out to journalctl for nothing); `log_source = None` renders counts as `-` to distinguish "no fires" from "logs unavailable"
- [x] 12 new unit tests: well-formed / malformed / empty log parsing; empty-template short-circuit; count propagation; unknown-count rendering; missing-user warn; resolved-user success; cmd-path MISSING marker; table render

## Test Coverage

Strong. 1016 unit tests + 95 integration tests all pass. The new tests cover:

- The full MCP round-trip (not just mocked layers — the daemon spawns a real subprocess and records its argv/stdin)
- Deterministic rejection per error variant (NUL, control char classes, oversize) in the directed tests
- Generator balance so a future refactor that quietly stops hitting one path causes a loud failure
- Every new doctor struct field has at least one test that asserts on it

No weak/always-passing tests spotted. One minor note: the `parse_hook_fire_counts_tolerates_malformed_lines_without_panic` test comments that it's pinning the "no panic" contract rather than a strict line-count, which is the correct choice for a lenient parser.

## Potential Bugs

**1. `parse_hook_fire_counts` has a theoretical false-positive on child stderr that embeds `template=` + `exit_code=`.** (Non-blocker.) The parser's pre-filter is `line.contains("template=") && line.contains("exit_code=")`, and `extract_kv` greps the first `template=` occurrence. On the real `run_and_log` line format, `template=<tag>` always precedes `stderr_tail=...`, so the genuine value wins even if stderr contains `template=foo`. That said, a WARN line emitted *separately* for a failed exec (`"hook_name=X exec error (kind): msg"`) has no `template=` token and gets skipped correctly, and the `hook exited with code` warn line also lacks `exit_code=` so it's filtered. Safe in practice, but a `line.contains("aimx::hook")` pre-filter would make it robust against log-format drift. Flag as a Non-blocker polish item — the current parser is bounded-correct for the current log format.

**2. Fuzz generator variant 10 is a pure OK-path producer.** (Not a bug, observation only.) Scalars `\u{D7FE}`, `\u{D7FF}`, `\u{E000}`, `\u{E001}` are all valid printable scalars. They'll always land in the OK branch. That's fine — it just means this variant inflates `ok_count` rather than contributing to `reject_count` balance. The existing variants 2 (control) and 5 (oversize) hit rejections often enough that the `>1000 reject` threshold passes.

**3. `gather_status_does_not_tail_service_logs` semantic change.** (Not a bug — expected.) The old test asserted "doctor never tails logs." The new behavior tails logs when templates are enabled (to compute fire counts), so the test was renamed to scope the assertion to the no-templates case. Correct per S6-4 AC.

No Blocker-level bugs found.

## Security Issues

None. The template-first safety model is the whole point of the track and this sprint doesn't touch the substitution/sandbox/UDS-verb layer. The e2e test proves the value-substitution guarantees hold end-to-end (no argv split, URL in last slot, header unsplit).

## Code Quality

Clean. Notes:

- `HookUserStatus::user` field always contains `"aimx-hook"` (copied from the `HOOK_SERVICE_USER` constant). It's carried for rendering so the template path doesn't need to import the constant, which is fine. Minor redundancy, not worth changing.
- The ASCII-table renderer in `render_hook_templates_table` uses hand-rolled frame/pad/cell logic that mirrors the existing mailbox table. Pattern is consistent with the rest of the file.
- `extract_kv` stops at the first ASCII whitespace, which is the right behavior for the current unquoted scalar format. If `stderr_tail` ever inlines JSON-escaped text without the trailing-field-wins guarantee, the parser would need to be taught about quoted values. Non-blocker.
- `init_tracing_subscriber` uses `try_init` with the documented reason (integration-test re-entry), writes to stderr, respects `RUST_LOG` with `info` default, and uses `compact()` for single-line records. Load-bearing but correct.

## Alignment with PRD

- PRD §7.3 "Observability" asked for `aimx doctor` to surface enabled templates with per-template 24h fire counts from service logs — delivered, plus the more robust `None`-vs-`Some(0)` distinction.
- PRD §7.3 also asked for `aimx logs` to surface the structured hook-fire line — already delivered in earlier sprints via the shared `run_and_log` path. The tracing-subscriber wiring is what actually makes this observable in production, which was an implicit assumption the earlier sprints papered over.
- PRD §10 risks around placeholder injection and binary path drift are both addressed (fuzz test + `cmd[0] (MISSING)` marker).
- PRD §9 "In Scope (v1)" is now fully delivered after this merge.

## Human Review Comments

No human review comments on PR #116 or the linked issue.

**Overall verdict: Ready to merge**

All S6 stories meet their acceptance criteria with the two intentional scope calls-out documented in the PR description (on_receive fixture gap and `mdbook build` non-applicability). Tests are substantive and the fuzz-generator balance check is the kind of forcing function that makes this track resilient to future drift. The tracing-subscriber discovery was a genuine in-flight fix, correctly minimally-scoped, and properly idempotent. Full `cargo test` (1016 unit + 95 integration) passes; `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

**Blockers:** none.

**Non-blockers (not blocking merge; consider as cleanup):**
- Tighten `parse_hook_fire_counts` pre-filter to require the `aimx::hook` target token so a pathological child stderr with `template=X exit_code=Y` inside `stderr_tail` can't be miscounted. Current format guarantees this doesn't misfire; the fix is defense-in-depth against format drift.
- Fuzz-generator variant 10 (near-surrogate boundary scalars) only ever exercises the OK path. Mix in a high-bit variant that also carries a control byte to harden the reject-side coverage of this class.

**Nice-to-haves (track for later):**
- Add an `on_receive` e2e companion once an `AIMX_TEST_FORCE_TRUSTED=1` escape hatch exists, to remove the fixture-dependency footnote under S6-1.
- After the non-blockers in the Sprint 1–3 backlog (`AIMX_SANDBOX_FORCE_FALLBACK` / `AIMX_TEST_SKIP_ROOT_CHECK` gating) are addressed, revisit `validate_hooks` error-message ordering (Sprint 1 nit) in the same cleanup sweep.

```
[Sprint 6] Integration tests + book docs + doctor

Closes out the hook-templates track:

- tests/integration.rs: end-to-end MCP → UDS → hook fire → sandbox
  verify using a mock-curl shell script. Asserts substituted argv
  (unsplit header, URL in last slot, cmd[0] verbatim), email_json
  stdin shape, structured hook-fire log fields (template=webhook,
  run_as=aimx-hook), and a root-gated subprocess UID drop assertion.
- src/hook_substitute.rs: 10K-iter seeded-xorshift64* fuzz loop with
  a generator covering NUL/control/oversize/shell-metachar/high-bit-
  Unicode/near-surrogate classes, plus directed tests for
  deterministic rejection and Unicode round-trip. Balance check
  (>1000 OK, >1000 reject) catches generator drift.
- src/doctor.rs: Hook templates section listing enabled templates
  with 24h fire + failure counts parsed from the aimx::hook log
  line; aimx-hook user presence + uid/gid + datadir read access
  checks; `-` rendering when log source is unavailable so operators
  can distinguish "no fires" from "logs unavailable".
- src/serve.rs: install tracing_subscriber::fmt() in run_serve so
  the structured hook-fire log line actually lands in stderr /
  journalctl. Discovered during S6-1; without it S6-4's fire-count
  parser has nothing to parse in production.
- book/: all 9 chapters per PRD §8.4 refreshed to lead with
  templates; socket path updated to /run/aimx/aimx.sock throughout.

With this merge, PRD docs/hook-templates-prd.md §9 "In Scope (v1)"
is fully delivered.
```

PR #117, opened 2026-04-21
cleanup: hook-templates non-blocking backlog

uzyn:
## Summary

Closes 5 of the 7 open nits in the hook-templates sprint review backlog now that the track is complete. Items 2 (`validate_hooks` error-message ordering) and 8 (true E2E SIGHUP sub-process test) skipped as low-value-per-effort with no concrete user signal.

### What changed

- **Item 3** — `platform::spawn_via_systemd_run` preflight no longer emits its own WARN when it delegates to `spawn_via_fork_setuid`. The fallback's WARN is now the single source of the unknown-user signal.
- **Items 4 + 7** — `AIMX_SANDBOX_FORCE_FALLBACK` (`src/platform.rs`) and `AIMX_TEST_SKIP_ROOT_CHECK` (`src/hooks.rs`, two sites) now gated behind `cfg(any(test, debug_assertions))`. Release binaries ignore both switches entirely. Extracts a `skip_root_check()` helper in `hooks.rs` to avoid duplicating the gate.
- **Item 5** — `aimx doctor` parses the 24h log tail for the `run_as '...' not found and caller is non-root` WARN and renders a `Sandbox:` warn line with a `sudo aimx setup` remediation hint when the count is non-zero. Log scan now also runs when an `aimx-hook` user exists even if no templates are configured, so a misconfigured box still surfaces the signal.
- **Item 6** — MCP `hook_delete` tool description and `book/mcp.md` explicitly spell out the `origin`-as-submission-channel-marker semantics and call out the CLI-template case. PRD intentionally left untouched to avoid post-launch churn (book is the operator-facing surface).

### What's skipped (and why)

- **Item 2** (Sprint 1 — `validate_hooks` error-message ordering for `dangerously_support_untrusted` on MCP-origin `after_send` hooks): both rejections are correct — only the text differs. Cosmetic with no operator-visible confusion observed.
- **Item 8** (Sprint 3 — true end-to-end sub-process SIGHUP test): the inline-replicated-branch test is functionally equivalent for "loop correctly re-arms". The backlog entry was conditional on Sprint 5/6 adding ordering between signal recv and reload, which never happened.

### Verification

- `cargo test`: 1019 unit + 95 integration tests pass (3 pre-existing ignored).
- Release-vs-debug binary check confirms `AIMX_TEST_SKIP_ROOT_CHECK=1` is inert in release (`-- cmd hooks require root`) and honored in debug (`Hook created`).

## Test plan

- [ ] CI green (cargo test, cargo clippy, cargo fmt check on both crates)
- [ ] Manual: `aimx doctor` on a live box without an `aimx-hook` user shows the `Sandbox:` warn row after the daemon has logged at least one fallback event
- [ ] Manual: release binary ignores both `AIMX_SANDBOX_FORCE_FALLBACK` and `AIMX_TEST_SKIP_ROOT_CHECK`


uzyn (commented):
## Standalone review — PR #117

### Changes Overview

Closes 5 of 7 non-blocking backlog items from the hook-templates sprint:

1. `platform::spawn_via_systemd_run` no longer emits a preflight WARN when it delegates to the fork-setuid fallback on an unresolvable `run_as`. The fallback emits the single canonical WARN.
2. `AIMX_SANDBOX_FORCE_FALLBACK` (in `platform.rs`) and `AIMX_TEST_SKIP_ROOT_CHECK` (in `hooks.rs`, two sites) are now gated behind `cfg(any(test, debug_assertions))` — release binaries ignore both. A `skip_root_check()` helper removes duplication.
3. `aimx doctor` parses the 24h log tail for the fallback WARN and renders a `Sandbox:` warn row with a `sudo aimx setup` hint when the count is non-zero. The log scan now also runs when an `aimx-hook` user exists even without templates configured.
4. MCP `hook_delete` tool description and `book/mcp.md` spell out the `origin`-as-submission-channel semantics and the CLI-template case.

Items 2 and 8 are intentionally left unchecked.

### Scope Alignment

Scope matches the PR description exactly — no scope creep, no scope gaps. Items checked off in `docs/hook-templates-sprint.md` each carry a `_Resolved in post-track cleanup: ..._` annotation naming the concrete change, which makes the bookkeeping auditable.

### Potential Bugs

**Non-blocker — `parse_unknown_user_fallback_count` uses a broad substring match** (`src/doctor.rs:287-295`).

The parser matches `line.contains("not found and caller is non-root")`, which will false-positive on any log line echoing that phrase via `stderr_tail="..."`. The structured hook-fire log line captures up to 64 KiB of a subprocess's stderr, so a hook whose binary happened to print "file not found and caller is non-root" (or similar) would register as a sandbox fallback event in `aimx doctor`.

The probability in practice is very low — the phrase is unusual — and the worst case is a misleading warn row rather than a crash, silent data loss, or security issue. But tightening the match to the full canonical prefix (e.g. `run_as '` + `' not found and caller is non-root:`) would remove the ambiguity for ~zero extra code.

Classified non-blocker: no crash, no correctness impact on the hot path, and the warn row is still informative enough for the operator to reach the logs.

### Security Issues

None. The change that touches security posture (gating the test-only env vars out of release) *tightens* behavior — release builds can no longer be coerced into skipping the root check or downgrading the sandbox. Verified the gating by running the release vs debug binary with `AIMX_TEST_SKIP_ROOT_CHECK=1`: release rejects with `--cmd hooks require root`, debug creates the hook.

### Test Coverage

Adequate. New tests:

- `parse_unknown_user_fallback_count_matches_production_warn_line` and `parse_unknown_user_fallback_count_returns_zero_for_empty_or_unrelated` cover the parser directly.
- `gather_hook_templates_section_counts_unknown_user_fallback_when_logs_present` and `gather_hook_templates_section_tails_logs_when_only_user_exists` cover the gather path, including the new "user-exists-but-no-templates triggers log tail" behavior.
- `format_hook_templates_section_renders_sandbox_warn_on_non_zero_fallback` and `format_hook_templates_section_suppresses_sandbox_warn_on_zero_fallback` cover the renderer.

**Nice-to-have:** no explicit test asserting that the `Sandbox:` warn is suppressed when `unknown_user_fallback_24h == None` (log source missing). The `None` path is exercised indirectly by `format_hook_templates_section_dashes_unknown_counts_when_log_source_missing`, which doesn't assert on the absence of `Sandbox:`. Tightening the existing test or adding a dedicated case would lock down the `if let Some(n) = ... && n > 0` guard against future refactors.

**Nice-to-have:** the PR's release-vs-debug behavioral verification for the env-var gating was manual only. Hard to express as a cargo test without spawning subprocess release builds, so not a real gap — just worth noting.

### Code Quality

- `skip_root_check()` helper in `src/hooks.rs` is the right extraction. Two-site duplication would have been fine too, but the helper also hosts the explanatory comment next to the gating, which is a nice bonus.
- The `if let Some(n) = section.unknown_user_fallback_24h && n > 0` let-chain is edition-2024 appropriate and reads cleanly.
- `docs/hook-templates-sprint.md` leaves items 2 and 8 unchecked. A future scrum-plan-updater or human reader might re-attempt them. **Nice-to-have:** add a one-liner note like `_Skipped intentionally in PR #117 — low value relative to effort._` so the decision sticks.

**Overall verdict: Ready to merge.** No blockers. The single non-blocker (broad substring match in the fallback-event parser) is worth addressing in a follow-up but does not gate merge.

**Blockers:** none.

**Non-blockers:**
- Tighten `parse_unknown_user_fallback_count` substring match to the full canonical prefix to eliminate stderr_tail echo false-positives (`src/doctor.rs:287-295`).

**Nice-to-haves:**
- Add an explicit assertion that the `Sandbox:` warn is suppressed when `unknown_user_fallback_24h == None`.
- Annotate the skipped items 2 and 8 in `docs/hook-templates-sprint.md` with a pointer to PR #117 so future readers don't re-attempt them.

```
cleanup: hook-templates non-blocking backlog (#117)

Closes 5 of 7 open nits from the hook-templates sprint review backlog now that the track is complete.

- Silence the systemd-run preflight WARN on unknown-user fallback; the fork-setuid fallback already emits a single WARN for the event.
- Gate AIMX_SANDBOX_FORCE_FALLBACK and AIMX_TEST_SKIP_ROOT_CHECK behind cfg(any(test, debug_assertions)) so release binaries ignore both switches entirely. Shared skip_root_check() helper in hooks.rs.
- aimx doctor now counts the unknown-user fallback WARN in the 24h log tail and surfaces a Sandbox: warn row with a `sudo aimx setup` hint when non-zero. Log scan also runs when an aimx-hook user exists even without configured templates.
- hook_delete MCP tool description and book/mcp.md now explicitly spell out origin-as-submission-channel semantics and call out the CLI-template case.

Items 2 (validate_hooks message ordering) and 8 (true E2E SIGHUP sub-process test) intentionally skipped as low-value-per-effort.
```

PR #122, opened 2026-04-21
fix(smtp): reject RCPT TO for unrelated domains

uzyn:
## Summary

Close the open-relay-adjacent ingestion vector by rejecting SMTP `RCPT TO` recipients whose domain is not `config.domain`. Previously, any address at `RCPT TO` was accepted and fell through to `catchall` for unknown local parts — so `anything@totally-unrelated-domain.com` was happily stored in `inbox/catchall/`. Now, if the recipient's domain is not the configured domain (case-insensitive, exact match), the server replies `550 5.7.1 relay not permitted` and does not enqueue the recipient.

## Requirements

- [x] RCPT TO with a recipient whose domain matches `config.domain` case-insensitively is accepted with the existing `250 OK`.
- [x] RCPT TO with a recipient whose domain does not exactly match `config.domain` (including unrelated domains and any subdomain) is rejected with `550 5.7.1 relay not permitted\r\n` and not added to `session_state.forward_paths`.
- [x] RCPT TO with an address that has no `@` is rejected with the same `550 5.7.1 relay not permitted\r\n`.
- [x] Empty-angle-addr `RCPT TO:<>` continues to return `501 Syntax: RCPT TO:<address>` (existing behavior, unchanged).
- [x] Multi-recipient transactions: a rejected recipient does not block subsequent RCPT TOs, and a transaction with at least one accepted recipient still proceeds to DATA.
- [x] A session whose only RCPT TOs were all rejected continues to refuse DATA with `503 Send RCPT TO first` (state never advances to RcptTo).
- [x] Rejected RCPT TO is logged to stderr with peer address, attempted recipient, and configured domain.
- [x] Existing SMTP integration tests and `tests/integration.rs` pass unchanged.
- [x] `book/mailboxes.md` explicitly documents that RCPT TO for other domains is rejected at SMTP time.

## Out of scope

- Submission-port (587/465) authentication — aimx has no submission port; outbound goes through the UDS signing boundary.
- SPF / DKIM / DMARC policy — those still gate hook firing, not storage.
- Accepting subdomains of `config.domain` — deliberately rejected (one-domain-per-instance layout).
- Configurable reject-code or per-operator allowlists of extra domains.
- Changes to `resolve_mailbox()` / `extract_local_part()` — the domain check runs earlier, at RCPT TO.
- Rate-limiting, blocklists, greylisting.
- IDN normalization beyond ASCII lowercasing — `config.domain` is stored as a plain string and compared case-insensitively.

## Technical approach

- Added `recipient_domain_matches(addr, configured_domain)` helper in `src/smtp/session.rs`. Uses `rsplit_once('@')` so quoted local parts split on the final `@`, and `eq_ignore_ascii_case` per RFC 5321 §2.3.11. Returns `false` for addresses with no `@` or an empty domain part.
- Wired into `SmtpSession::handle_rcpt_to` immediately after `extract_angle_addr` returns a non-empty address and before `session_state.forward_paths.push`. Uses `self.params.config_handle.load().domain` so hot-swapped configs (MAILBOX-* between EHLO and DATA) would be honoured.
- Rejection path: `eprintln!` the peer / recipient / configured_domain, return `550 5.7.1 relay not permitted\r\n`. State is untouched so subsequent RCPT TOs still work and DATA with zero accepted recipients still 503s (via the existing `handle_data_precheck`).
- Doc update in `book/mailboxes.md` under "Routing logic" — added an explicit sentence plus two example bullets for the reject cases (other domain, subdomain).

## Test plan

Unit tests added in `src/smtp/session.rs::unit_tests` (all pass):
- `recipient_domain_matches_exact`
- `recipient_domain_matches_case_insensitive`
- `recipient_domain_matches_rejects_different_domain`
- `recipient_domain_matches_rejects_subdomain`
- `recipient_domain_matches_rejects_parent_domain`
- `recipient_domain_matches_rejects_missing_at`
- `recipient_domain_matches_rejects_empty_domain`
- `recipient_domain_matches_quoted_local_part`

SMTP integration tests added in `src/smtp/tests.rs` (all pass):
- `test_rcpt_rejects_unrelated_domain` — `user@evil.example` rejected with `550 5.7.1`; DATA returns 503; no file in `inbox/catchall/`.
- `test_rcpt_rejects_subdomain` — `user@sub.test.local` rejected.
- `test_rcpt_rejects_missing_at` — `<alice>` rejected with `550 5.7.1` (policy reject, not 501).
- `test_rcpt_mixed_valid_invalid` — alice accepted, mallory@evil.example rejected, DATA proceeds; one `.md` in `inbox/alice/`, zero in `inbox/catchall/`.
- `test_rcpt_accepts_case_insensitive_domain` — `alice@TEST.LOCAL` accepted; file lands in `inbox/alice/`.

Results:
- `cargo test`: 1026 unit/bin tests pass; 95 integration tests pass (one pre-existing flake unrelated to this change — `after_send_hook_fires_with_delivered_status` intermittently times out waiting for a UDS socket under heavy parallel load; passes in isolation and on a second run).

## Notes

- The check runs against `config_handle.load().domain` per RCPT TO, so any future hot-swap that changes `domain` mid-session would be honoured on the next RCPT line.
- No new dependencies; logic is pure ASCII case-insensitive string compare.

### DIFF
--- a/book/mailboxes.md
+++ b/book/mailboxes.md
@@ -32,10 +32,14 @@ as a sibling file when attachments are present.
 
 When an email arrives, aimx matches the local part of the recipient address (the part before `@`) against mailbox names in the config. If a mailbox with that exact name exists, the email is delivered there. Otherwise it falls through to the `catchall` mailbox.
 
+RCPT TO addresses whose domain is not the configured `domain` (case-insensitive exact match) are rejected at SMTP time with `550 5.7.1 relay not permitted` and never reach storage. aimx is not an open relay: `catchall` only covers unrecognized local parts *at your configured domain*, not unrelated domains or subdomains.
+
 For example, with mailboxes `support` and `catchall` configured:
 - `support@agent.yourdomain.com` -> delivered to the `support` mailbox
 - `billing@agent.yourdomain.com` -> delivered to the `catchall` mailbox (no `billing` mailbox exists)
 - `anything@agent.yourdomain.com` -> delivered to the `catchall` mailbox
+- `anything@some-other-domain.com` -> rejected at RCPT TO with `550 5.7.1 relay not permitted`
+- `anything@sub.agent.yourdomain.com` -> rejected at RCPT TO with `550 5.7.1 relay not permitted`
 
 ## Managing mailboxes
 
--- a/src/smtp/session.rs
+++ b/src/smtp/session.rs
@@ -373,6 +373,14 @@ impl SmtpSession {
         if addr.is_empty() {
             return "501 Syntax: RCPT TO:<address>\r\n".to_string();
         }
+        let configured_domain = self.params.config_handle.load().domain.clone();
+        if !recipient_domain_matches(&addr, &configured_domain) {
+            eprintln!(
+                "[{}] RCPT rejected (relay): recipient={} configured_domain={}",
+                self.params.peer_addr, addr, configured_domain
+            );
+            return "550 5.7.1 relay not permitted\r\n".to_string();
+        }
         session_state.forward_paths.push(addr);
         session_state.state = State::RcptTo;
         "250 OK\r\n".to_string()
@@ -649,6 +657,16 @@ fn extract_angle_addr(s: &str) -> String {
     s.to_string()
 }
 
+fn recipient_domain_matches(addr: &str, configured_domain: &str) -> bool {
+    let Some((_, domain)) = addr.rsplit_once('@') else {
+        return false;
+    };
+    if domain.is_empty() {
+        return false;
+    }
+    domain.eq_ignore_ascii_case(configured_domain)
+}
+
 #[cfg(test)]
 mod unit_tests {
     use super::*;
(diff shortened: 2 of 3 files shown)


uzyn (commented):
## Standalone Review — PR #122 `fix(smtp): reject RCPT TO for unrelated domains`

### Changes Overview

Adds a domain check to `SmtpSession::handle_rcpt_to` that rejects any `RCPT TO` whose recipient domain does not exactly match `config.domain` (case-insensitive, ASCII). The check lives in a new pure helper `recipient_domain_matches` in `src/smtp/session.rs`, is wired in immediately after `extract_angle_addr` and before the forward-path push, returns `550 5.7.1 relay not permitted` on failure, logs to stderr, and leaves `session_state.state` untouched so all-rejected transactions still 503 at DATA. Ships 8 unit tests + 5 SMTP integration tests and an explicit note in `book/mailboxes.md`.

### Scope Alignment

Every requirement in the PR description maps to code and a test. No scope creep; the change is limited to the RCPT TO path, the new helper, the book update, and their tests. The explicit out-of-scope items (subdomains, IDN normalization, submission auth, per-operator allowlists, rate limiting) are all respected — e.g. `alice@test.local.` with a trailing dot is rejected (strict match), matching the stated intent.

### Potential Bugs

None found. I read the full helper, the call site, and the surrounding session state machine:

- `rsplit_once('@')` is the right choice — `"weird@name"@test.local` splits on the last `@` and keeps the quoted local part intact (covered by `recipient_domain_matches_quoted_local_part`).
- `eq_ignore_ascii_case` is the RFC 5321 §2.3.11 compare for domains. IDN is explicitly out of scope.
- `addr@extra@evil.com`-style bypass attempts are handled correctly: the final `@` wins, so `alice@test.local@evil.com` rejects with `evil.com`.
- `session_state.state` is deliberately not mutated on reject, so a session whose only RCPTs were rejected correctly falls to `503 Send RCPT TO first` via `handle_data_precheck`. Verified in `test_rcpt_rejects_unrelated_domain`.
- Mixed accept/reject transactions still advance to `RcptTo` on the first accepted recipient — verified in `test_rcpt_mixed_valid_invalid`.
- Empty-domain and missing-`@` inputs both return `false` from the helper before the `eq_ignore_ascii_case` call, so a hypothetical empty `config.domain` cannot accidentally admit `"alice@"`.
- `config_handle.load()` is taken fresh per RCPT, so a `MAILBOX-CREATE` / hot-swap mid-session sees the fresh `domain` on the next RCPT line, which matches the description's claim.

### Security Issues

None. The change closes the open-relay-adjacent vector described in the PR. No new attack surface is introduced:

- `eprintln!` matches the existing session logging pattern; volume is bounded by SMTP session limits already enforced elsewhere.
- No user input is reflected into the response past the fixed `550 5.7.1 relay not permitted` string.

### Test Coverage

Strong and well-targeted:

- Unit tests exercise the pure helper across exact match, case-folding on both sides, different-domain, subdomain, parent-domain, missing-`@`, empty-domain, and quoted-local-part cases.
- SMTP integration tests exercise the wire-level behaviour end-to-end — unrelated domain, subdomain, missing-`@`, mixed accept/reject with subsequent DATA, and case-insensitive accept — and verify both the SMTP response codes and on-disk effects (no stray files in `catchall` for rejected RCPTs; accepted RCPTs land in the expected mailbox).
- The existing SMTP test suite (45 tests) still passes, so the behavioural change for `<alice>` (no `@`) moving from accepted-then-catchall to 550 at RCPT time doesn't regress any previously codified contract.

I ran locally and confirmed: `cargo fmt -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, all 8 unit tests and 5 new SMTP tests pass, and the full 45-test SMTP suite passes.

### Code Quality

Nothing to flag. The helper is 8 lines, single-purpose, and sits next to `extract_angle_addr` which is the natural neighbour. Naming is clear. The call site is a three-line addition in the obvious place.

**Overall verdict: Ready to merge**

- No Blockers.
- No Non-blockers.
- No Nice-to-haves — the PR already scopes out the obvious extensions (allowlists, subdomain acceptance, IDN, rate-limiting) cleanly.

```
fix(smtp): reject RCPT TO for unrelated domains

Close the open-relay-adjacent ingestion path by rejecting any RCPT TO
whose recipient domain does not exactly match `config.domain`
(case-insensitive, ASCII). Previously, `anything@totally-unrelated.com`
fell through to `catchall` and was stored locally. The SMTP server now
replies `550 5.7.1 relay not permitted` for unrelated domains,
subdomains of the configured domain, and addresses missing an `@`,
without advancing transaction state — so all-rejected sessions still
503 at DATA, and mixed-recipient transactions still deliver accepted
RCPTs. Adds a `recipient_domain_matches` helper, unit + integration
tests, and a note in `book/mailboxes.md`.
```

PR #125, opened 2026-04-21
[Sprint 1] Config schema + invariants

uzyn:
Implements [Sprint 1 of the agent-integration track](../docs/agent-integration-sprint.md#sprint-1--config-schema--invariants-days-125-in-progress). Pure schema + validation — no disk-ownership changes (those land in Sprint 2).

## Stories landed

**S1-1 MailboxConfig.owner** — required `owner: String` field (no `#[serde(default)]`); `owner_uid` / `owner_gid` helpers via `getpwnam`; missing-owner load failure with the actionable message specified by the AC; TOML round-trip preserves `owner` exactly.

**S1-2 Retire VALID_HOOK_RUN_AS_VALUES** — allowlist + `default_hook_run_as` gone. New `validate_run_as(&str) -> Result<RunAsKind, ConfigError>` helper: reserved names short-circuit (`root`, `aimx-catchall`), regex gate `[a-z_][a-z0-9_-]*[$]?`, then `getpwnam`. Regex failures are hard (`InvalidUsername`), lookup misses are soft (`OrphanUser`). A thin `user_resolver` module wraps `libc::getpwnam` with a trait seam so tests can inject a deterministic resolver — the `runner` / `ubuntu` user on CI is never in the way.

**S1-3 check_hook_owner_invariant** — shared helper enforced at both `Config::load` (via `validate_hooks`) and UDS `HOOK-CREATE` (via the existing `validate_single_hook` path in `hook_handler`). Rule: `hook.run_as == mailbox.owner OR "root"`; catchall relaxes to `"aimx-catchall" OR "root"`. Error messages name the offending mailbox, hook, actual `run_as`, expected owner, and suggested fix — as specified by the AC.

**S1-4 Orphan tolerance** — `Config::load` now returns `Result<(Config, Vec<LoadWarning>), _>`. `LoadWarning` enum surfaces `OrphanMailboxOwner` / `OrphanHookRunAs` / `OrphanTemplateRunAs` / `LegacyAimxHookUser`. Orphan mailboxes / templates are retained in the parsed `Config` but flagged inactive via a `ConfigResolved` side-table. Startup warnings are logged under `aimx::config`; SIGHUP reload logs them again on every refresh.

## Deferred / not in scope for Sprint 1

- Ingest / MCP / UDS paths do not yet consult `ConfigResolved::is_mailbox_active` — those wire-ups are Sprint 2 + 4. The helper exists and is test-covered; callers annotate `#[allow(dead_code)]` until they're wired.
- `owner_uid` / `owner_gid` are exposed but unused in production — Sprint 2's chown paths consume them.
- Disk ownership, setup UX, UDS SO_PEERCRED, agent-setup rewrite: Sprints 2 / 3 / 4 / 5 / 6.

## Test strategy

- New unit tests on every AC in `src/config.rs`:
  - Mailbox owner: accept / reject missing / orphan warns / TOML round-trip
  - `validate_run_as`: reserved accept, regex failures (uppercase, unicode, leading digit), `getpwnam` success + orphan via mocked resolver
  - Invariant: matching owner passes, mismatch fails with the right message, root always passes (catchall + non-catchall), catchall accepts `aimx-catchall`, non-catchall rejects `aimx-catchall`, load-time enforcement
  - Orphan tolerance: valid config → empty warnings, single orphan owner → one warning + inactive flag, orphan template → one warning + inactive flag, mixed valid + orphan → only orphans surface
- Existing test fixtures updated: `run_as = \"aimx-hook\"` → `run_as = \"root\"` (invariant-safe on every CI host); every mailbox TOML fixture gains `owner = \"root\"` / `owner = \"aimx-catchall\"`
- `user_resolver` test override uses a two-layer `RwLock + Mutex` scheme: read-only callers never block, and parallel tests that install overrides serialize through the mutex — no deadlocks under `--test-threads=1` or default parallel runs

## Verification

- `cargo test` — 1063 unit + 96 integration passing
- `services/verifier` crate independently built + tested — clean

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -403,6 +403,16 @@ pub enum LoadWarning {
     /// The template is retained but flagged orphan; Sprint 6's doctor
     /// work surfaces it.
     OrphanTemplateRunAs { template: String, run_as: String },
+    /// The hook/mailbox owner invariant (PRD §6.3) was skipped at load
+    /// because either the mailbox owner or the hook's effective `run_as`
+    /// is orphan-flagged. Per PRD §6.1 the daemon stays up — the hook is
+    /// unfireable until the user reappears, at which point the next SIGHUP
+    /// / restart re-runs the invariant check.
+    HookInvariantSkippedDueToOrphan {
+        mailbox: String,
+        hook_name: String,
+        reason: String,
+    },
     /// A stale `aimx-hook` system user is present on the host. aimx no
     /// longer creates this user; doctor notes its presence so the
     /// operator can clean up.
@@ -430,6 +440,15 @@ impl LoadWarning {
             LoadWarning::OrphanTemplateRunAs { template, run_as } => format!(
                 "hook template '{template}' has run_as='{run_as}' which does \
                  not resolve; template marked orphan"
+            ),
+            LoadWarning::HookInvariantSkippedDueToOrphan {
+                mailbox,
+                hook_name,
+                reason,
+            } => format!(
+                "hook '{hook_name}' on mailbox '{mailbox}': owner/run_as \
+                 invariant skipped — {reason}; hook will be soft-skipped \
+                 on fire until the user reappears (PRD §6.1)"
             ),
             LoadWarning::LegacyAimxHookUser => {
                 "legacy 'aimx-hook' system user present; aimx no longer \
@@ -517,13 +536,23 @@ pub fn check_hook_owner_invariant(
     }
 
     let hook_label = hook.name.clone().unwrap_or_else(|| "<anonymous>".into());
+    // When the mailbox owner is already `root` the fallback hint
+    // (`or run_as='root'`) would duplicate the primary suggestion, so
+    // drop the `or` clause for that one case.
+    let fix_suggestion = if mailbox.owner == RESERVED_RUN_AS_ROOT {
+        format!("set run_as='{RESERVED_RUN_AS_ROOT}'")
+    } else {
+        format!(
+            "set run_as='{owner}' or run_as='{root}'",
+            owner = mailbox.owner,
+            root = RESERVED_RUN_AS_ROOT,
+        )
+    };
     Err(format!(
         "hook '{hook_label}' on mailbox '{mailbox_name}' has run_as='{run_as}' \
-         but the mailbox is owned by '{owner}'; fix: set run_as='{owner}' \
-         or run_as='{root}' so the hook can read this mailbox's files \
-         (PRD §6.3)",
+         but the mailbox is owned by '{owner}'; fix: {fix_suggestion} so the \
+         hook can read this mailbox's files (PRD §6.3)",
         owner = mailbox.owner,
-        root = RESERVED_RUN_AS_ROOT,
     ))
 }
 
@@ -601,7 +630,66 @@ fn reject_legacy_on_receive_schema(toml_text: &str) -> Result<(), Box<dyn std::e
     Ok(())
 }
 
-pub(crate) fn validate_hooks(config: &Config) -> Result<(), String> {
+/// Context passed to [`validate_hooks`] describing which mailbox
+/// owners / template `run_as` names were already flagged orphan by the
+/// surrounding load-time validators. `validate_hooks` uses the sets to
+/// downgrade the hook/owner invariant check (PRD §6.3) to a warning
+/// when either side is unresolvable — PRD §6.1's orphan tolerance
+/// requires that a user deletion cannot kill the daemon.
+///
+/// Callers that create fresh hooks (UDS `HOOK-CREATE`, `aimx hooks
+/// create`) pass `OrphanSkipContext::strict()`: at create time we do
+/// not want to silently accept a hook whose `run_as` can't be resolved.
+#[derive(Debug, Default)]
+pub(crate) struct OrphanSkipContext {
+    /// Mailboxes whose `owner` did not resolve at load time.
+    pub orphan_mailbox_owners: std::collections::HashSet<String>,
+    /// Hook templates whose `run_as` did not resolve at load time.
+    pub orphan_templates: std::collections::HashSet<String>,
+    /// Per-mailbox hook names whose explicit `run_as` did not resolve.
+    /// Keyed by `(mailbox, hook_effective_name)` so we only skip the
+    /// specific hook that was flagged, not every hook on the mailbox.
+    pub orphan_hook_run_as: std::collections::HashSet<(String, String)>,
+}
+
+impl OrphanSkipContext {
+    /// Strict context: no orphan skipping. Used by UDS / CLI create
+    /// paths where the operator is actively introducing a new hook.
+    pub(crate) fn strict() -> Self {
+        Self::default()
+    }
+
+    /// Build a context from an in-progress warning vector. Called by
+    /// [`Config::load`] after the owner / template / hook-run_as
+    /// validators have run so the invariant pass can see which names
+    /// are already known orphan.
+    pub(crate) fn from_warnings(warnings: &[LoadWarning]) -> Self {
+        let mut ctx = Self::default();
+        for w in warnings {
+            match w {
+                LoadWarning::OrphanMailboxOwner { mailbox, .. } => {
+                    ctx.orphan_mailbox_owners.insert(mailbox.clone());
+                }
+                LoadWarning::OrphanTemplateRunAs { template, .. } => {
+                    ctx.orphan_templates.insert(template.clone());
+                }
+                LoadWarning::OrphanHookRunAs {
+                    mailbox, hook_name, ..
+                } => {
+                    ctx.orphan_hook_run_as
+                        .insert((mailbox.clone(), hook_name.clone()));
+                }
+                _ => {}
+            }
+        }
+        ctx
+    }
+}
+
+pub(crate) fn validate_hooks(
+    config: &Config,
+    orphan_ctx: &OrphanSkipContext,
+) -> Result<Vec<LoadWarning>, String> {
     // Effective-name map: name -> (mailbox, is_explicit).
     let mut seen: HashMap<String, (String, bool)> = HashMap::new();
     let template_names: std::collections::HashSet<&str> = config
@@ -609,6 +697,7 @@ pub(crate) fn validate_hooks(config: &Config) -> Result<(), String> {
         .iter()
         .map(|t| t.name.as_str())
         .collect();
+    let mut warnings: Vec<LoadWarning> = Vec::new();
 
     for (mailbox_name, mb) in &config.mailboxes {
         for hook in &mb.hooks {
@@ -684,9 +773,43 @@ pub(crate) fn validate_hooks(config: &Config) -> Result<(), String> {
             // Hook/mailbox owner invariant (PRD §6.3). Runs after the
             // template-reference check so we know the template exists
             // (when referenced) before resolving the effective run_as.
-            check_hook_owner_invariant(config, mailbox_name, mb, hook)?;
+            //
+            // PRD §6.1 orphan tolerance: if either the mailbox owner or
+            // the effective run_as is orphan-flagged by the surrounding
+            // validators, downgrade the invariant mismatch to a warning
+            // — the hook is unfireable anyway and the daemon must stay
+            // up on a user-deletion event.
+            let hook_effective_name = effective_hook_name(hook);
+            let owner_is_orphan = orphan_ctx.orphan_mailbox_owners.contains(mailbox_name);
+            let hook_run_as_is_orphan = orphan_ctx
+                .orphan_hook_run_as
+                .contains(&(mailbox_name.clone(), hook_effective_name.clone()));
+            let template_is_orphan = hook
+                .template
+                .as_ref()
+                .is_some_and(|t| orphan_ctx.orphan_templates.contains(t));
+            if owner_is_orphan || hook_run_as_is_orphan || template_is_orphan {
+                if let Err(_reason) = check_hook_owner_invariant(config, mailbox_name, mb, hook) {
+                    let skip_reason = if owner_is_orphan {
+                        format!("mailbox owner '{}' is orphan", mb.owner)
+                    } else if template_is_orphan {
+                        let tmpl = hook.template.as_deref().unwrap_or("<unknown>");
+                        format!("template '{tmpl}' run_as is orphan")
+                    } else {
+                        let run_as = hook.run_as.as_deref().unwrap_or("<unknown>");
+                        format!("hook run_as '{run_as}' is orphan")
+                    };
+                    warnings.push(LoadWarning::HookInvariantSkippedDueToOrphan {
+                        mailbox: mailbox_name.clone(),
+                        hook_name: hook_effective_name.clone(),
+                        reason: skip_reason,
+                    });
+                }
+            } else {
+                check_hook_owner_invariant(config, mailbox_name, mb, hook)?;
+            }
 
-            let effective = effective_hook_name(hook);
+            let effective = hook_effective_name;
             let is_explicit = hook.name.is_some();
             if let Some((prior_mb, prior_explicit)) =
                 seen.insert(effective.clone(), (mailbox_name.clone(), is_explicit))
@@ -714,7 +837,7 @@ pub(crate) fn validate_hooks(config: &Config) -> Result<(), String> {
             }
         }
     }
-    Ok(())
+    Ok(warnings)
 }
 
 /// Expose for the daemon handler, which needs to pre-validate a single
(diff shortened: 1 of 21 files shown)


uzyn (commented):
## Sprint Goal Assessment

Sprint 1 lands the schema foundation for the agent-integration track cleanly. `MailboxConfig.owner` is a required field, `VALID_HOOK_RUN_AS_VALUES` is retired in favor of runtime `getpwnam`, the hook/owner invariant is enforced at both `Config::load` and UDS `HOOK-CREATE` (via `validate_hooks`), and `Config::load` now returns a warning vector instead of failing on orphan users. Everything in this sprint is pure schema + validation per the stated scope; no disk-ownership changes leak in. `cargo test` (1065 unit + 96 integration), `cargo clippy --all-targets -- -D warnings`, and `cargo fmt -- --check` all pass locally.

**S1-1 MailboxConfig.owner**
- [x] `pub owner: String` added as a required non-defaulted field
- [x] Missing `owner` fails `Config::load` with an actionable error naming the mailbox (`src/config.rs:1070-1078`)
- [x] `owner_uid` / `owner_gid` helpers exposed via `validate_run_as` — `#[allow(dead_code)]` until Sprint 2 consumes them
- [x] Unit tests cover valid / missing / orphan / empty-string owners
- [x] TOML round-trip test `load_mailbox_owner_roundtrip_preserved` preserves `owner`
- [x] Linters clean

**S1-2 Retire `VALID_HOOK_RUN_AS_VALUES`; runtime `getpwnam`**
- [x] Constant and `default_hook_run_as()` removed; `HookTemplate.run_as` is required
- [x] `validate_run_as` returns `RunAsKind::{Reserved, User}` on success, `ConfigError::{InvalidUsername, OrphanUser}` on failure
- [x] Unit tests cover reserved accept / regex-fail variants / mocked `getpwnam` success / orphan signal
- [x] Existing `run_as = "aimx-hook"` fixtures updated to `run_as = "root"` (invariant-safe on any host)
- [x] Linters clean

**S1-3 Enforce hook/mailbox invariant at load and UDS**
- [x] `check_hook_owner_invariant` helper exposed from `src/config.rs`
- [x] Called from `validate_hooks` at load time (`src/config.rs:687`)
- [x] UDS `HOOK-CREATE` enforces it via `validate_hooks(&new_config)` inside `hook_handler::handle_hook_create` (`src/hook_handler.rs:209`) — AC met through the `validate_hooks` gate rather than a direct `check_hook_owner_invariant` call. This is functionally equivalent and correct.
- [x] Error message names mailbox, hook, actual run_as, expected owner, and suggested fix
- [x] Catchall variant names `aimx-catchall` as the expected value
- [x] Unit tests cover all the PRD §6.3 branches

**S1-4 Orphan tolerance**
- [x] `LoadWarning` enum with the four variants from the spec
- [x] `Config::load` signature now returns `Result<(Self, Vec<LoadWarning>), _>`
- [x] `ConfigResolved` side-table with `is_mailbox_active` / `is_template_active`
- [x] Startup warnings logged under `aimx::config` target in `src/main.rs::emit_load_warnings` and in `src/serve.rs::reload_config` SIGHUP path
- [x] Unit tests cover valid → empty warnings, single orphan owner, single orphan template, mixed valid + orphan
- [x] Test-only `user_resolver::set_test_resolver` with two-layer `RwLock + Mutex` serialization — behaves correctly in parallel tests
- [x] Linters clean

## Potential Bugs

### Non-blocker: orphan-template + invariant interplay hard-fails the config

**File:** `src/config.rs::resolve_effective_run_as` + `check_hook_owner_invariant`
**Scenario:** A template with `run_as = "old-sam"` (orphan — no longer on the host) is referenced by a hook attached to a mailbox with a different valid owner. `resolve_effective_run_as` returns `"old-sam"` from the template; `check_hook_owner_invariant` compares it against the mailbox owner, finds a mismatch, and returns `Err`. `Config::load` then refuses to start the daemon.

Reproduced locally:

```toml
[[hook_template]]
name = "t1"
run_as = "totally-missing-user-xyzzy"
cmd = ["/bin/true"]

[mailboxes.alice]
owner = "root"

[[mailboxes.alice.hooks]]
event = "on_receive"
template = "t1"
```

`aimx doctor` output: `Error: hook 'uses-t1' on mailbox 'alice' has run_as='totally-missing-user-xyzzy' but the mailbox is owned by 'root'; ...`

Per PRD §6.1, "Hook templates with unknown `run_as` users are retained in the in-memory config as orphan-flagged rather than rejected" and "Hooks attached to mailboxes whose `run_as` user is missing are soft-skipped at fire time." The spirit of orphan tolerance is that a user deletion should not kill the daemon. The current behavior does.

Mitigations in the existing design are real (the hook would soft-skip at fire time anyway), but the daemon never gets the chance because load itself rejects. One fix: if the effective `run_as` is orphan-flagged via `ConfigResolved`, downgrade the invariant mismatch to a warning (the hook is unfireable anyway). Alternatively, short-circuit the invariant when `resolve_effective_run_as` returns a name that isn't resolvable.

Classifying this as **Non-blocker** because: (a) the sprint AC S1-3 literally says "invariant violation returns an actionable error" at load time, with no orphan-tolerance carve-out; (b) the bug is latent (needs orphan template + hook attached to different-owner mailbox); (c) Sprint 8 strips `invoke-<agent>` defaults, shrinking the surface. But it should be tracked — the test suite has no coverage for this interaction.

### Non-blocker: error message duplicates `root` in the suggested-fix clause when owner is `root`

**File:** `src/config.rs:366-374`
**Example output:** `"fix: set run_as='root' or run_as='root'"` (reproduced above).

The message template reads `fix: set run_as='{owner}' or run_as='{root}'`. When `owner` is already `"root"`, both placeholders render identically and the operator sees two copies of the same suggestion. Low-impact (message is still technically correct), but sloppy UX for the one case where operators put `root` on a non-catchall mailbox.

### Non-blocker: asymmetric treatment of regex-invalid usernames between owner and hook `run_as`

**Files:** `src/config.rs::validate_mailbox_owners` vs. `validate_hook_run_as`

`validate_hook_run_as` hard-rejects regex-invalid values (`ConfigError::InvalidUsername` → `return Err(...)`) while `validate_mailbox_owners` downgrades regex-invalid owners to `OrphanMailboxOwner` warnings. A user typing `owner = "Bad Name"` gets a silent warning + inactive mailbox; typing `run_as = "Bad Name"` on a hook kills the daemon.

The code comment at `config.rs:970-974` acknowledges this is intentional ("We return this as a warning rather than a hard failure so the daemon stays up"), but the inconsistency is surprising and not called out in the PRD. Either direction would be defensible; doing it the same way across both call sites would be less surprising. Worth a follow-up.

## Security Issues

Nothing I can see. The PR adds no new attack surface — it retires a stale allowlist and threads owner metadata through the schema. `getpwnam` is called with the exact string the operator committed to `config.toml`, which is already privileged input. No new unsafe that isn't the single documented `libc::getpwnam` call in `user_resolver.rs`, and the pointer-dereference is correctly scoped to immediately-read-and-copy.

## Test Coverage

Coverage is strong for everything the sprint explicitly promises. Unit tests hit every S1-1 through S1-4 AC; the test-resolver override enables deterministic regex/getpwnam branches on any CI host. Gaps I'd flag:

- **Orphan template + invariant interaction is not tested.** The scenario in "Potential Bugs" above has no test coverage. Adding a case that asserts the *current* behavior (either hard-fail or soft-pass, whichever we pick) would lock in whichever semantics we commit to.
- **Integration test `hook_templates_end_to_end_mcp_to_sandbox` weakened.** Previously asserted `uid != 0` when running as root (proving the privilege-drop fired for `run_as = "aimx-hook"`). Now asserts `uid == 0` because the fixture was flipped to `run_as = "root"`. The Sprint 1 goal is schema-only, so there's nothing to drop to yet — but the integration-level coverage that the daemon can actually change uid is now gone. Flagging so Sprint 2 restores this explicitly with a real test user.
- **Stale comment in `setup_test_env_with_template`** (`tests/integration.rs:4352-4355`): the comment says `run_as = "aimx-hook"` is retained to exercise the orphan path, but the literal `format!` on the next line uses `run_as = "root"`. The comment is misleading; code is fine.

## Code Quality

Good overall — new code is well-commented, ownership is clear, the test-resolver abstraction is a clean seam. Minor observations:

- `user_resolver::_RESOLVER_SCAFFOLD` is dead code justified by a speculative SIGHUP cache. Marked `#[allow(dead_code)]`. Either use it or delete it; dead scaffold has a way of growing dust. Non-blocker.
- `Config::load_ignore_warnings` / `load_resolved_ignore_warnings` are pragmatic transitional helpers. Once ingest / MCP / UDS start consuming warnings (Sprint 2+), these should be audited for call sites that legitimately should be surfacing warnings but are currently silencing them. Worth a TODO-style comment pointing at the audit.
- `check_hook_owner_invariant` uses `Result<(), String>` for its error type. Consistent with the rest of `config.rs` but losing structured error information; a dedicated error enum would let the UDS handler render a different (structured) response frame. Low priority.

## Alignment with PRD

Matches the PRD's intent. The only drift is the orphan-template/invariant interaction noted above (§6.1 + §6.3 tension). P0 requirements from §5 are correctly scoped — multi-user isolation and UDS authn land in Sprints 2–4, not here.

One doc/spec observation: the PRD says (§6.1) that `run_as = root` / `run_as = aimx-catchall` template creation is rejected by the UDS regardless of caller. That's Sprint 5 territory (TEMPLATE-CREATE verb doesn't exist yet), but nothing in this PR prevents it from landing later.

**Overall verdict:** Needs minor fixes — specifically the orphan-template/invariant interaction surfaces as a Non-blocker but the absence of a test case around it is worth covering so the semantics are pinned. None of this is a merge blocker.

**Blockers:** None.

**Non-blockers:**
1. Orphan-template + hook invariant hard-fails `Config::load` when it should probably soft-pass per PRD §6.1 orphan tolerance — at minimum, add a test that pins the chosen semantics so Sprint 2+ doesn't regress.
2. Error message duplicates `root` in the suggested-fix clause when mailbox owner is `root`.
3. Asymmetric regex-invalid handling between `validate_mailbox_owners` (warning) and `validate_hook_run_as` (hard error) — pick one.
4. Integration test `hook_templates_end_to_end_mcp_to_sandbox` flipped its UID assertion from `!= 0` to `== 0`; restore real privilege-drop E2E coverage in Sprint 2.

**Nice-to-haves:**
- Stale comment in `setup_test_env_with_template` says `run_as = "aimx-hook"` but code sets `root`.
- `_RESOLVER_SCAFFOLD` dead scaffold — use or delete.
- `check_hook_owner_invariant` returns `Result<(), String>` rather than a structured error enum.

uzyn:
## Review-cycle follow-ups

Addressed the four non-blockers from the latest review. Pushed as `b544d97`.

### Fixed

1. **Orphan-template + invariant hard-fail (PRD §6.1/§6.3 tension).** Added `LoadWarning::HookInvariantSkippedDueToOrphan` and threaded a new `OrphanSkipContext` through `validate_hooks`. On the load path, a hook whose effective `run_as` (or attached template's `run_as`, or mailbox owner) is already orphan-flagged now downgrades the invariant mismatch to a warning rather than killing the daemon — the hook is unfireable anyway, which is consistent with PRD §6.1's user-deletion tolerance. UDS `HOOK-CREATE` and `aimx hooks create` keep `OrphanSkipContext::strict()`: fresh creates still hard-reject so operators don't silently commit a broken stanza. Test coverage:
   - `load_orphan_template_with_invariant_mismatch_becomes_warning` (reproduces the exact reviewer fixture: template with `run_as = "totally-missing-user"` attached to an `owner = "root"` mailbox → load succeeds with one `OrphanTemplateRunAs` + one `HookInvariantSkippedDueToOrphan`).
   - `load_orphan_mailbox_owner_with_invariant_mismatch_becomes_warning` (orphan owner + non-root mismatched run_as → soft-pass).
   - `load_fresh_hook_create_still_hard_rejects_invariant_mismatch` (pins strict context semantics).

2. **Duplicate `run_as='root'` in the suggested-fix clause.** `check_hook_owner_invariant` now renders `set run_as='root'` (no `or` clause) when the mailbox owner is already root. Non-root owners still get the full `set run_as='X' or run_as='root'` hint. Tests: `invariant_error_drops_redundant_root_when_owner_is_root` + `invariant_error_retains_or_clause_for_non_root_owner`.

3. **Asymmetric regex-invalid handling between owner and hook `run_as`.** Aligned `validate_mailbox_owners` with `validate_hook_run_as` (the stricter of the two, per PRD §6.1: "Username regex failures are hard-rejected at load"). A mailbox `owner = "Bad Name"` now hard-fails `Config::load` with a clear message instead of silently producing an `OrphanMailboxOwner` warning. Test: `load_regex_invalid_mailbox_owner_hard_fails`. The existing `load_mailbox_orphan_owner_warns_not_errors` test uses a regex-valid-but-unresolvable name (`deleted-user`), so it continues to exercise the orphan-warning path.

### Intentionally left as-is

4. **Integration test `hook_templates_end_to_end_mcp_to_sandbox` `uid == 0` assertion.** Per the review, this is explicitly tracked for Sprint 2 — restoring `uid != 0` requires a real non-root test user that Sprint 2 creates. Left a `TODO(sprint-2)` comment at the assertion site so the tracking is visible in-tree.

### Also swept up

- Stale `aimx-hook` comment in `setup_test_env_with_template` — updated to match the code (which uses `run_as = "root"`).
- Existing test `load_enforces_invariant_on_mailbox_hook` was relying on `run_as = "someoneelse"` being a getpwnam miss on the CI host. With the new orphan-skip in place, that would no longer hard-fail; the test now installs a test resolver that resolves `someoneelse` so the invariant check is exercised cleanly, independent of the orphan-tolerance path.

### Not addressed (explicitly out of scope for this fix cycle)

- `_RESOLVER_SCAFFOLD` dead-code scaffold — kept as-is per the original justification (SIGHUP cache); can be revisited in Sprint 2 when the caching policy firms up.
- `check_hook_owner_invariant` still returns `Result<(), String>` rather than a structured enum — consistent with the rest of `config.rs`.
- `Config::load_ignore_warnings` / `load_resolved_ignore_warnings` — still transitional; Sprint 2+ will audit call sites as ingest / MCP start consuming warnings.

### Verification

- `cargo test` — 1069 unit + 96 integration, all passing.
- `services/verifier` crate — unaffected, still clippy-clean.

uzyn (commented):
## Re-Review — All Four Non-Blockers Resolved

Verified against commit `b544d97`. Ran `cargo test` on the six new unit tests plus the full `config::tests` suite (96 passed), `cargo clippy --all-targets -- -D warnings` (clean), and `cargo fmt -- --check` (clean).

### Resolved

1. **Orphan-template + invariant hard-fail** — Fixed correctly. `LoadWarning::HookInvariantSkippedDueToOrphan` is added, and `OrphanSkipContext` is threaded through `validate_hooks`. The skip branch correctly consults three orphan sources (mailbox owner, explicit hook `run_as`, referenced template) and keyed on `(mailbox, hook_effective_name)` so it only skips the specific hook flagged — not every hook on the mailbox. The `strict()` context is used in both UDS `HOOK-CREATE` (`src/hook_handler.rs:215`) and `aimx hooks create` (`src/hooks.rs:496`), so fresh creates still hard-reject as required. Regression tests `load_orphan_template_with_invariant_mismatch_becomes_warning`, `load_orphan_mailbox_owner_with_invariant_mismatch_becomes_warning`, and `load_fresh_hook_create_still_hard_rejects_invariant_mismatch` pin the semantics on both branches.

2. **Duplicate `root` in suggested-fix clause** — Fixed. `check_hook_owner_invariant` (`src/config.rs:542-550`) now renders `set run_as='root'` (no redundant `or` clause) when the mailbox owner is `root`, and retains the full `set run_as='X' or run_as='root'` hint for non-root owners. Both branches are covered by `invariant_error_drops_redundant_root_when_owner_is_root` and `invariant_error_retains_or_clause_for_non_root_owner`.

3. **Asymmetric regex-invalid handling** — Fixed. `validate_mailbox_owners` (`src/config.rs:1090-1112`) now returns `Result<Vec<LoadWarning>, String>` and hard-rejects on `ConfigError::InvalidUsername` with a clear message, symmetric with `validate_hook_run_as`. The pre-existing stale comment justifying the asymmetry (previously at ~L970) is replaced with an accurate doc-comment. Covered by `load_regex_invalid_mailbox_owner_hard_fails`. Existing orphan-tolerance test (`load_mailbox_orphan_owner_warns_not_errors`) still passes because it uses a regex-valid name (`deleted-user`).

4. **Integration test `hook_templates_end_to_end_mcp_to_sandbox` UID assertion** — Intentionally deferred to Sprint 2 as expected. `TODO(sprint-2)` comment added at `tests/integration.rs:5024-5029` explaining why `uid == 0` is a Sprint 1 concession and when to flip it back. Acceptable per original review guidance.

### Also swept up (bonus)

- Stale `run_as = "aimx-hook"` comment in `setup_test_env_with_template` (previously flagged as nice-to-have) was updated to accurately describe the `run_as = "root"` fixture.
- The existing `load_enforces_invariant_on_mailbox_hook` test was updated to install a test resolver that resolves `someoneelse`, so the invariant-check branch is isolated from the new orphan-skip path. Good defensive update.

### Still unresolved

None — all four flagged issues are addressed or intentionally deferred.

### New issues found

None. Reviewed the `validate_hooks` orphan-skip logic for correctness:

- Three orphan predicates (`owner_is_orphan`, `hook_run_as_is_orphan`, `template_is_orphan`) correctly combined with `||`.
- The `hook_effective_name` is computed once and reused both for the orphan-lookup key and the subsequent duplicate-detection pass — avoids any chance of a key-mismatch against the `OrphanHookRunAs` warning emitted earlier.
- The skip path still calls `check_hook_owner_invariant` (rather than skipping blindly) and only downgrades *on mismatch* — a matching hook on an orphan mailbox produces no spurious warning.
- The strict context used by both create paths (UDS + CLI) correctly has all three `HashSet`s empty, so no skipping can occur there.
- `OrphanSkipContext::from_warnings` iterates on the warning vector built by `validate_hook_templates`, `validate_mailbox_owners`, and `validate_hook_run_as` *before* `validate_hooks` runs — the ordering in `Config::load` is correct.

### Verdict

**READY_TO_MERGE**

All acceptance criteria for S1-1 through S1-4 are met, all four prior non-blockers are resolved with dedicated regression coverage, and the sprint stays within its schema-only scope. Disk ownership, setup UX, UDS SO_PEERCRED, and the agent-setup rewrite remain correctly deferred to later sprints.

```
[Sprint 1] Config schema + invariants (agent-integration track)

- S1-1: MailboxConfig.owner required field + owner_uid/gid helpers
- S1-2: Retire VALID_HOOK_RUN_AS_VALUES; runtime validate_run_as with
  reserved-name / regex / getpwnam layers and a trait-seam test resolver
- S1-3: Shared check_hook_owner_invariant enforced at Config::load and
  UDS HOOK-CREATE (via validate_hooks); catchall-aware error messages
- S1-4: Config::load returns Vec<LoadWarning>; orphan mailboxes and
  templates retained but flagged inactive via ConfigResolved; SIGHUP
  reload re-emits warnings
- Orphan tolerance × invariant: HookInvariantSkippedDueToOrphan soft-
  passes load when either side is orphan-flagged (PRD §6.1); UDS/CLI
  create paths stay strict via OrphanSkipContext::strict()
- Regex-invalid usernames hard-reject symmetrically across owner and
  hook run_as (PRD §6.1)

1069 unit + 96 integration tests passing; clippy + fmt clean.
```

PR #127, opened 2026-04-22
[Sprint 3] Catchall narrowing + setup refactor

uzyn:
## Summary

Delivers the four Sprint 3 stories for the agent-integration track (§6.4 / §6.8 in the PRD). The daemon now only provisions the `aimx-catchall` service user when a catchall mailbox is actually configured, the legacy `chown_datadir_for_hook_user` and `configure_hook_templates` code paths are gone, and `Config::load` rejects `owner = \"root\"` on any non-catchall mailbox with a clear hint toward the new `allow_root_catchall` escape hatch.

- **S3-1** — `ensure_hook_user` → `ensure_catchall_user`; gated on configured catchall; `aimx-catchall` created with `--system`, `/usr/sbin/nologin`, home `/var/lib/aimx-catchall` (0700 `aimx-catchall:aimx-catchall`); idempotent. Section title flips `[User]` → `[Catchall]`; `MockSystemOps` takes the new `create_system_user(user, home)` / `ensure_owned_directory` shape.
- **S3-2** — deletes `chown_datadir_for_hook_user`; adds `prompt_mailbox_owner(address, sys)` with local-part default (when `getpwnam(local_part)` resolves), actionable re-prompt bounded to 5 attempts, and `AIMX_NONINTERACTIVE=1` fallback. Catchall owner stays hard-coded to `aimx-catchall` in `finalize_setup`.
- **S3-3** — removes `configure_hook_templates` and every helper it called; setup no longer prints the \"Hook Templates\" checkbox section. The `grep -r \"configure_hook_templates\\|tick the template\\|interactive checkbox\" src/ tests/` AC returns zero hits. `aimx hooks list --templates` and `aimx doctor` no-template hints now point at `aimx agent-setup <agent>`.
- **S3-4** — `Config::load` rejects `owner = \"root\"` on non-catchall mailboxes with the PRD wording; new `MailboxConfig.allow_root_catchall` (`#[serde(default)]` + `skip_serializing_if`) gates the catchall escape hatch and surfaces `LoadWarning::RootCatchallAccepted` so the elevation is audit-logged. `allow_root_catchall` on a non-catchall mailbox is rejected.

Test fixtures across the crate that used `owner = \"root\"` as a convenient default for non-catchall mailboxes have been moved to `owner = \"ops\"` (orphan-warn, not hard-fail). Sprint 1's `load_accepts_mailbox_with_owner_root` test is renamed and flipped to assert the new rejection; new S3-4 tests cover every branch.

## Test plan

- [x] `cargo test` — 1090 unit + 97 integration + 1 isolation (ignored/root-gated) green
- [x] Sprint 3 grep AC: `grep -r \"configure_hook_templates\\|tick the template\\|interactive checkbox\" src/ tests/` returns no hits
- [x] `services/verifier` crate still builds and tests (43 tests pass)

### DIFF
--- a/book/cli.md
+++ b/book/cli.md
@@ -108,6 +108,10 @@ Alias: `aimx mailbox` works identically to `aimx mailboxes`.
 
 Create a mailbox registering `<name>@<domain>` and directories under `inbox/<name>/` and `sent/<name>/`. Hot-reloaded by the daemon; no restart needed.
 
+| Flag | Description |
+|------|-------------|
+| `--owner <user>` | Linux user that owns the mailbox's storage. When omitted, the CLI prompts (defaulting to `<name>` if a user with that name already exists). Under `AIMX_NONINTERACTIVE=1` the default is accepted when available, otherwise the command errors hard so scripted installs fail fast. |
+
 ### `aimx mailboxes list`
 
 List all mailboxes with addresses, total count, and unread count.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -183,10 +183,12 @@ pub enum MailboxCommand {
         /// Mailbox name (local part of email address)
         name: String,
 
-        /// Linux user that owns the mailbox's storage. Defaults to the
-        /// local part of the address if a user with that name already
-        /// exists on the host; otherwise required. The owner must
-        /// resolve via `getpwnam` at daemon load time.
+        /// Linux user that owns the mailbox's storage. When omitted,
+        /// the CLI prompts (defaulting to the local part of the
+        /// address when a user with that name already exists on the
+        /// host). Under `AIMX_NONINTERACTIVE=1` the default is
+        /// accepted when available, or the command errors hard. The
+        /// owner must resolve via `getpwnam` at daemon load time.
         #[arg(long)]
         owner: Option<String>,
     },
--- a/src/config.rs
+++ b/src/config.rs
@@ -1110,8 +1110,11 @@ fn is_valid_placeholder_name(s: &str) -> bool {
 /// mailbox. On a catchall it is accepted only when
 /// `allow_root_catchall = true` — the escape hatch — and that
 /// acceptance surfaces as a [`LoadWarning::RootCatchallAccepted`] so
-/// the elevation is logged. `allow_root_catchall` on any non-catchall
-/// mailbox, or on a catchall with a non-root owner, is rejected.
+/// the elevation is logged. `allow_root_catchall = true` on any
+/// non-catchall mailbox is rejected hard. On a catchall whose owner is
+/// not `root` (for example the default `aimx-catchall`), the flag is a
+/// no-op — it only gates the root escape hatch — and is silently
+/// accepted.
 fn validate_mailbox_owners(config: &Config) -> Result<Vec<LoadWarning>, String> {
     let mut warnings = Vec::new();
     for (name, mb) in &config.mailboxes {
--- a/src/config.rs
+++ b/src/config.rs
@@ -79,7 +79,7 @@ pub struct Config {
 
     /// Installed hook templates. Each entry is a pre-vetted command shape
     /// an agent can reference via MCP `hook_create`. Populated by
-    /// `aimx setup`'s interactive checkbox list. Validated at load time.
+    /// `aimx agent-setup <agent>` (Sprint 3+). Validated at load time.
     ///
     /// Serialized as `[[hook_template]]` blocks (singular) to match the
     /// PRD wording and the usual TOML convention for array-of-tables.
@@ -305,6 +305,17 @@ pub struct MailboxConfig {
     /// [`MailboxConfig::effective_trusted_senders`] to resolve.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub trusted_senders: Option<Vec<String>>,
+
+    /// Escape hatch for running a catchall mailbox with `owner = "root"`
+    /// (PRD §6.2). Default `false`; only honoured when the mailbox is a
+    /// catchall AND `owner = "root"`. `Config::load` rejects the flag on
+    /// any non-catchall mailbox and emits a `LoadWarning` when it is set
+    /// alongside `owner = "root"` on a catchall so operators see the
+    /// escape-hatch acknowledgement in `aimx logs`. Omitted from the
+    /// serialized `config.toml` when false so standard configs stay
+    /// minimal.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub allow_root_catchall: bool,
 }
 
 impl MailboxConfig {
@@ -418,6 +429,10 @@ pub enum LoadWarning {
     /// operator can clean up.
     #[allow(dead_code)]
     LegacyAimxHookUser,
+    /// Catchall mailbox is configured with `owner = "root"` and the
+    /// `allow_root_catchall = true` escape hatch (PRD §6.2). Logged at
+    /// WARN so the audit trail records the elevation.
+    RootCatchallAccepted { mailbox: String },
 }
 
 impl LoadWarning {
@@ -455,6 +470,11 @@ impl LoadWarning {
                  manages it — remove via 'userdel aimx-hook' when safe"
                     .to_string()
             }
+            LoadWarning::RootCatchallAccepted { mailbox } => format!(
+                "catchall mailbox '{mailbox}' is running with owner='root' \
+                 and allow_root_catchall=true; this is a documented escape \
+                 hatch (PRD §6.2) — mail lands owned by uid 0"
+            ),
         }
     }
 }
@@ -1084,9 +1104,14 @@ fn is_valid_placeholder_name(s: &str) -> bool {
 /// failures hard-reject (symmetric with [`validate_hook_run_as`] — per
 /// PRD §6.1 a regex-invalid owner can never resolve, so it is treated
 /// as a config typo rather than an orphan). `getpwnam` misses surface
-/// as [`LoadWarning::OrphanMailboxOwner`] (PRD §6.2). Reserved names
-/// (`root`, `aimx-catchall`) pass silently — the extra rules around
-/// root / catchall owners land in Sprint 3's S3-4 check.
+/// as [`LoadWarning::OrphanMailboxOwner`] (PRD §6.2).
+///
+/// S3-4 (PRD §6.2): `owner = "root"` is rejected on any non-catchall
+/// mailbox. On a catchall it is accepted only when
+/// `allow_root_catchall = true` — the escape hatch — and that
+/// acceptance surfaces as a [`LoadWarning::RootCatchallAccepted`] so
+/// the elevation is logged. `allow_root_catchall` on any non-catchall
+/// mailbox, or on a catchall with a non-root owner, is rejected.
 fn validate_mailbox_owners(config: &Config) -> Result<Vec<LoadWarning>, String> {
     let mut warnings = Vec::new();
     for (name, mb) in &config.mailboxes {
@@ -1106,6 +1131,39 @@ fn validate_mailbox_owners(config: &Config) -> Result<Vec<LoadWarning>, String> 
                     owner: mb.owner.clone(),
                 });
             }
+        }
+
+        let is_catchall = mb.is_catchall(config);
+        let owner_is_root = mb.owner == RESERVED_RUN_AS_ROOT;
+
+        if mb.allow_root_catchall && !is_catchall {
+            return Err(format!(
+                "mailbox '{name}' sets allow_root_catchall=true but is not a \
+                 catchall (*@{domain}); remove the flag or change the address",
+                domain = config.domain,
+            ));
+        }
+
+        if owner_is_root && !is_catchall {
+            return Err(format!(
+                "mailbox '{name}' cannot be owned by root; use a regular \
+                 Linux user or set 'allow_root_catchall = true' on a \
+                 catchall mailbox"
+            ));
+        }
+
+        if is_catchall && owner_is_root {
+            if !mb.allow_root_catchall {
+                return Err(format!(
+                    "catchall mailbox '{name}' has owner='root' but \
+                     allow_root_catchall=false; set allow_root_catchall=true \
+                     to opt into the root-catchall escape hatch, or change \
+                     owner to 'aimx-catchall' (the default)"
+                ));
+            }
+            warnings.push(LoadWarning::RootCatchallAccepted {
+                mailbox: name.clone(),
+            });
         }
     }
     Ok(warnings)
@@ -1444,7 +1502,7 @@ owner = "aimx-catchall"
 
 [mailboxes.support]
 address = "support@agent.example.com"
-owner = "root"
+owner = "ops"
 
 [[mailboxes.support.hooks]]
 name = "support_inbound"
@@ -1492,7 +1550,7 @@ domain = "test.com"
 
 [mailboxes.support]
 address = "support@test.com"
-owner = "root"
+owner = "ops"
 trust = "strict"
 "#,
         )
@@ -1525,7 +1583,7 @@ owner = "aimx-catchall"
 
 [mailboxes.public]
 address = "hello@test.com"
-owner = "root"
+owner = "ops"
 trust = "none"
 "#,
         )
@@ -1568,7 +1626,7 @@ domain = "test.com"
 
 [mailboxes.support]
 address = "support@test.com"
-owner = "root"
+owner = "ops"
 
 [[mailboxes.support.hooks]]
 name = "support_env"
@@ -1594,7 +1652,7 @@ domain = "test.com"
 
 [mailboxes.support]
 address = "support@test.com"
-owner = "root"
+owner = "ops"
 
 [[mailboxes.support.hooks]]
 event = "on_receive"
@@ -1625,7 +1683,7 @@ domain = "test.com"
 
 [mailboxes.support]
 address = "support@test.com"
-owner = "root"
+owner = "ops"
 
 [[mailboxes.support.hooks]]
 event = "on_receive"
@@ -1652,7 +1710,7 @@ domain = "test.com"
 
 [mailboxes.support]
 address = "support@test.com"
-owner = "root"
+owner = "ops"
 
 [[mailboxes.support.hooks]]
 event = "on_receive"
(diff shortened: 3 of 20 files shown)


uzyn (commented):
## Sprint Goal Assessment

Sprint 3's goal — `aimx setup` no longer creates `aimx-hook`, no longer chowns the whole datadir, no longer runs the `configure_hook_templates` checkbox, *does* create `aimx-catchall` only when a catchall is configured, and `Config::load` rejects `owner = "root"` on non-catchall mailboxes — is substantially achieved. The refactor is clean and the code deletion is significant (~500 lines out of setup.rs, plus the full `configure_hook_templates` surface). The catchall-narrowing gate, the `allow_root_catchall` escape hatch, and the full four-branch rejection matrix all land. One acceptance criterion lands in a vacuous form rather than an engaged one — see **Alignment with PRD** below.

**[S3-1] Rename `ensure_hook_user` → `ensure_catchall_user`, on-demand only**
- [x] `ensure_hook_user` renamed to `ensure_catchall_user`; `aimx-hook` string constants removed (confirmed via grep: `HOOK_SERVICE_USER` is gone from the codebase)
- [x] `run_setup` invokes `ensure_catchall_user` only when `config_has_catchall(&config_path)` is true
- [x] Section title flips `[User]` → `[Catchall]` (`src/setup.rs:240`)
- [x] `MockSystemOps` carries the new `create_system_user(user, home)` + `ensure_owned_directory(path, owner, group, mode)` shape; tests assert the recorded tuple
- [x] `useradd` invoked with `--system`, `--home-dir`, `/usr/sbin/nologin` per PRD §6.4; Alpine/BusyBox `adduser` fallback has the corresponding `-h <home>` flag
- [x] Re-entrant setup: the catchall detection only cares about the on-disk mailbox stanza, so the same guard applies to fresh and re-entrant runs
- [x] `cargo clippy --all-targets -- -D warnings` + `cargo fmt -- --check` clean on this branch

**[S3-2] Remove `chown_datadir_for_hook_user`; add mailbox-owner prompt**
- [x] `chown_datadir_for_hook_user` deleted; its call site removed
- [x] `prompt_mailbox_owner(address, sys) -> Result<String>` helper added with local-part default, re-prompt (bounded to `MAX_OWNER_PROMPT_ATTEMPTS = 5`), `AIMX_NONINTERACTIVE=1` fallback with hard error when no default is available
- [x] Catchall owner is hard-coded to `aimx-catchall` by `finalize_setup` — never prompted
- [x] Unit tests exercise: non-interactive with resolvable default, non-interactive with no default (errors with `useradd` hint), bounded-attempts constant check
- [ ] **PARTIAL:** `run_setup` does not call `prompt_mailbox_owner` for any mailbox. The helper is `#[allow(dead_code)]` with a note that Sprint 5+ will wire it into CLI paths. This is vacuously AC-met under today's flow (setup only ever creates the catchall, which is hard-coded), but the prompt has no live caller. See **Alignment with PRD** below.

**[S3-3] Remove `configure_hook_templates` interactive phase**
- [x] `configure_hook_templates` + `current_hook_templates` + `apply_selected_templates` + `HookTemplateChoice` + `prompt_hook_template_selection` + `is_stdin_tty` all deleted (confirmed via regression test `legacy_template_phase_is_absent_from_setup`)
- [x] `grep -r "configure_hook_templates\|tick the template\|interactive checkbox" src/ tests/` returns zero hits (verified locally)
- [x] `aimx hooks list --templates` and `aimx doctor` no-template hints now point at `aimx agent-setup <agent>` (`src/hooks.rs:99`, `src/doctor.rs:767`)
- [x] `hook_templates_defaults.rs` keeps the compile-time validation of the embedded TOML as a build-time guard (marked `#![allow(dead_code)]` since no runtime caller remains); Sprint 8 will strip the `invoke-*` blocks
- [x] Snapshot tests updated

**[S3-4] Reject `owner = "root"` for non-catchall mailboxes at load time**
- [x] `MailboxConfig.allow_root_catchall: bool` added with `#[serde(default, skip_serializing_if = "std::ops::Not::not")]` so default-false doesn't bloat `config.toml`; `allow_root_catchall_default_false_omitted_from_serialized_toml` locks the serialization contract
- [x] Non-catchall with `owner = "root"` is rejected with the PRD wording ("cannot be owned by root"; mentions `allow_root_catchall`); `load_rejects_non_catchall_owner_root` covers it
- [x] Catchall with `owner = "root"` + `allow_root_catchall = false` rejected with "set allow_root_catchall=true" hint
- [x] Catchall with `owner = "root"` + `allow_root_catchall = true` accepted AND emits `LoadWarning::RootCatchallAccepted` so the elevation is audit-logged
- [x] `allow_root_catchall = true` on a non-catchall mailbox is rejected hard
- [x] All four branches covered by unit tests

## Test Coverage

Good. 238 setup tests + 101 config tests + 46 doctor tests pass locally. The S3-4 matrix is fully covered, the regression test for the removed functions prevents silent reintroduction, and the `MockSystemOps` reshape gives tests a clean seam to assert the new `create_system_user(user, home)` + `ensure_owned_directory(path, owner, group, mode)` tuples.

Gaps worth noting:

- `prompt_mailbox_owner` has no test for the **re-prompt-on-invalid-user** interactive path. `prompt_mailbox_owner_max_attempts_is_bounded` asserts only the ceiling constant — it does not exercise the re-prompt loop with a bounded-length scripted stdin to verify the error message, re-prompt message, and final "after 5 attempts" failure path. Since the helper has no live callers today, this is non-blocking, but when Sprint 5+ wires it in, the interactive branch should get real coverage.
- The docstring for `validate_mailbox_owners` (config.rs:1113-1114) claims "`allow_root_catchall` on ... a catchall with a non-root owner, is rejected." The code does not reject that combination (`owner = "aimx-catchall"` + `allow_root_catchall = true` silently loads). No test exists to pin either behavior. Small scope drift between the doc and the code — non-blocking because the flag is inert when owner is not root, but the doc should match the code.
- `run_setup_skips_catchall_user_when_no_catchall_configured` (src/setup.rs:2327) has an inline comment claiming the fixture has `owner = "root"` when it actually has `owner = "ops"`, and describing a load-failure path that doesn't happen. The test still asserts the right thing (`!config_has_catchall` for a non-wildcard mailbox), but the comment misleads anyone later trying to understand the test's reasoning.

## Potential Bugs

None functionally blocking. Two minor observations:

- **[Non-blocker]** `validate_mailbox_owners` silently accepts `allow_root_catchall = true` on a catchall with `owner = "aimx-catchall"`. The flag is inert in that case, so it's not a security or correctness hazard, but it diverges from the docstring above and means an operator who toggles the flag with a non-root owner gets no feedback. Either tighten the check or fix the docstring.
- **[Non-blocker]** In `run_setup`, `let _ = non_interactive;` at `src/setup.rs:1899` keeps the parameter on the `fn` signature while discarding the value. The rationale (stable CLI signature while the interactive phase is being replaced) is reasonable, but a post-Sprint-8 follow-up should either wire it to `prompt_mailbox_owner` if the prompt ever lands in setup, or drop the argument entirely if it never will.

## Security Issues

No new issues. Deleting `chown_datadir_for_hook_user` tightens the permissions story (per-mailbox chown now owns the ownership invariant, per Sprint 2), and restricting `aimx-catchall` creation to catchall-configured hosts reduces unnecessary system-user footprint. The `RESERVED_RUN_AS_ROOT` / `allow_root_catchall` combo is the only path that can land mail as uid 0, and it fires a `LoadWarning::RootCatchallAccepted` on every `Config::load` so the elevation is persistently audible.

## Code Quality

- **[Non-blocker]** `prompt_mailbox_owner` and its helpers (`MAX_OWNER_PROMPT_ATTEMPTS`, `NONINTERACTIVE_ENV`, `is_noninteractive_env`) are all gated on `#[allow(dead_code)]` because nothing in the released binary calls them. Marking a new public `fn` dead-code on the day it lands is a code smell — either wire it into the one Sprint-3 caller (even for the catchall, to future-proof when non-`aimx-catchall` catchall owners become possible) or hold the helper until Sprint 5+ where it will have a real caller. Shipping dead code today means Sprint 5+ has no way to tell if the helper still matches intent when they finally use it.
- **[Non-blocker]** `legacy_template_phase_is_absent_from_setup` reads `include_str!("setup.rs")` to grep the source. This is the same trick Sprint 3's grep-AC uses, so it's consistent with the plan, but it fires on the *compiled* `setup.rs` (including the test body itself) — the "built-at-runtime needle" workaround (`let fn_kw = "pub(crate) fn "`) is there to avoid self-matching. That's clever but fragile: a future test that literally writes `"pub(crate) fn configure_hook_something"` in a string literal would trip the assertion. A stronger guard would use `syn` / `rust-analyzer` or just keep the `grep` check in CI rather than in-source.

## Alignment with PRD

PRD §6.8 specifies a "new mailbox-owner prompt during the setup flow for each configured mailbox." The implementer's flagged concern is real: today's `run_setup` flow never touches `prompt_mailbox_owner`, only the hard-coded catchall.

I read the S3-2 AC ("`run_setup` calls `prompt_mailbox_owner` for each mailbox it creates") as **vacuously satisfied** rather than missed, because:

1. `finalize_setup` only ever creates the catchall mailbox, whose owner is PRD-mandated to be `aimx-catchall` (§6.2: "Other owners on a catchall entry are rejected hard"). Prompting for the catchall's owner would contradict §6.2.
2. PRD §6.8 says "Re-entrant setup (detected via `is_already_configured`) skips mailbox creation entirely; operators use `aimx mailboxes create` for additions." So the only code path in the whole codebase that creates a non-catchall mailbox interactively is `aimx mailboxes create`, which is owned by Sprint 2 + later re-plumbing.
3. The PR does not add a new "which mailboxes do you want to create?" phase to `run_setup`. That phase would be needed if S3-2 intended operator-driven additional-mailbox creation during setup. It isn't specified in the sprint plan, the PR, or the PRD text beyond the one sentence in §6.8 — and that sentence is incompatible with the re-entrant-setup rule two paragraphs later.

So I'm treating this as **not a blocker for Sprint 3**, but flagging a **consistency gap for Sprint 5+ / Sprint 6**: when `aimx mailboxes create` or `aimx agent-setup` get re-plumbed through this helper, please confirm `prompt_mailbox_owner` is the right seam and wire it up. If Sprint 5 ends without a live caller, revisit whether the helper should exist at all.

The four explicit PR-body bullet points map cleanly to PRD §6.4 / §6.8 and all land the stated behavior. No other PRD divergences noted.

## Human Review Comments

No human comments on the PR or any linked issue at review time.

- **Overall verdict:** Needs minor fixes (but none are hard blockers — see below)
- **Blockers:** None.
- **Non-blockers:**
  1. `validate_mailbox_owners` docstring (src/config.rs:1113) claims `allow_root_catchall` on a catchall with a non-root owner is rejected. The code accepts that combination silently. Fix the docstring to match the code (or add the check — either works).
  2. `run_setup_skips_catchall_user_when_no_catchall_configured` (src/setup.rs:2344-2348) has a misleading inline comment describing `owner = "root"` and a load-failure path that don't match the actual fixture (`owner = "ops"`). Update the comment.
  3. `prompt_mailbox_owner` is `#[allow(dead_code)]` with no live caller. Either wire it into `aimx mailboxes create`'s `resolve_create_owner` (even as the on-miss interactive branch) or explicitly defer the helper to the Sprint that actually uses it, with a tracking note in the sprint file.
- **Nice-to-haves:**
  - Drop `non_interactive` from `run_setup`'s signature if no future sprint will reintroduce an interactive phase in setup.
  - Add a real scripted-stdin test for `prompt_mailbox_owner`'s re-prompt loop when the helper gets its first caller.

The S3-1 / S3-3 / S3-4 work is solid. S3-2 lands the helper without wiring it — flagging for discussion rather than blocking the merge; the pre-launch track has Sprint 5+ to close the loop.

uzyn:
Addressed the three flagged non-blockers in 5c3aa75:

1. **Docstring drift on `allow_root_catchall`** (`src/config.rs`): picked option (a). `validate_mailbox_owners` docstring now explicitly says the flag is a no-op when owner is not root on a catchall, and is only rejected when set on a non-catchall mailbox. Matches the code and PRD §6.2 intent (flag gates only the root escape hatch).

2. **Stale inline comment** (`src/setup.rs` `run_setup_skips_catchall_user_when_no_catchall_configured`): rewrote the comment to describe the actual `owner = "ops"` orphan-warn path — `load_ignore_warnings` happily loads the mailbox (orphan owner is a warning not an error), and `config_has_catchall` returns false because no entry is `*@<domain>`.

3. **`prompt_mailbox_owner` dead code**: picked option (a). Wired it into `aimx mailboxes create` as the on-miss interactive branch. Flow:
   - `--owner <user>` still wins when supplied.
   - When omitted, the CLI now calls `setup::prompt_mailbox_owner(address, &RealSystemOps)` — operator sees the prompt defaulted to the local part when `getpwnam(local_part)` resolves; invalid input re-prompts up to `MAX_OWNER_PROMPT_ATTEMPTS = 5` times.
   - Under `AIMX_NONINTERACTIVE=1` the helper accepts the local-part default when available or errors hard with a `useradd` hint, so scripted callers fail fast rather than block on stdin.
   - Dropped `#[allow(dead_code)]` from `MAX_OWNER_PROMPT_ATTEMPTS`, `NONINTERACTIVE_ENV`, `is_noninteractive_env`, and `prompt_mailbox_owner`.
   - Added four unit tests in `mailbox.rs::tests` for `resolve_create_owner`: explicit-wins, explicit-empty-errors, noninteractive-uses-local-part-default, noninteractive-errors-when-no-default.
   - Updated `book/cli.md` and clap help for `--owner` to describe the new prompt + noninteractive behavior.

Verification:
- `cargo test`: 1094 unit + 97 integration pass (0 failed, 3 ignored by design).
- `services/verifier`: `cargo fmt --check` + `cargo clippy -- -D warnings` clean.

Nice-to-haves (deliberately not taken in this patch; flag for a follow-up sprint):
- The `non_interactive` param still sits on `run_setup`'s signature for CLI compatibility.
- A scripted-stdin test for `prompt_mailbox_owner`'s re-prompt loop would strengthen coverage now that the helper has a live caller.

uzyn (commented):
## Re-Review: All 3 Non-Blockers Resolved

The implementer addressed every flagged item from my first-round review in commit 5c3aa75. I verified each fix independently against the diff and test outputs. No new issues found.

### Resolved Issues

1. **Docstring drift on `validate_mailbox_owners`** — FIXED at `src/config.rs:1109-1117`. The docstring now says `allow_root_catchall = true` on a non-catchall is rejected hard, and on a catchall whose owner is not `root` the flag is a no-op and silently accepted. That matches the code and matches PRD §6.2's intent (the flag only gates the root escape hatch).

2. **Stale inline comment in `run_setup_skips_catchall_user_when_no_catchall_configured`** — FIXED at `src/setup.rs:2338-2343`. The comment now correctly describes the orphan-warn `owner = \"ops\"` load path and explains why `config_has_catchall` returns `false` (no wildcard entry) — matching the actual fixture.

3. **`prompt_mailbox_owner` dead code** — FIXED. The helper is now wired into `aimx mailboxes create` via `mailbox::resolve_create_owner` (`src/mailbox.rs:219-220, 254-268`). All four `#[allow(dead_code)]` attributes (on `MAX_OWNER_PROMPT_ATTEMPTS`, `NONINTERACTIVE_ENV`, `is_noninteractive_env`, `prompt_mailbox_owner`) are gone. Four new unit tests in `src/mailbox.rs:690-758` cover the explicit-wins, explicit-empty, noninteractive-default, and noninteractive-no-default branches. Book (`book/cli.md:111-113`) and clap help (`src/cli.rs:186-193`) updated to describe the new prompt + `AIMX_NONINTERACTIVE=1` behavior.

### Verification

- All 32 `mailbox::tests::*` pass (32 passed; 0 failed)
- All 238 `setup::tests::*` pass (238 passed; 0 failed)
- The 4 new `resolve_create_owner_*` tests pass
- No new `#[allow(dead_code)]` markers added anywhere

### Notes (nice-to-haves for future sprints — not blocking)

- The nice-to-have from round 1 — adding a real scripted-stdin test for `prompt_mailbox_owner`'s re-prompt loop — still stands; today only the non-interactive branches are exercised directly. The interactive branch is reachable via `aimx mailboxes create` now, so a test using a mock `SystemOps` that fails `user_exists` a few times and a scripted stdin would lock that behavior in. Non-blocking; recommend Sprint 6 or Sprint 7 pick it up when `aimx agent-setup` becomes the second live caller (per the helper's doc).
- Behavioral shift worth flagging for the release notes: previously, `aimx mailboxes create alice` (no `--owner`) on a host where `alice` does not exist errored immediately with a `useradd` hint. Now, on a TTY, it prompts interactively; under `AIMX_NONINTERACTIVE=1` it errors with a similar hint. On a piped/closed stdin *without* `AIMX_NONINTERACTIVE=1` the prompt loop burns its 5 attempts (each `read_line` returns EOF immediately) before erroring. Functionally still fails; noisier output but not a regression.

### Verdict

**Ready to merge.** All three non-blockers are resolved cleanly with tests, clippy is clean, and the `aimx mailboxes create` change is a genuine UX improvement over the previous hard-fail-with-useradd-hint path while preserving the scripted-install contract via `AIMX_NONINTERACTIVE=1`.

```
[Sprint 3] Catchall narrowing + setup refactor (#127)

Delivers the four Sprint 3 stories for the agent-integration track:

- S3-1: rename `ensure_hook_user` -> `ensure_catchall_user`; gate on
  configured catchall mailbox only. The `aimx-catchall` system user is
  now created with `--system`, `/usr/sbin/nologin`, home
  `/var/lib/aimx-catchall` (0700 owned by aimx-catchall). Section title
  flips `[User]` -> `[Catchall]`.

- S3-2: delete `chown_datadir_for_hook_user` (per-mailbox chown now
  lives in `mailbox_handler`). Add `prompt_mailbox_owner` with
  local-part default, bounded re-prompt, and `AIMX_NONINTERACTIVE=1`
  fallback. Wired into `aimx mailboxes create` as the on-miss
  interactive branch (PRD §6.8). Catchall owner stays hard-coded to
  `aimx-catchall` by `finalize_setup`.

- S3-3: delete `configure_hook_templates` and every helper it called;
  remove the \"Hook Templates\" checkbox section. `aimx hooks list
  --templates` and `aimx doctor` point at `aimx agent-setup <agent>`.

- S3-4: `Config::load` rejects `owner = \"root\"` on any non-catchall
  mailbox with PRD §6.2 wording; new `MailboxConfig.allow_root_catchall`
  (default false, omitted from serialized TOML) gates the catchall
  escape hatch and emits `LoadWarning::RootCatchallAccepted` so the
  elevation is audit-logged.

Test fixtures across the crate swapped from `owner = \"root\"` to `owner
= \"ops\"` where root was used as a convenient default. Sprint 1's
`load_accepts_mailbox_with_owner_root` flipped to assert the new
rejection; new S3-4 tests cover every branch.

cargo fmt + cargo clippy --all-targets -- -D warnings clean;
1094 unit + 97 integration tests pass.
```

PR #128, opened 2026-04-22
[Sprint 4] SO_PEERCRED UDS authentication

uzyn:
## Summary

- Every UDS verb now authorizes the caller via `SO_PEERCRED` against the resolved mailbox owner (or requires root for `MAILBOX-CREATE` / `MAILBOX-DELETE`).
- Cross-user attacks (bob `SEND`ing as alice, `MARK-READ`ing alice's mailbox, creating hooks on alice's mailbox) return `AIMX/1 ERR EACCES` with a structured `Code: EACCES` header.
- Root bypasses every ownership check and each bypass emits a `tracing::info!` line under `aimx::uds` for audit.
- Implements PRD §6.5 in full: EACCES, ENOENT, root_bypass logging, and a cross-user root-gated integration test.

## Stories

- **S4-1** — `Caller { uid, gid, pid, username }` plumbed from `accept()` into every handler (`send_handler`, `state_handler`, `mailbox_handler`, `hook_handler`). Username is lazy-memoized via `getpwuid`.
- **S4-2** — `require_mailbox_owner_or_root` / `require_root` helpers in new `src/uds_authz.rs`. `HOOK-CREATE` authz runs after mailbox resolution; unknown mailbox → `ENOENT` (not EACCES, avoiding the existence side-channel).
- **S4-3** — `ErrCode` gains `Eaccess` / `Enoent`. Error frames carry both the legacy inline token and an explicit `Code:` header on the following line; existing single-line clients continue to parse. `tracing-test`-backed assertions for every authz decision.
- **S4-4** — `tests/uds_authz.rs` spawns a real `aimx serve` subprocess, issues raw `AIMX/1` frames via `runuser -u aimx-it-bob python3`, and asserts EACCES on SEND / MARK-READ / HOOK-CREATE. Root control emits `root_bypass`. `Drop` guard kills the daemon + userdels fixtures on scope exit. CI `integration-isolation` job extended.

## Test plan

- [x] `cargo test` — 1207 tests pass across bin + integration + uds_authz
- [x] `cargo fmt -- --check`
- [x] `sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 cargo test --test uds_authz -- --ignored` — bob-as-alice attacks each return EACCES, root control emits `root_bypass`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,15 @@ jobs:
         run: |
           sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 cargo test --test isolation -- --ignored
 
+      # Sprint 4 S4-4: root-gated UDS authz integration test. Reuses
+      # the `aimx-it-alice` / `aimx-it-bob` fixture from Sprint 2 and
+      # asserts that bob cannot SEND/MARK-READ/HOOK-CREATE as alice
+      # over the UDS. Same runtime gating (AIMX_INTEGRATION_SUDO=1 +
+      # sudo + PATH propagation) as the isolation test above.
+      - name: Run UDS authz integration test (root)
+        run: |
+          sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 cargo test --test uds_authz -- --ignored
+
   verifier-tests:
     runs-on: ubuntu-latest
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,23 @@ jobs:
         run: |
           sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 cargo test --test uds_authz -- --ignored
 
+      # Sprint 4 review NB-1: the five MAILBOX-CRUD UDS integration
+      # tests in `tests/integration.rs` are `#[ignore]`-gated because
+      # `MAILBOX-CREATE` / `MAILBOX-DELETE` are root-only (PRD §6.5)
+      # and the daemon chowns freshly-created mailbox dirs to the
+      # configured owner — a non-root `cargo test` cannot exercise the
+      # code path. This step runs them explicitly under sudo, one name
+      # per `--exact`, so any future skip regresses here rather than
+      # silently.
+      - name: Run MAILBOX-CRUD integration tests (root)
+        run: |
+          sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 cargo test --test integration -- --ignored --exact \
+            mailbox_create_via_uds_hotswaps_config_and_routes_new_mail \
+            mailbox_delete_via_uds_refuses_nonempty_and_succeeds_after_cleanup \
+            mailbox_delete_force_yes_wipes_contents_and_succeeds \
+            mailbox_delete_force_without_yes_prompts_and_aborts_on_n \
+            concurrent_mailbox_create_and_ingest_does_not_deadlock
+
   verifier-tests:
     runs-on: ubuntu-latest
     defaults:
--- a/src/hook_handler.rs
+++ b/src/hook_handler.rs
@@ -41,7 +41,7 @@ use crate::send_protocol::{
     AckResponse, ErrCode, HookCreateRequest, HookDeleteRequest, HookTemplateCreateBody,
 };
 use crate::state_handler::StateContext;
-use crate::uds_authz::{Caller, enforce_mailbox_owner_or_root, log_decision};
+use crate::uds_authz::{Caller, LogDecision, enforce_mailbox_owner_or_root, log_decision};
 
 /// Handle an `AIMX/1 HOOK-CREATE` request. Template-only. Takes the
 /// per-mailbox write lock for the addressed mailbox (outer) plus
@@ -106,7 +106,7 @@ pub async fn handle_hook_create(
                 "HOOK-CREATE",
                 caller,
                 Some(&req.mailbox),
-                "reject",
+                LogDecision::Reject,
                 Some("mailbox not found"),
             );
             return AckResponse::Err {
@@ -292,7 +292,7 @@ pub async fn handle_hook_delete(
                 "HOOK-DELETE",
                 caller,
                 None,
-                "reject",
+                LogDecision::Reject,
                 Some("hook not found"),
             );
             return AckResponse::Err {
@@ -305,10 +305,24 @@ pub async fn handle_hook_delete(
     // Sprint 4 §6.5: the caller must own the target mailbox OR be
     // root. Runs before the origin check so a non-owner never learns
     // whether the hook is MCP- or operator-origin.
-    if let Some(mailbox_cfg) = current.mailboxes.get(&owner)
-        && let Err(reject) =
-            enforce_mailbox_owner_or_root("HOOK-DELETE", caller, &owner, mailbox_cfg)
-    {
+    //
+    // `owner` came from the same `current.mailboxes` iteration above
+    // so the `None` branch is unreachable today. Handle it explicitly
+    // with `ENOENT` so authz can't be silently skipped if a future
+    // refactor separates the owner-scan and lookup steps.
+    let mailbox_cfg = match current.mailboxes.get(&owner) {
+        Some(m) => m,
+        None => {
+            return AckResponse::Err {
+                code: ErrCode::Enoent,
+                reason: format!(
+                    "mailbox '{owner}' resolved but not found in config \
+                     (race with concurrent MAILBOX-DELETE)"
+                ),
+            };
+        }
+    };
+    if let Err(reject) = enforce_mailbox_owner_or_root("HOOK-DELETE", caller, &owner, mailbox_cfg) {
         return AckResponse::Err {
             code: reject.code,
             reason: reject.reason,
--- a/src/hook_handler.rs
+++ b/src/hook_handler.rs
@@ -41,6 +41,7 @@ use crate::send_protocol::{
     AckResponse, ErrCode, HookCreateRequest, HookDeleteRequest, HookTemplateCreateBody,
 };
 use crate::state_handler::StateContext;
+use crate::uds_authz::{Caller, enforce_mailbox_owner_or_root, log_decision};
 
 /// Handle an `AIMX/1 HOOK-CREATE` request. Template-only. Takes the
 /// per-mailbox write lock for the addressed mailbox (outer) plus
@@ -50,6 +51,7 @@ pub async fn handle_hook_create(
     state_ctx: &StateContext,
     mb_ctx: &MailboxContext,
     req: &HookCreateRequest,
+    caller: &Caller,
 ) -> AckResponse {
     // --- Parse JSON body (rejects `cmd`, `run_as`, etc. via `deny_unknown_fields`).
     let body: HookTemplateCreateBody = match serde_json::from_slice(&req.body) {
@@ -94,11 +96,32 @@ pub async fn handle_hook_create(
 
     let current = mb_ctx.config_handle.load();
 
-    // --- Resolve mailbox.
-    if !current.mailboxes.contains_key(&req.mailbox) {
+    // --- Resolve mailbox. Sprint 4 §6.5: unknown mailbox returns
+    // `ENOENT` (not `EACCES`) so the authz check itself cannot leak
+    // which mailboxes exist. Authz runs after the mailbox is found.
+    let mailbox_cfg = match current.mailboxes.get(&req.mailbox) {
+        Some(m) => m,
+        None => {
+            log_decision(
+                "HOOK-CREATE",
+                caller,
+                Some(&req.mailbox),
+                "reject",
+                Some("mailbox not found"),
+            );
+            return AckResponse::Err {
+                code: ErrCode::Enoent,
+                reason: format!("mailbox '{}' does not exist", req.mailbox),
+            };
+        }
+    };
+
+    if let Err(reject) =
+        enforce_mailbox_owner_or_root("HOOK-CREATE", caller, &req.mailbox, mailbox_cfg)
+    {
         return AckResponse::Err {
-            code: ErrCode::Mailbox,
-            reason: format!("mailbox '{}' does not exist", req.mailbox),
+            code: reject.code,
+            reason: reject.reason,
         };
     }
 
@@ -242,6 +265,7 @@ pub async fn handle_hook_delete(
     state_ctx: &StateContext,
     mb_ctx: &MailboxContext,
     req: &HookDeleteRequest,
+    caller: &Caller,
 ) -> AckResponse {
     if !is_valid_hook_name(&req.name) {
         return AckResponse::Err {
@@ -264,12 +288,32 @@ pub async fn handle_hook_delete(
     let owner = match owner {
         Some(n) => n,
         None => {
+            log_decision(
+                "HOOK-DELETE",
+                caller,
+                None,
+                "reject",
+                Some("hook not found"),
+            );
             return AckResponse::Err {
-                code: ErrCode::NotFound,
+                code: ErrCode::Enoent,
                 reason: format!("hook '{}' not found", req.name),
             };
         }
     };
+
+    // Sprint 4 §6.5: the caller must own the target mailbox OR be
+    // root. Runs before the origin check so a non-owner never learns
+    // whether the hook is MCP- or operator-origin.
+    if let Some(mailbox_cfg) = current.mailboxes.get(&owner)
+        && let Err(reject) =
+            enforce_mailbox_owner_or_root("HOOK-DELETE", caller, &owner, mailbox_cfg)
+    {
+        return AckResponse::Err {
+            code: reject.code,
+            reason: reject.reason,
+        };
+    }
 
     // Origin check: before acquiring any locks, refuse operator-origin
     // hooks up front so callers get a precise error without waiting on
@@ -488,7 +532,7 @@ mod tests {
             name: Some("my_hook".into()),
             body: body(&[("prompt", "hello world")]),
         };
-        match handle_hook_create(&state_ctx, &mb_ctx, &req).await {
+        match handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await {
             AckResponse::Ok => {}
             other => panic!("expected Ok, got {other:?}"),
         }
@@ -522,7 +566,7 @@ mod tests {
             body: body(&[("prompt", "anon")]),
         };
         assert!(matches!(
-            handle_hook_create(&state_ctx, &mb_ctx, &req).await,
+            handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await,
             AckResponse::Ok
         ));
         let live = mb_ctx.config_handle.load();
@@ -542,7 +586,7 @@ mod tests {
             name: None,
             body: json.to_vec(),
         };
-        match handle_hook_create(&state_ctx, &mb_ctx, &req).await {
+        match handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await {
             AckResponse::Err { code, reason } => {
                 assert_eq!(code, ErrCode::Validation);
                 assert!(reason.contains("cmd"), "{reason}");
@@ -566,7 +610,7 @@ mod tests {
             name: None,
             body: json.to_vec(),
         };
-        match handle_hook_create(&state_ctx, &mb_ctx, &req).await {
+        match handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await {
             AckResponse::Err { code, reason } => {
                 assert_eq!(code, ErrCode::Validation);
                 assert!(reason.contains("run_as"), "{reason}");
@@ -588,7 +632,7 @@ mod tests {
             name: None,
             body: json.to_vec(),
         };
-        match handle_hook_create(&state_ctx, &mb_ctx, &req).await {
+        match handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await {
             AckResponse::Err { code, reason } => {
                 assert_eq!(code, ErrCode::Validation);
                 assert!(reason.contains("dangerously_support_untrusted"), "{reason}");
@@ -609,7 +653,7 @@ mod tests {
             name: None,
             body: body(&[]),
         };
-        match handle_hook_create(&state_ctx, &mb_ctx, &req).await {
+        match handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await {
             AckResponse::Err { code, reason } => {
                 assert_eq!(code, ErrCode::Validation);
                 assert!(reason.contains("unknown-template"), "{reason}");
@@ -631,7 +675,7 @@ mod tests {
             name: None,
             body: body(&[]),
         };
-        match handle_hook_create(&state_ctx, &mb_ctx, &req).await {
+        match handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await {
             AckResponse::Err { code, reason } => {
                 assert_eq!(code, ErrCode::Validation);
                 assert!(reason.contains("missing-param"), "{reason}");
@@ -653,7 +697,7 @@ mod tests {
             name: None,
             body: body(&[("prompt", "hi"), ("bogus", "zz")]),
         };
-        match handle_hook_create(&state_ctx, &mb_ctx, &req).await {
+        match handle_hook_create(&state_ctx, &mb_ctx, &req, &Caller::internal_root()).await {
             AckResponse::Err { code, reason } => {
                 assert_eq!(code, ErrCode::Validation);
                 assert!(reason.contains("unknown-param"), "{reason}");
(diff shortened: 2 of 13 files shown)


uzyn (commented):
## Sprint Goal Assessment

Sprint 4's goal — SO_PEERCRED-backed per-verb UDS authorization with EACCES on cross-user attacks, ENOENT on leak-free unknown-target shapes, and info-level `root_bypass` audit logs — is achieved. The `Caller` type is threaded from accept through every handler, the PRD §6.5 authz table is encoded in `src/uds_authz.rs` as `require_mailbox_owner_or_root` / `require_root`, the wire protocol gains a `Code:` header that co-exists cleanly with the legacy inline token, and `tests/uds_authz.rs` drives real `aimx serve` with bob-as-alice raw frames on SEND / MARK-READ / HOOK-CREATE. All 4 stories' acceptance criteria are met and all 3 CI jobs pass.

**[S4-1] Ucred plumbing: accept loop → every handler**
- [x] Accept loop calls `peer_cred()`; errors return `ERR code=EACCES reason="peer cred unavailable"` before any body read (`src/serve.rs:747-764` via `caller_from_stream`)
- [x] `Caller { uid, gid, pid, username: OnceLock<Option<String>> }` in new `src/uds_authz.rs`
- [x] Every handler (`send_handler`, `state_handler`, `mailbox_handler`, `hook_handler`) accepts `&Caller`
- [x] Compilation clean — every call site threaded
- [x] `Caller::username` memoized; `getpwuid` miss → `None`; uid 0 → `Some("root")` via fast-path
- [x] Unit tests for fake caller + handlers (e.g. `serve.rs` mod tests; `uds_authz.rs` mod tests)
- [x] `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt -- --check` clean

**[S4-2] Per-verb authz in handlers**
- [x] Helpers live in `src/uds_authz.rs`
- [x] Every handler enforces PRD §6.5 row; unauthorized returns `ERR code=EACCES`
- [x] Root bypass emits `tracing::info!` under `aimx::uds` with structured fields
- [x] Mailbox-not-found returns `ErrCode::Enoent` not `Eaccess` (HOOK-CREATE and MARK-* both verified)
- [x] Unit tests: owner accept / non-owner EACCES / root bypass + info / unknown mailbox ENOENT
- [x] `cargo test`, `cargo clippy`, `cargo fmt` clean

**[S4-3] ErrResponse Code header + structured aimx::uds logs**
- [x] `ErrCode` gains `Eaccess` / `Enoent` + `from_str` reverse parser
- [x] `write_response` / `write_ack_response` emit both legacy inline token and `Code:` header; terminator is `\n\n`
- [x] Every authz-rejecting path constructs with the correct `code`
- [x] tracing events: debug (accept), info (root_bypass), warn (reject) under `aimx::uds`
- [x] Client-side `parse_response_line` (send.rs) + `parse_ack_response` (mcp.rs) read the `Code:` header; header takes precedence over inline token
- [x] `tracing-test`-backed assertions for accept / root_bypass / reject decisions

**[S4-4] Cross-user UDS integration tests**
- [x] `tests/uds_authz.rs` root-gated (`#[ignore]` + `AIMX_INTEGRATION_SUDO=1`)
- [x] Three bob-as-alice attacks each return `ERR EACCES`: SEND with `From: alice`, MARK-READ alice, HOOK-CREATE alice
- [x] Root control succeeds + `root_bypass` info log asserted from captured serve stderr
- [x] CI `integration-isolation` job extended with a second sudo step
- [x] Drop-based teardown kills serve + userdels even on assertion failure

## Test Coverage

The Sprint 4 story tests are thorough: unit coverage in `src/uds_authz.rs::tests` (accept, reject, root_bypass, username-lookup, Clone-preserves-cache), tracing-test-backed log assertions, and a real cross-user integration test that goes all the way through `aimx serve` over `/run/aimx/aimx.sock`.

**One real coverage gap** is called out below under Non-blockers (the `skip_if_mailbox_crud_not_root` gate is never tripped the other way in CI — see #1).

## Potential Bugs

### Non-blockers

**NB-1 — `skip_if_mailbox_crud_not_root()` permanently skips 5 mailbox-CRUD tests in CI.** (tests/integration.rs:514-538). The CI `integration-isolation` job only runs `cargo test --test isolation -- --ignored` and `cargo test --test uds_authz -- --ignored`. It does **not** run `cargo test --test integration` under sudo. The core-tests job runs `cargo test` as `runner` (non-root). Combined, these five tests never actually exercise the code path they claim to cover:

- `mailbox_create_via_uds_hotswaps_config_and_routes_new_mail`
- `mailbox_delete_via_uds_refuses_nonempty_and_succeeds_after_cleanup`
- `mailbox_delete_force_yes_wipes_contents_and_succeeds`
- `mailbox_delete_force_without_yes_prompts_and_aborts_on_n`
- `concurrent_mailbox_create_and_ingest_does_not_deadlock`

Either add a sudo-wrapping CI step that invokes `cargo test --test integration <names> -- --exact` (after marking them `#[ignore]`), or add `--test integration` to the existing integration-isolation job with a filter. As-is, Sprint 4 silently traded previously-passing coverage for permanently-skipped coverage on the MAILBOX-CRUD verbs. The `uds_authz.rs` integration test does not cover MAILBOX-CRUD, so there is no substitute.

**NB-2 — `Caller::new()` and `Caller::internal_root()` use `#[allow(dead_code)]` instead of `#[cfg(test)]`.** (src/uds_authz.rs:68, 82). Every call site in the codebase is inside a `#[cfg(test)]` block or a test file. Gating the constructors behind `#[cfg(test)]` (or a dedicated `test-helpers` feature) tightens the production API and removes the dead-code allow. Purely hygienic — no correctness risk.

**NB-3 — Authz silently skipped on `if let Some(mailbox_cfg) = ...` fallthrough.** Three sites use the pattern:

- `send_handler.rs:182-194` (SEND)
- `hook_handler.rs:307-315` (HOOK-DELETE)
- `state_handler.rs:143-150` (MARK)

All three have already established that the mailbox exists earlier in the function, so the `None` branch is unreachable **today**. But if a future refactor introduces a race between the "does mailbox exist" check and the `config.mailboxes.get(&name)` call, authz is quietly skipped instead of returning an error. A defensive `else` branch returning `ErrCode::Enoent` (or an internal `unreachable!`/`debug_assert!`) would be safer. Not a live bug.

**NB-4 — `log_decision` fallthrough to debug on unknown decision string.** (src/uds_authz.rs:225-254). The match on `decision` has explicit arms for `"reject"` and `"root_bypass"`, and `_` routes to `tracing::debug!("uds authz accept")`. If a future caller passes an unexpected string, it's logged as "accept" at debug level — potentially masking a misuse. An explicit `"accept" =>` arm plus a `_ => tracing::error!` wildcard would be tighter.

## Security Issues

None. The authz table matches PRD §6.5 exactly:
- Non-root callers whose username doesn't match the mailbox owner get `EACCES` on SEND, MARK-*, HOOK-CREATE, HOOK-DELETE.
- `MAILBOX-CREATE` / `MAILBOX-DELETE` are root-only.
- Unknown mailboxes return `ENOENT` before any authz check, preventing existence-leak side-channels.
- Root bypasses all mailbox-ownership checks and every bypass is audited under `aimx::uds` at info level.
- `getpwuid(0)` is skipped via a uid-zero fast-path, so a broken local NSS that returns null for uid 0 can't make the daemon reject root as "unresolvable username".
- Caller with a resolvable uid but no passwd entry correctly rejects with `EACCES` rather than defaulting to "owner-match" fallthrough.
- `OnceLock` is non-poisoning in `std::sync`; `lookup_username` returns `Option<String>` and never panics, so the lazy getpwuid miss cannot poison or crash the handler.

## Code Quality

The `src/uds_authz.rs` module is well-scoped: `Caller` + `lookup_username` + `require_*` + `enforce_*` + `log_decision`. Docstrings explicitly point back at PRD §6.5. The `#[cfg(test)] fn _hashmap_sanity` at the bottom is pure noise (noted in its own comment) — worth deleting on a follow-up.

## Alignment with PRD

PRD §6.5 is implemented in full. `TEMPLATE-CREATE` / `TEMPLATE-DELETE` are explicitly deferred to Sprint 5 (not a gap for Sprint 4).

**Overall verdict:** Needs minor fixes (CI coverage).

**Blockers:** None.

**Non-blockers:**
- NB-1 — `integration-isolation` CI job doesn't run the 5 root-gated `MAILBOX-CRUD` tests; they skip in every configuration. Add a sudo-wrapping step for `cargo test --test integration` with the appropriate filter, or restructure the gate.
- NB-2 — Gate `Caller::new` / `Caller::internal_root` behind `#[cfg(test)]` instead of `#[allow(dead_code)]`.
- NB-3 — Tighten the three `if let Some(mailbox_cfg) = ...` authz sites so authz can't be silently skipped on a future refactor.
- NB-4 — Make `log_decision`'s match explicit for `"accept"` and log a warning on unexpected values.

**Nice-to-haves:**
- Delete `_hashmap_sanity` from `src/uds_authz.rs::tests`.

None of these block Sprint 4 from shipping functionally — the authz layer works, the integration test proves it end-to-end, and all CI passes. NB-1 is the most significant because it erodes long-term confidence in the mailbox-CRUD path. Worth addressing in-sprint or the Sprint 7.5 cleanup.

uzyn:
Review feedback addressed in c1548d3. Summary of fixes:

## NB-1 — CI now runs the 5 MAILBOX-CRUD UDS integration tests

Marked all five tests `#[ignore]` with an explanatory `reason` string and added a new CI step (`Run MAILBOX-CRUD integration tests (root)`) that invokes them under sudo via `cargo test --test integration -- --ignored --exact <name1> ... <name5>`. Each test is listed explicitly so any future rename or removal surfaces here rather than silently regressing.

Coverage that was previously invisible is now exercised on every push:
- `mailbox_create_via_uds_hotswaps_config_and_routes_new_mail`
- `mailbox_delete_via_uds_refuses_nonempty_and_succeeds_after_cleanup`
- `mailbox_delete_force_yes_wipes_contents_and_succeeds`
- `mailbox_delete_force_without_yes_prompts_and_aborts_on_n`
- `concurrent_mailbox_create_and_ingest_does_not_deadlock`

## NB-2 — `#[cfg(test)]` on `Caller::new` / `Caller::internal_root`

Replaced `#[allow(dead_code)]` with `#[cfg(test)]`. Verified every call site lives inside a `#[cfg(test)] mod tests` block (serve.rs, state_handler.rs, mailbox_handler.rs, hook_handler.rs, send_handler.rs, and uds_authz.rs itself). Production surface is now tighter.

## NB-3 — Tightened three authz sites

`send_handler.rs`, `hook_handler.rs`, and `state_handler.rs` no longer use `if let Some(mailbox_cfg) = ...` for the authz lookup. Each is a `match` with an explicit `None` arm returning a concrete error (`ENOENT` for state/hook, `MAILBOX` for send) with a reason that names a concurrent-MAILBOX-DELETE race as the only plausible trigger. A future refactor can't silently skip authz.

## NB-4 — Typed `LogDecision` enum

`log_decision`'s `decision: &str` parameter is now `decision: LogDecision` with variants `Accept`, `RootBypass`, `Reject`. The compiler rejects unknown decisions; the implicit "debug accept" wildcard arm that previously masked misuse is gone. `LogDecision::as_str` preserves the existing wire-format strings so log-based assertions (`tracing-test`) keep working unchanged.

## Nice-to-have — removed `_hashmap_sanity`

Deleted the leftover helper and its dangling `use std::collections::HashMap` from `src/uds_authz.rs::tests`.

## Verification

- `cargo test` — 1110 bin + 92 integration + 1 uds_authz passing; 5 MAILBOX-CRUD + 1 isolation + 1 uds_authz cross-user + 3 `#[ignore]` already-root-gated tests are correctly ignored locally. All three will run under the new and existing root-gated CI steps.

uzyn (commented):
# Re-review — Sprint 4 (agent-integration track)

Re-reviewed commit `c1548d3` against the five items flagged in the prior round.

## Resolved issues

### NB-1 — CI coverage gap for mailbox-CRUD tests — **RESOLVED**
- New CI step `Run MAILBOX-CRUD integration tests (root)` added to the `integration-isolation` job (`.github/workflows/ci.yml:79-94`), directly below the existing Sprint 2 isolation and Sprint 4 `uds_authz` root-gated steps. Uses the same `sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1` pattern.
- All 5 test functions are `#[ignore]`-gated with a descriptive `reason` string (`tests/integration.rs` lines 3082, 3209, 3314, 3376, 3775). `cargo test --test integration` cleanly reports `92 passed; 5 ignored`, and `cargo test --test integration -- --ignored --list --exact <5 names>` resolves to exactly 5 matches — so a future rename would cause the CI step to fail "no tests matched" instead of silently skipping.
- Each test is listed one-per-line under `--exact`, exactly as requested.

### NB-2 — `#[allow(dead_code)]` on test-only helpers — **RESOLVED**
- `Caller::new` and `Caller::internal_root` are now `#[cfg(test)]` (`src/uds_authz.rs:68-74` and `:86-93`) with explanatory doc-comments.
- I independently verified every call site — every one of the ~80 references to `Caller::new` / `Caller::internal_root` across `serve.rs`, `state_handler.rs`, `mailbox_handler.rs`, `hook_handler.rs`, `send_handler.rs`, and `uds_authz.rs` lives inside a `#[cfg(test)] mod tests` block (notably `serve.rs:1883` and `:2010` sit under the `mod tests` block that begins at line 1204). `cargo check --all-targets` passes cleanly, confirming the non-test build has no latent dependency on these helpers.

### NB-3 — silent authz skip on `if let Some(mailbox_cfg)` — **RESOLVED**
- `send_handler.rs` (~:194), `hook_handler.rs` (~:308), and `state_handler.rs` (~:140) all refactored from `if let Some(cfg) = ...` to an explicit `match` with a concrete `None` arm.
- `send_handler` returns `ErrCode::Mailbox`, `hook_handler` and `state_handler` return `ErrCode::Enoent`. Each reason string names "race with concurrent MAILBOX-DELETE" so an operator debugging the log knows the only plausible trigger.
- Authz can no longer be silently skipped by a future refactor that accidentally rearranges the resolve/lookup steps.
- Bonus: `state_handler` also flattened the `if let Some(mb_cfg) = &mailbox_cfg { chown_as_owner(...) }` block now that `mailbox_cfg` is unconditional. No functional change but removes parallel dead-code.

### NB-4 — `log_decision` wildcard arm — **RESOLVED**
- New `pub enum LogDecision { Accept, RootBypass, Reject }` in `src/uds_authz.rs:226-243`. The `match` in `log_decision` now has one arm per variant; no wildcard. Adding a fourth outcome would require a new arm.
- `LogDecision::as_str` preserves the exact wire-format strings (`"accept"` / `"root_bypass"` / `"reject"`), so the existing tracing-based assertions (`src/uds_authz.rs:453,464,476,487` and `tests/uds_authz.rs`) continue to match unchanged. I confirmed by inspection of the grep output.

### Nice-to-have — delete `_hashmap_sanity` — **RESOLVED**
- Helper and its dangling `use std::collections::HashMap` are gone from `src/uds_authz.rs::tests` (visible in the diff at `:354` and `:487-494`).

## Still unresolved
None.

## New issues found
None. No regression surface in the diff. `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and the full `cargo test --bin aimx` + `cargo test --test integration` suites all pass (1110 bin + 92 integration with 5 correctly ignored, reproducing the implementer's numbers).

## Summary

All four NB items and the nice-to-have are cleanly resolved. NB-1's CI wiring is correct: the step sits in the `integration-isolation` job alongside the other root-gated steps, invokes the five test names under `--exact` (so a silent skip regression is impossible), and all five names resolve to exactly the `#[ignore]`-gated tests. Ready to merge.

```
[Sprint 4] SO_PEERCRED UDS authentication

Every AIMX/1 verb now authorizes the UDS caller against the resolved
mailbox owner (or requires root for MAILBOX-CREATE / MAILBOX-DELETE)
via SO_PEERCRED. Cross-user attacks return AIMX/1 ERR EACCES with a
structured Code: EACCES header; root bypass is logged at info under
aimx::uds for audit. Unknown mailboxes return ENOENT without leaking
existence to the authz error path. Implements PRD section 6.5 in full,
including a cross-user root-gated integration test.

- S4-1: Caller { uid, gid, pid, username } plumbed from accept() into
  every handler; username lazy-memoized via getpwuid
- S4-2: require_mailbox_owner_or_root / require_root helpers in
  src/uds_authz.rs; HOOK-CREATE authz runs after mailbox resolution
- S4-3: ErrCode gains Eaccess / Enoent; legacy inline token preserved
  alongside new Code: header; tracing-test-backed assertions
- S4-4: tests/uds_authz.rs spawns real aimx serve, issues raw AIMX/1
  frames via runuser; CI integration-isolation job extended

Review follow-ups (c1548d3): CI now runs the five #[ignore]-gated
MAILBOX-CRUD tests under sudo via --exact; Caller::new /
Caller::internal_root gated #[cfg(test)]; three authz sites
refactored from if-let to explicit match so authz cannot be silently
skipped; log_decision now takes a typed LogDecision enum.
```

uzyn:
## CI fix: MAILBOX-CRUD tests failing under sudo

### Root cause
Not a parallelism / port / socket race — the daemon was crashing at startup with:

```
mailbox 'alice' cannot be owned by root; use a regular Linux user or set 'allow_root_catchall = true' on a catchall mailbox
```

`tests::setup_test_env` uses `current_username()` to populate `mailboxes.alice.owner`. The `integration-isolation` step runs as root (sudo), so `geteuid()` returned 0 and `current_username()` returned `"root"`. Sprint 3's `validate_mailbox_owners` then rejected the config via the `owner_is_root && !is_catchall` branch (PRD §6.2).

`start_serve` timed out waiting on the never-bound port and panicked with the uninformative "did not start within 30s" message, which is why the real error was hidden in the CI log. Reproduced locally with a single test under `sudo` — no parallelism involved, the daemon refused to start even solo.

### Fix
- `current_username()` now prefers `$SUDO_USER` when running as root (sudo preserves this env), and falls back to `nobody` (universal, regex-valid, resolvable). Non-root callers are unchanged.
- `start_serve` captures and surfaces daemon stderr on both the 30s timeout and an early-exit detection path, so any future startup regression shows the cause instead of a bare timeout.

### Verified locally
- `sudo -E env PATH=$PATH AIMX_INTEGRATION_SUDO=1 cargo test --test integration -- --ignored --exact <5 names>`: **5 passed in ~5s**.
- Non-root `cargo test`: 92 passed, 5 ignored (root-gated as expected).
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings`: clean.

Commit: 169ca9b

uzyn (commented):
# Re-review cycle 2 — Sprint 4 CI-fix commit `169ca9b`

Re-reviewed the one-commit delta since the previous "Ready to merge" cycle — `169ca9b fix(tests): pick non-root owner when running under sudo`. Diff scope is strictly `tests/integration.rs` (66 lines). No production source changed.

## Verification

### 1. `current_username()` — `$SUDO_USER` fallback is correct

- Under sudo, `$SUDO_USER` is set by sudo and the CI job uses `sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 cargo test ...`, which preserves the invoking user's env. On `ubuntu-latest`, `SUDO_USER=runner` and `getpwnam("runner")` resolves.
- Non-sudo local runs: `uid != 0`, so the new `if uid == 0 { ... }` branch is skipped entirely and behaviour is byte-identical to before except that a `getpwuid` miss now returns `"nobody"` instead of `"root"`. Given Sprint 3 §6.2 rejects `owner = "root"` on non-catchall mailboxes, returning `"nobody"` on the orphan path is a *safer* default than before — no regression.
- Paranoid edge case: if someone runs `sudo -u alice` where alice happens to be a real passwd entry, `SUDO_USER=alice`, which is fine — `getpwnam("alice")` resolves, daemon accepts the config. If someone clobbers `SUDO_USER=""` or `SUDO_USER=root`, the code explicitly skips and falls through to `"nobody"`. Covered.

### 2. `nobody` resolves via `getpwnam` on common distros

- `nobody` is part of the POSIX/Linux base user set: present on Debian/Ubuntu (uid 65534), Alpine, Fedora, Arch, RHEL. `ubuntu-latest` GitHub runners ship it. This is the correct choice for a universal fallback.
- Even if `getpwnam` misses (e.g. a minimal container without `/etc/passwd`), the config load emits `LoadWarning::OrphanMailboxOwner` (non-fatal, from `validate_mailbox_owners` in `src/config.rs:1131-1136`) rather than rejecting — same "best-effort" posture Sprint 3 already chose for exotic runners. `"nobody"` passes the `[a-z_][a-z0-9_-]*[$]?` regex cleanly.

### 3. `start_serve` stderr capture — no deadlock, no leaked children

Both the timeout branch and the early-exit branch are safe:

- **Timeout branch** (`tests/integration.rs:1982-1998`): `let _ = child.kill();` sends SIGKILL (infallible on Unix; kernel cannot block it). The child's stderr write-side closes as soon as the kernel finalises the process. `read_to_string` on the pipe then drains whatever was buffered and returns EOF. No deadlock possible — SIGKILL reaps in milliseconds.
- **Early-exit branch** (`tests/integration.rs:2002-2015`): `try_wait` returns `Some(status)` only when the child has already exited, so the stderr write-side is guaranteed closed; `read_to_string` drains and returns immediately.
- No FD leak: `child` is dropped at the end of the panic unwind (or the implicit end of `start_serve`), which closes the remaining handles. The previous code also relied on the same drop path.
- No stderr-pipe-fill deadlock at startup: tracing is info-level by default and startup log volume is well under the 64KiB Linux pipe buffer before the daemon binds (sub-second window).

### 4. RootBypass is the correct authz path for these tests

- The test subprocess (`aimx mailbox create ... --owner <nobody|SUDO_USER>`) inherits the sudo'd root uid. Its UDS connection's `SO_PEERCRED` reports `uid == 0`.
- `Caller::is_root()` returns `true`, so both `require_mailbox_owner_or_root` (`src/uds_authz.rs:173`) and `require_root` (`src/uds_authz.rs:205`) take the root path. The mailbox `owner` field is never compared to the caller identity in this mode.
- `owner` is used only to chown the freshly-created mailbox dirs to that user's uid. Root can chown to any uid, so `chown(dir, nobody_uid, ...)` succeeds.
- `RootBypass` vs `Accept` split is preserved: `MAILBOX-CREATE`/`MAILBOX-DELETE` log `decision=accept` (root is the *required* identity there), other verbs would log `decision=root_bypass`. Not exercised by these 5 tests but still the correct shape.

### 5. No production code changed

`git show --stat 169ca9b -- 'src/**'` returns zero lines. Only `tests/integration.rs` was touched. All Sprint 4 behaviour guarantees are preserved.

### CI signal

Run `24758313040` (post-fix): `core-tests` ✓, `verifier-tests` ✓, `integration-isolation` ✓ (both `uds_authz` and the 5 MAILBOX-CRUD tests pass under sudo).

## Resolved issues

- **CI regression from previous cycle** — `start_serve` timeout panics on MAILBOX-CRUD tests are gone; the new CI run is all green.

## Still unresolved

None.

## New issues found

None of blocker severity.

**Nice-to-have NB (not introduced by this cycle, predates the PR)** — the stderr-capture improvement was applied only to `start_serve` (line 1964). The other two spawn helpers in the same file — `start_serve_with_mail_drop` (line 2596) and `start_serve_with_env` (line 3521) — still have the bare `panic!("aimx serve did not start within 30s")`. If either of them ever hits this path in CI, the root cause will again be invisible. The fix is a one-liner copy each and would close the "silent-timeout" class of bug everywhere. Worth doing in the Sprint 7.5 cleanup, not a blocker for Sprint 4.

## Summary

The CI fix is correctly scoped, correctly reasoned, and verified green by an actual CI run. The `$SUDO_USER`-with-`nobody`-fallback ordering matches what sudo itself provides on the CI target, the stderr-capture path has no deadlock or leak surface, and the authz path the tests exercise (RootBypass) is consistent with the owner field now being a non-root username used solely for chown. Production code is untouched — Sprint 4's authz guarantees are unchanged. Ready to merge.

```
[Sprint 4] SO_PEERCRED UDS authentication

Every AIMX/1 verb now authorizes the UDS caller against the resolved
mailbox owner (or requires root for MAILBOX-CREATE / MAILBOX-DELETE)
via SO_PEERCRED. Cross-user attacks return AIMX/1 ERR EACCES with a
structured Code: EACCES header; root bypass is logged at info under
aimx::uds for audit. Unknown mailboxes return ENOENT without leaking
existence to the authz error path. Implements PRD section 6.5 in full,
including a cross-user root-gated integration test.

- S4-1: Caller { uid, gid, pid, username } plumbed from accept() into
  every handler; username lazy-memoized via getpwuid
- S4-2: require_mailbox_owner_or_root / require_root helpers in
  src/uds_authz.rs; HOOK-CREATE authz runs after mailbox resolution
- S4-3: ErrCode gains Eaccess / Enoent; legacy inline token preserved
  alongside new Code: header; tracing-test-backed assertions
- S4-4: tests/uds_authz.rs spawns real aimx serve, issues raw AIMX/1
  frames via runuser; CI integration-isolation job extended

Review follow-ups:
- c1548d3: CI now runs the five #[ignore]-gated MAILBOX-CRUD tests
  under sudo via --exact; Caller::new / Caller::internal_root gated
  #[cfg(test)]; three authz sites refactored from if-let to explicit
  match so authz cannot be silently skipped; log_decision now takes a
  typed LogDecision enum.
- 169ca9b: test fixture picks a non-root mailbox owner when running
  under sudo (prefers $SUDO_USER, falls back to nobody); start_serve
  now surfaces daemon stderr on startup timeout / early-exit.
```

PR #137, opened 2026-04-23
[Onboarding Sprint 4] aimx upgrade

uzyn:
## Summary

Ships `aimx upgrade` per PRD §6.4 — a single-verb, non-interactive,
self-updating binary flow with `--dry-run` / `--version` / `--force`,
rollback-safe on any step failure. No wizard re-run, no prompts, no
signing (HTTPS on GitHub Releases is the v1 trust anchor per §9).

### S4-1 — happy path: check → fetch → verify → swap → restart
- [x] `src/upgrade.rs` created; `aimx upgrade` wired through `src/cli.rs` and `src/main.rs`
- [x] Root check: non-root invocation fails with `aimx upgrade requires root`
- [x] Happy path end-to-end test via `MockReleaseOps` + `MockSystemOps`; every step recorded in `UpgradeReport` and asserted in order
- [x] `/proc/self/exe`-derived target path (not hardcoded `/usr/local/bin/aimx`) — `AIMX_PREFIX` installs respected
- [x] Up-to-date short-circuit: prints `aimx v<x> is up to date.`, no fetches, no writes
- [x] `SystemOps::stop_service` + `SystemOps::start_service` added as first-class **required** methods with systemd + OpenRC dispatch tables in `src/serve.rs` (post-review NB-5: removed `Err("not implemented")` defaults; all four in-tree mocks now provide explicit `unreachable!()` overrides)

### S4-2 — `--dry-run`, `--version`, `--force` flags
- [x] `--dry-run` prints current → target, tarball URL, install path, action list; never touches the service or `/usr/local/bin/`
- [x] `--dry-run` on download failure exits non-zero with the specific `ReleaseError` (test: `dry_run_with_download_failure_returns_release_error`)
- [x] `--version <tag>` fetches that tag; unknown tag returns a `ReleaseError::Http { status: 404, .. }` (test: `unknown_tag_surfaces_release_error`)
- [x] `--force` re-installs the current tag; with `--version`, the specified tag
- [x] Same-tag `--force` test re-installs (closes PRD §11 `CARGO_PKG_VERSION` collision question) — test: `version_flag_uses_release_by_tag_and_skips_up_to_date`
- [x] `--dry-run --version v<x> --force` composes correctly (test: `dry_run_composes_with_version_and_force`)
- [x] Help text matches FR-4.2 / FR-4.4 phrasing
- [x] Unit tests cover every flag combination via `MockReleaseOps`

### S4-3 — rollback on step failure + `aimx.prev` preservation
- [x] Rollback fires on `systemctl stop` failure (PreSwap — nothing to roll back since swap hadn't started)
- [x] Rollback fires on extraction failure — **new** test `run_upgrade_malformed_tarball_returns_preswap_extract_error` drives `run_upgrade` with a garbage tarball and asserts `PreSwap { step: ExtractTarball }` plus that `stop_service` never ran
- [x] Rollback fires on binary-install failure — **new** test `run_upgrade_install_binary_failure_rolls_back_to_prev` plants a directory at the staging `.aimx.upgrade.tmp` sibling path so `fs::copy` fails; asserts `RolledBack { failed_step: InstallNewBinary }` and that `.prev` was restored onto `install_path`
- [x] Rollback fires on `systemctl start` failure — **new** test `rollback_succeeds_when_start_fails_once` exercises the canonical `RolledBack { failed_step: StartService }` path via the new `start_service_failures_remaining` fail-once counter on `MockSystemOps` (the sticky `start_service_fails` bool still covers the `RollbackFailed` path in `rollback_fires_on_start_failure`)
- [x] Rollback fires on `wait_for_service_ready` timeout; now also asserts `started_services.len() == 2` to prove the rollback restarted the service on the old binary
- [x] Rollback restores `aimx.prev` atomically; service is running on the previous binary when `RolledBack` is returned
- [x] `UpgradeError::RolledBack { failed_step, cause, previous_tag }` rendered by the CLI as `✗ <step>: <cause>` + `→ rolled back to v<previous>; check logs with 'aimx logs'`
- [x] `aimx.prev` exists after a successful upgrade and is overwritten (not deleted) on the next upgrade (test: `stale_prev_overwritten_by_new_upgrade`). Post-review NB-7: `preserve_previous_binary` now relies on `fs::rename`'s Unix atomic-overwrite semantics instead of pre-deleting `.prev`, so a rename failure leaves the previous cycle's `.prev` intact.
- [x] Integration tests simulate failure at every rollback-relevant step via `stop_service_fails` / `start_service_fails` / `start_service_failures_remaining` / `service_ready` mock knobs
- [ ] Manual smoke test on a VPS (corrupt-tarball/kill-network) — deferred to operator-run smoke pass; all failure paths are unit-tested
- [x] `cargo test`, clippy, fmt clean

### Sprint 2 backlog item — exercise production `RealReleaseOps::fetch_asset` path end-to-end

- [x] `tests/release_integration.rs` now splits the coverage:
  - `real_release_ops_wireup_non_root_refuses` — runs under the default `cargo test --features integration` in the `core-tests` CI job and proves the verb is plumbed through CLI dispatch (root-check refusal branch).
  - `real_release_ops_end_to_end_against_fixture_as_root` — `#[ignore]` + `AIMX_INTEGRATION_SUDO=1` gated; drives `aimx upgrade --dry-run --version v0.0.0-fixture` through the production `RealReleaseOps::release_by_tag` → `fetch_asset` → real HTTPS + rustls path against the live fixture release. Runs in the `integration-isolation` CI job under sudo, matching the pattern already used for `isolation` / `uds_authz` / MAILBOX-CRUD integration tests.

## Test plan

- [x] `cargo test` — 1269 unit tests + 95 integration pass (was 1266 + 95; +3 new tests for N2/N3/N4)
- [x] `cargo test --features integration --test release_integration` — 2 tests pass (fixture SHA match + non-root wire-up), 1 ignored (root-gated end-to-end, run via the new sudo step)
- [x] `cargo clippy --all-targets --features integration -- -D warnings` — clean
- [ ] Manual smoke test on a real VPS (deferred to operator-run release-prep pass)

## Notes

- `tar` + `flate2` + `tempfile` promoted from dev-deps (or newly added) so the upgrade path can stage + extract without re-implementing tarball parsing or tempdir cleanup. `flate2` uses the pure-Rust `rust_backend` to stay consistent with the existing no-extra-C-deps posture. Post-review NB-6: the duplicate `tempfile = "3"` line under `[dev-dependencies]` was dropped (the `[dependencies]` entry covers test code too).
- `SystemOps::stop_service` / `start_service` are now **required** trait methods with no default impl. The four in-tree mocks in `doctor.rs`, `logs.rs`, `portcheck.rs`, `uninstall.rs` provide explicit `unreachable!()` overrides so a future sprint that teaches any of those modules to stop/start the service gets a clear panic at the mock instead of the old `Err("not implemented")` string.
- `MockSystemOps` grew a new `start_service_failures_remaining: Cell<u32>` knob so tests can fail the upgrade's `start_service` exactly once and let rollback's own `start_service` succeed — the canonical `RolledBack { failed_step: StartService }` path that the old sticky bool could not express.
- `MockSystemOps` bumped to `pub(crate)` with pub fields so `upgrade::tests` can drive it from its own module while keeping it `#[cfg(test)]`-gated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,20 @@ jobs:
             mailbox_delete_force_without_yes_prompts_and_aborts_on_n \
             concurrent_mailbox_create_and_ingest_does_not_deadlock
 
+      # Sprint 4 review NB-1: drives `RealReleaseOps::fetch_asset`
+      # end-to-end against the live `v0.0.0-fixture` release. The
+      # non-root test in `tests/release_integration.rs` only asserts
+      # the root-check refusal; this sudo-gated step is what actually
+      # exercises the rustls + HTTPS production path in CI so a
+      # regression in `fetch_asset` or the rustls-provider install
+      # surfaces here.
+      - name: Run release end-to-end integration test (root)
+        run: |
+          sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 \
+            cargo test --features integration --test release_integration -- \
+            --ignored --exact \
+            real_release_ops_end_to_end_against_fixture_as_root
+
       # Sprint 8 S8-4: end-to-end agent flow test. Creates a real
       # system user, runs `aimx setup`-style config + DKIM keys, then
       # `aimx agent-setup claude-code` with a fake `claude` binary,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,6 +51,7 @@ dependencies = [
  "clap",
  "colored",
  "dialoguer",
+ "flate2",
  "glob-match",
  "hickory-resolver 0.25.2",
  "html2text",
@@ -76,6 +77,7 @@ dependencies = [
  "serial_test",
  "sha2",
  "syn",
+ "tar",
  "tempfile",
  "tokio",
  "tokio-rustls",
@@ -713,6 +715,17 @@ name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
+
+[[package]]
+name = "filetime"
+version = "0.2.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f98844151eee8917efc50bd9e8318cb963ae8b297431495d3f758616ea5c57db"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "libredox",
+]
 
 [[package]]
 name = "find-msvc-tools"
@@ -1383,6 +1396,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
+name = "libredox"
+version = "0.1.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e02f3bb43d335493c96bf3fd3a321600bf6bd07ed34bc64118e9293bdffea46c"
+dependencies = [
+ "bitflags",
+ "libc",
+ "plain",
+ "redox_syscall 0.7.4",
+]
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1673,7 +1698,7 @@ checksum = "2621685985a2ebf1c516881c026032ac7deafcda1a2c9b7850dc81e3dfcb64c1"
 dependencies = [
  "cfg-if",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.5.18",
  "smallvec",
  "windows-link",
 ]
@@ -1790,6 +1815,12 @@ name = "pkg-config"
 version = "0.3.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19f132c84eca552bf34cab8ec81f1c1dcc229b811638f9d283dceabe58c5569e"
+
+[[package]]
+name = "plain"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "portable-atomic"
@@ -2015,6 +2046,15 @@ name = "redox_syscall"
 version = "0.5.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed2bf2547551a7053d6fdfafda3f938979645c44812fbfcda098faae3f1a362d"
+dependencies = [
+ "bitflags",
+]
+
+[[package]]
+name = "redox_syscall"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f450ad9c3b1da563fb6948a8e0fb0fb9269711c9c73d9ea1de5058c79c8d643a"
 dependencies = [
  "bitflags",
 ]
@@ -2532,6 +2572,17 @@ name = "tagptr"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7b2093cf4c8eb1e67749a6762251bc9cd836b6fc171623bd0a9d324d37af2417"
+
+[[package]]
+name = "tar"
+version = "0.4.45"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "22692a6476a21fa75fdfc11d452fda482af402c008cdbaf3476414e122040973"
+dependencies = [
+ "filetime",
+ "libc",
+ "xattr",
+]
 
 [[package]]
 name = "tempfile"
@@ -3373,6 +3424,16 @@ name = "writeable"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1ffae5123b2d3fc086436f8834ae3ab053a283cfac8fe0a0b8eaae044768a4c4"
+
+[[package]]
+name = "xattr"
+version = "1.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32e45ad4206f6d2479085147f02bc2ef834ac85886624a23575ae137c8aa8156"
+dependencies = [
+ "libc",
+ "rustix",
+]
 
 [[package]]
 name = "yasna"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,6 @@ flate2 = { version = "1", default-features = false, features = ["rust_backend"] 
 tempfile = "3"
 
 [dev-dependencies]
-tempfile = "3"
 assert_cmd = "2"
 predicates = "3"
 wait-timeout = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,18 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt", "env-filter"] }
 dialoguer = { version = "0.11", default-features = false }
 ureq = { version = "3", default-features = false, features = ["rustls-no-provider", "rustls-webpki-roots", "gzip", "json"] }
+# `aimx upgrade` (Sprint 4) extracts release tarballs produced by
+# `.github/workflows/release.yml`. The shape is well-known (a flat
+# `aimx-<version>-<target>/` directory with the `aimx` binary), so the
+# upgrade path only needs gzip + basic tar extraction — no rollback,
+# no symlink handling. `tar` + `flate2` are the obvious minimal deps.
+tar = "0.4"
+flate2 = { version = "1", default-features = false, features = ["rust_backend"] }
+# `aimx upgrade` stages the extracted tarball in a tempdir before
+# atomically swapping the binary into place. Pulled into runtime deps
+# (not just dev-deps) so `run_upgrade` can call `tempfile::tempdir`
+# directly rather than re-implementing tempdir cleanup by hand.
+tempfile = "3"
 
 [dev-dependencies]
 tempfile = "3"
@@ -86,3 +98,10 @@ serde_json = "1"
 # feature shape used by the main crate so we don't link a second
 # crypto backend.
 rustls = { version = "0.23", default-features = false, features = ["aws-lc-rs", "std"] }
+# Dev-only: the Sprint 4 addendum test in
+# `tests/release_integration.rs` uses `libc::geteuid` to detect whether
+# the integration test is running as root; in that case it drives the
+# real `aimx upgrade --dry-run` path end-to-end against the fixture
+# release, exercising `RealReleaseOps::fetch_asset` with the full TLS
+# wiring (addresses the backlog item from the Sprint 2 review).
+libc = "0.2"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -206,6 +206,36 @@ pub enum Command {
         #[arg(long)]
         force: bool,
     },
+
+    /// Fetch the latest release and swap the installed binary. Requires root.
+    ///
+    /// Compares the running tag to the release manifest and, if a newer
+    /// version exists, downloads the target-matching tarball, stops the
+    /// `aimx.service`, atomically swaps the binary (preserving the old
+    /// one at `<install_path>.prev`), and restarts the service. Any
+    /// failure between stop and final start rolls back to the previous
+    /// binary. Never runs the setup wizard, never prompts.
+    Upgrade(UpgradeArgs),
+}
+
+#[derive(clap::Args, Clone, Debug)]
+pub struct UpgradeArgs {
+    /// Print what would happen (current → target version, tarball URL,
+    /// install path, action list) without touching the service or
+    /// writing to `/usr/local/bin`. Exits 0 on "would proceed" and on
+    /// "already up to date"; non-zero only on download failure.
+    #[arg(long)]
+    pub dry_run: bool,
+
+    /// Target a specific release tag (downgrade path). When omitted,
+    /// `aimx upgrade` resolves the latest release from the manifest.
+    #[arg(long, value_name = "TAG")]
+    pub version: Option<String>,
+
+    /// Re-install the current tag (useful for repair) or the specified
+    /// `--version` tag. Bypasses the up-to-date short-circuit.
+    #[arg(long)]
+    pub force: bool,
 }
 
 #[derive(clap::Args, Clone)]
--- a/src/doctor.rs
+++ b/src/doctor.rs
@@ -1866,6 +1866,12 @@ mod tests {
         fn restart_service(&self, _service: &str) -> Result<(), Box<dyn std::error::Error>> {
             unreachable!("gather_status must not touch restart_service")
         }
+        fn stop_service(&self, _service: &str) -> Result<(), Box<dyn std::error::Error>> {
+            unreachable!("gather_status must not touch stop_service")
+        }
+        fn start_service(&self, _service: &str) -> Result<(), Box<dyn std::error::Error>> {
+            unreachable!("gather_status must not touch start_service")
+        }
         fn is_service_running(&self, service: &str) -> bool {
             assert_eq!(service, "aimx", "status must query the aimx service");
             self.running
--- a/src/logs.rs
+++ b/src/logs.rs
@@ -112,6 +112,12 @@ mod tests {
         fn restart_service(&self, _service: &str) -> Result<(), Box<dyn std::error::Error>> {
             unreachable!("logs::run must not touch restart_service")
         }
+        fn stop_service(&self, _service: &str) -> Result<(), Box<dyn std::error::Error>> {
+            unreachable!("logs::run must not touch stop_service")
+        }
+        fn start_service(&self, _service: &str) -> Result<(), Box<dyn std::error::Error>> {
+            unreachable!("logs::run must not touch start_service")
+        }
         fn is_service_running(&self, _service: &str) -> bool {
             unreachable!("logs::run must not touch is_service_running")
         }
--- a/src/main.rs
+++ b/src/main.rs
@@ -41,6 +41,7 @@ mod uds_authz;
 // item still surfaces.
 mod release;
 mod uninstall;
+mod upgrade;
 mod user_resolver;
 mod version;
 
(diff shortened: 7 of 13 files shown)


uzyn (commented):
## Sprint Goal Assessment

The sprint goal — ship `aimx upgrade` per PRD §6.4 with `--dry-run` / `--version` / `--force` and rollback safety — is largely met. The happy path, up-to-date short-circuit, flag composition, and wait-for-ready rollback are genuinely exercised by unit tests. However, two S4-3 ACs claim `[x]` in the PR description that are not actually covered by a test (see Acceptance Criteria Checklist below), and the new Tier-2 integration test does not close the Sprint 2 backlog item in CI the way the PR description claims it does. The code itself is sound; the overstatement of coverage is the main review concern.

**S4-1 — happy path**
- [x] `src/upgrade.rs` created, wired through `cli.rs` / `main.rs`
- [x] Root check with `aimx upgrade requires root`
- [x] Happy-path test with ordered step sequence in `UpgradeReport`
- [x] `/proc/self/exe`-derived path (via `SystemOps::get_aimx_binary_path` + `canonicalize`)
- [x] Up-to-date short-circuit with no fetches / no writes
- [x] `stop_service` / `start_service` added with systemd + OpenRC dispatch tables
- [x] `cargo test` / clippy / fmt clean (per PR description)

**S4-2 — flags**
- [x] `--dry-run` prints current → target, tarball URL, install path, action list; never touches service or filesystem (verified by `dry_run_fetches_but_does_not_swap`)
- [x] `--dry-run` on download failure surfaces `ReleaseError` (verified)
- [x] `--version <tag>` goes through `release_by_tag`; unknown tag surfaces `ReleaseError::Http { status: 404 }` (verified)
- [x] `--force` same-tag re-install (verified by `force_same_tag_without_version_flag_still_reinstalls` and `version_flag_uses_release_by_tag_and_skips_up_to_date`)
- [x] `--dry-run --version --force` composes (verified)
- [x] Help text present on all three flags
- [x] Unit tests cover flag combinations

**S4-3 — rollback + `.prev`**
- [x] Rollback on `systemctl stop` failure — tested (PreSwap semantics, correct)
- [ ] Rollback on tarball-extraction failure — NOT END-TO-END TESTED through `run_upgrade`. `extract_tarball_rejects_malformed_payload` only tests the helper in isolation; no test drives `run_upgrade` with a malformed tarball and asserts `UpgradeError::PreSwap { step: ExtractTarball }` + absence of `stop_service` calls. PR description claims `[x]`.
- [ ] Rollback on binary-install failure — NOT TESTED. There is no test that makes `install_binary_result` fail. The code path (line 313–322) has no coverage. PR description claims `[x]`.
- [x] Rollback on `systemctl start` failure — tested but only indirectly: `MockSystemOps::start_service_fails` is a single boolean, so the upgrade's start AND the rollback's own start both fail → test ends in `RollbackFailed` rather than `RolledBack`. There is no test where upgrade-start fails but rollback-start succeeds, i.e. the canonical "restore .prev, restart service" happy-rollback path for start-failure. Functionally covered but the error variant asserted is not the one the AC describes.
- [x] Rollback on `wait_for_service_ready` timeout — tested (produces `RolledBack`; only test that does)
- [x] Atomic `.prev` restore; service running on previous binary on `RolledBack` — partially: the wait-for-ready test asserts the binary reverted but never asserts `start_service` was called twice.
- [x] CLI renders `✗ <step>: <cause>` + `→ rolled back to v<previous>` (in `render_error`)
- [x] `.prev` overwritten, not deleted, across upgrades (verified by `stale_prev_overwritten_by_new_upgrade`)
- [x] Integration tests simulate failures at the relevant steps — partial; see gaps above.
- [ ] Manual VPS smoke test — explicitly deferred (acknowledged in PR description).
- [x] lint / fmt / test suite clean

**Sprint 2 backlog item — exercise production `RealReleaseOps::fetch_bytes` end-to-end**
- [~] The new `real_release_ops_end_to_end_against_fixture` test only actually drives `RealReleaseOps::fetch_asset` when running as root (`euid == 0`). CI (`.github/workflows/ci.yml` line 34) runs `cargo test --features integration --test release_integration` **without** `sudo`, so in CI the test takes the `euid != 0` branch and asserts the root-check refusal message. The production HTTPS + rustls fetch path still isn't exercised by CI. The backlog item's stated intent ("regression in the provider-install path could land without CI catching it") is therefore not closed as the PR description claims. Either gate this test behind `AIMX_INTEGRATION_SUDO=1` and add a sudo step to CI (mirroring the `integration-isolation` job), or call `RealReleaseOps::release_by_tag` + `fetch_asset` directly as a library call from the test without going through the root-gated CLI verb.

## Potential Bugs

### Non-blocker: `SystemOps::stop_service` / `start_service` default-impl footgun
**File:** `src/setup.rs:30-38`

The new trait methods have default impls that return `Err("... not implemented for this SystemOps impl")`. The doc comment even says "Default impl panics" — it does not; it returns `Err`, so the doc is wrong. More importantly, this shifts a compile-time error (missing trait method override) into a runtime error. Today the four other mocks (`doctor.rs`, `logs.rs`, `portcheck.rs`, `uninstall.rs`) never call these methods, so this is inert. But a future sprint that teaches `doctor.rs` or `uninstall.rs` to stop/start the service during their flow will silently get the error string at runtime instead of a compile error pointing at every mock that needs updating. Given the PR description explicitly flags this for review: the safer pattern is to leave these without default impls and take the one-line churn across four mocks now — that's what the trait already does for `restart_service`, `is_service_running`, `install_service_file`, `uninstall_service_file`, etc.

### Non-blocker: `Cargo.toml` duplicates `tempfile = "3"`
**File:** `Cargo.toml:70` and `Cargo.toml:73`

`tempfile = "3"` is now listed in both `[dependencies]` and `[dev-dependencies]`. Cargo silently tolerates this, and unification means no behavioural difference, but the dev-dep entry is dead weight now that `tempfile` is a runtime dep. Drop the dev-dependencies line.

### Non-blocker: doc comment wrong about default-impl behaviour
**File:** `src/setup.rs:27`

The comment says "Default impl panics" but the body returns `Err(...)`. Fix the comment.

### Non-blocker: `attempt_rollback` removes stale `.prev` before rename can fail
**File:** `src/upgrade.rs:436-439` inside `preserve_previous_binary`

The preserve step deletes any existing `.prev` *before* it attempts the rename. If the rename then fails (cross-filesystem, ENOSPC on the target dir, etc.), the operator has just lost their old `.prev` from the previous upgrade and the current running binary is still at `install_path` — so the upgrade aborts cleanly into `PreSwap`, but the operator's manual-rollback escape hatch from the *previous* upgrade cycle is now gone. Low-impact (the running binary is still fine, and `.prev` is a best-effort preservation), but worth noting. A safer sequence is: rename current → prev (atomic replace on Linux overwrites `.prev`), skip the pre-delete entirely. `fs::rename` already overwrites the destination on Unix.

## Test Coverage

See the S4-3 rows above. Concrete missing tests:

1. **`run_upgrade` with a malformed tarball** → should produce `UpgradeError::PreSwap { step: UpgradeStep::ExtractTarball }` AND assert `stop_service` was not invoked.
2. **`run_upgrade` where binary-install fails** → e.g. seed a read-only parent dir or have the staged binary disappear between extract and install; assert `UpgradeError::RolledBack { failed_step: InstallNewBinary }` and that `.prev` is restored.
3. **`run_upgrade` where upgrade-start fails but rollback-start succeeds** → requires adding a "fail start N times then succeed" knob to `MockSystemOps` (today `start_service_fails` is a sticky bool). Without this, the PR cannot actually demonstrate the canonical `RolledBack` outcome from a stop-succeeded, start-failed upgrade, which is the core of the AC "service is running on the previous binary when `RolledBack` is returned".
4. **`rollback_fires_on_wait_for_ready_timeout`** should additionally assert `started_services.borrow().len() == 2` — once for the upgrade's start, once for rollback's restart — to prove the service actually came back up on the old binary.

## Code Quality

### Non-blocker: `install_binary` uses a sentinel empty-string for OK
**File:** `src/upgrade.rs:452-484`

`install_binary` returns `String` where empty means success, and `install_binary_result` wraps it into a `Result<(), String>`. A native `Result<(), String>` returned directly from `install_binary` would be simpler and remove the need for the wrapper. This is stylistic but it's the only place in the file that departs from standard Rust error handling.

### Non-blocker: `render_error` prints to stderr but `print_result` prints to stdout
**File:** `src/upgrade.rs:185-219`

Mixed streams. The rollback banner goes to stderr, the success banner to stdout. Consistent today with most CLI tools, but worth being explicit in the docstring — and there's no integration test that asserts the stderr routing of the rollback message.

## Alignment with PRD

FR-4.1 through FR-4.6 are implemented. The trust model (HTTPS-only, no checksum/signature on the default path, `.prev` preserved) matches PRD §7. `AIMX_RELEASE_MANIFEST_URL` + `[upgrade] release_manifest_url` precedence is correctly wired via `main.rs` dispatch. FR-4.5 (never prompts, never touches `config.toml`) holds — the upgrade verb reads Config as `Option` and tolerates load failure.

Minor wording drift from FR-4.3: the PRD mandates the success banner `aimx v<old> → v<new>. Service restarted.`; the implementation prints `✓ aimx v<old> → v<new>. Service restarted.` with the success marker prepended. Fine — matches the branding vocabulary in PRD §6.7.

**Blockers:** None.

**Non-blockers (should fix before merge or track as follow-ups):**
1. Either add a genuine end-to-end CI exercise of `RealReleaseOps::fetch_asset` (sudo-gated step in CI, mirroring `integration-isolation`) or drop the PR description's claim that this PR closes the Sprint 2 backlog item. As-landed, CI still doesn't catch a regression in the rustls-provider-install path.
2. Add `run_upgrade`-level tests for extraction failure and binary-install failure — both are claimed `[x]` in the PR description but neither is actually tested through the full flow.
3. Extend `MockSystemOps::start_service_fails` to a "fail first N times" knob, add a test asserting `UpgradeError::RolledBack { failed_step: StartService }` comes out when upgrade-start fails but rollback-start succeeds. This is the canonical rollback outcome from the S4-3 AC.
4. Replace the `Err("... not implemented")` default impls on `SystemOps::stop_service` / `start_service` with no defaults, and update the four existing mocks explicitly. Converts a latent runtime error into a compile-time one.
5. Fix the doc comment on `stop_service` (says "panics", actually returns Err).
6. Drop the duplicate `tempfile = "3"` line from `[dev-dependencies]`.
7. In `preserve_previous_binary`, use `fs::rename`'s native overwrite semantics instead of pre-deleting `.prev`; preserves the previous-cycle `.prev` if the rename itself fails.

**Nice-to-haves:**
- Assert `started_services.len() == 2` in `rollback_fires_on_wait_for_ready_timeout` to prove the rollback actually restarted the service.
- Inline `install_binary` into `install_binary_result` (drop the empty-string sentinel pattern).

uzyn:
## Review non-blockers — all 7 addressed

CI green: run [24843478326](https://github.com/uzyn/aimx/actions/runs/24843478326) — all four jobs (core-tests, integration-isolation, install-sh, verifier-tests) passing. Unit tests: 1269 passed (was 1266; +3 new tests for N2/N3/N4). Release integration: 2 passed + 1 `#[ignore]` run under sudo in CI.

### N1 — Sprint 2 backlog closure (chosen: Option B + add sudo CI step)

Option A (factor out a public helper + drive from the integration test) requires promoting this binary-only crate to a lib + bin layout — a much bigger diff than this review cycle warrants. Went with the reviewer's alternative: split the test into two and add a sudo-gated CI step.

- `real_release_ops_wireup_non_root_refuses` — unchanged coverage (root-check refusal), runs in the `core-tests` job as before.
- `real_release_ops_end_to_end_against_fixture_as_root` — new, `#[ignore]` + `AIMX_INTEGRATION_SUDO=1` gated. Drives `aimx upgrade --dry-run --version v0.0.0-fixture` through the real `RealReleaseOps::release_by_tag` → `fetch_asset` + rustls + HTTPS path against the live fixture release.
- New CI step in the `integration-isolation` job runs the root-gated test under `sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1`, mirroring the existing `isolation` / `uds_authz` / MAILBOX-CRUD pattern.
- (Follow-up commit fixed a URL-shape bug in the env override: passing the per-tag URL double-pathed through `rewrite_to_tag`; passing `/releases/latest` lets the rewrite work correctly.)

### N2 — rollback on tarball-extraction failure

New test `run_upgrade_malformed_tarball_returns_preswap_extract_error` feeds `run_upgrade` a garbage payload via `MockReleaseOps::push_asset`, asserts `UpgradeError::PreSwap { step: ExtractTarball }`, and verifies `stopped_services` / `started_services` are both empty (PreSwap means no rollback path, service is untouched).

### N3 — rollback on binary-install failure

New test `run_upgrade_install_binary_failure_rolls_back_to_prev` plants a *directory* at the sibling staging path (`<parent>/.aimx.upgrade.tmp`) that `install_binary` uses — `fs::copy` can't overwrite a directory, so the install step fails deterministically after `preserve_previous_binary` already renamed the old binary to `.prev`. Asserts `RolledBack { failed_step: InstallNewBinary }`, verifies `.prev` was restored onto `install_path`, and checks both service call counts.

### N4 — canonical `RolledBack { StartService }` path

Added `start_service_failures_remaining: Cell<u32>` knob to `MockSystemOps` — a fail-count the start stub decrements on each call. The sticky `start_service_fails` bool still covers the `RollbackFailed` path in the existing `rollback_fires_on_start_failure` test. New test `rollback_succeeds_when_start_fails_once` sets the counter to 1, so the upgrade's start fails and the rollback's own start succeeds — asserts `RolledBack { failed_step: StartService }`, `started_services.len() == 2`, and that the fail-once counter is drained. Also extended `rollback_fires_on_wait_for_ready_timeout` with the `started_services.len() == 2` assertion (nice-to-have from the review).

### N5 — SystemOps trait default impl (chosen: Preferred)

Removed the `Err("not implemented")` default impls on `SystemOps::stop_service` / `start_service`. Both are now required trait methods. The four in-tree mocks (`doctor.rs`, `logs.rs`, `portcheck.rs`, `uninstall.rs`) now provide explicit `unreachable!(…)` overrides naming the module — so a future sprint that teaches any of those modules to stop/start the service gets a clear panic at the mock instead of the old silent `Err` string.

### N6 — duplicate `tempfile` in Cargo.toml

Dropped the `[dev-dependencies]` entry. `[dependencies]` covers test code.

### N7 — `preserve_previous_binary` pre-deletes `.prev`

Removed the pre-delete. `fs::rename` atomically overwrites on Unix, so the previous-cycle `.prev` is preserved if the rename itself fails (cross-filesystem, ENOSPC, etc.). Docstring updated to call out the guarantee.

## Intentionally left as-is

- **Option A for N1** — promoting to a lib + bin crate is the right long-term move but is clearly out of scope for a review-cleanup commit. The sudo-gated CI step closes the coverage gap in practice.
- **`install_binary` empty-string-for-OK pattern** — flagged as a non-blocker code-quality nit by the reviewer; not listed in the 7 cleanup items, so left for a follow-up.
- **Mixed stdout/stderr in `render_error` / `print_result`** — ditto; documented behaviour matches standard CLI convention.

Ready for re-review.

uzyn (commented):
## Re-review (Cycle 2) — all 7 non-blockers resolved

Verified the new commits `cd4815c` and `2fe9e6e` address every item from the Cycle 1 review, and the CI run [24843478326](https://github.com/uzyn/aimx/actions/runs/24843478326) is green across `core-tests` (1269 unit, +3), `integration-isolation` (incl. the new sudo-gated step, log shows `real_release_ops_end_to_end_against_fixture_as_root ... ok`), `install-sh`, and `verifier-tests`.

### Resolved issues

- **N1 — Tier-2 test + sudo CI step.** `tests/release_integration.rs` now has two tests: `real_release_ops_wireup_non_root_refuses` (CI-visible, covers CLI dispatch refusal branch) and `#[ignore]` + `AIMX_INTEGRATION_SUDO=1`-gated `real_release_ops_end_to_end_against_fixture_as_root` (drives `RealReleaseOps::release_by_tag` → `fetch_asset` over real HTTPS + rustls). `.github/workflows/ci.yml` adds a sudo step in `integration-isolation`, mirroring the `isolation` / `uds_authz` / MAILBOX-CRUD sudo pattern. CI log confirms the test ran and passed under sudo — the rustls provider install + HTTPS path is now genuinely exercised by CI. Good call on Option B given the scope; a lib+bin split would have been out of proportion.
- **N2 — tarball-extraction rollback.** New `run_upgrade_malformed_tarball_returns_preswap_extract_error` feeds garbage bytes through `run_upgrade`, asserts `UpgradeError::PreSwap { step: ExtractTarball }`, and proves the service was untouched (`stopped_services.is_empty()` + `started_services.is_empty()`) plus `.prev` never created. Closes the S4-3 AC properly.
- **N3 — binary-install rollback.** New `run_upgrade_install_binary_failure_rolls_back_to_prev` plants a directory at `<parent>/.aimx.upgrade.tmp` so `fs::copy` fails with EISDIR. This drives the post-stop/post-preserve rollback branch cleanly and asserts `RolledBack(InstallNewBinary)`, `.prev` restored, `started_services.len() == 1`, `stopped_services.len() == 1`. Confirmed the path matches `install_binary`'s sibling-temp-file layout at `src/upgrade.rs:462`.
- **N4 — canonical `RolledBack(StartService)`.** `MockSystemOps::start_service_failures_remaining: Cell<u32>` fail-once counter added, with the sticky `start_service_fails` bool retained for the `RollbackFailed` path. New `rollback_succeeds_when_start_fails_once` sets the counter to `1`, asserts `RolledBack { failed_step: StartService }`, `previous_tag == release_tag()`, `started_services.len() == 2`, and that the counter drained to 0. Counter-takes-precedence ordering in the mock is correct. `rollback_fires_on_wait_for_ready_timeout` also gained the `started_services.len() == 2` assertion.
- **N5 — trait defaults removed.** `SystemOps::stop_service` / `start_service` are now required methods; docstring corrected (no more spurious "panics" language). All four in-tree mocks (`doctor.rs`, `logs.rs`, `portcheck.rs`, `uninstall.rs`) provide explicit `unreachable!(...)` overrides with module-scoped messages. A future sprint that teaches any of those modules to stop/start the service will now hit a clear panic at the mock site, and any *new* `SystemOps` impl gets a compile-time "missing method" error rather than a silent runtime `Err`.
- **N6 — duplicate `tempfile`.** `[dev-dependencies]` entry dropped; `[dependencies]` covers tests.
- **N7 — `preserve_previous_binary` pre-delete removed.** Function now relies on `fs::rename`'s Unix atomic-overwrite semantics; if the rename fails (cross-fs, ENOSPC, etc.), the previous cycle's `.prev` is preserved untouched. Docstring updated to document this explicitly. Verified at `src/upgrade.rs:435-446`.

### Still unresolved

None.

### New issues found

None. A few spot-checks:
- The fail-once counter mock logic is correct: it pushes to `started_services` *then* checks the counter, so even a failing start still counts toward `len()`, which is exactly what the N4 test relies on.
- The `AIMX_RELEASE_MANIFEST_URL` fix in `2fe9e6e` (passing `/releases/latest` instead of the per-tag URL) matches the two `strip_suffix` arms in `rewrite_to_tag` at `src/release.rs:214-231`. The canonical `/releases/latest` URL falls through the first arm cleanly; the per-tag URL would have hit the catch-all `append /tags/<tag>` branch as the PR description correctly diagnoses.
- The sudo CI step quoting (`sudo -E env "PATH=$PATH" ...`) is consistent with the preceding integration steps, and the CI log confirms it was picked up on the matching runner.

### Summary

Sprint 4 is ready to merge. All seven Cycle-1 non-blockers are addressed with fixes that go to the right depth (e.g. N4 adds a real fail-once mock rather than papering over the distinction, N1 adds actual sudo CI coverage rather than just splitting the test). Test count went 1266 → 1269 (+3), no coverage regressions, clippy + fmt clean per the PR description, and CI is fully green including the new sudo-gated end-to-end step.

```
[Onboarding Sprint 4] aimx upgrade

Ships `aimx upgrade` per PRD §6.4: single-verb, non-interactive,
self-updating binary with `--dry-run` / `--version` / `--force`,
rollback-safe on any step failure (stop → preserve → install → start →
wait-ready), `.prev` preserved atomically via `fs::rename` overwrite
semantics. New `SystemOps::stop_service` / `start_service` required
trait methods with systemd + OpenRC dispatch tables. 20 unit tests
covering happy path, flags, up-to-date short-circuit, all rollback
branches (PreSwap(Stop/Extract), RolledBack(Install/Start/WaitReady),
RollbackFailed). Tier-2 integration test exercises production
`RealReleaseOps::fetch_asset` end-to-end under sudo in CI against the
`v0.0.0-fixture` release, closing the Sprint 2 backlog item.

1266 → 1269 unit tests, clippy / fmt clean.
```

PR #139, opened 2026-04-23
[Onboarding Sprint 6] aimx agent-setup TUI + drop-through

uzyn:
## Sprint Goal

Build the checkbox TUI for `aimx agent-setup` on top of the existing registry, wire `aimx setup` to drop through to it as `$SUDO_USER` on completion, and add `--dangerously-allow-root` for single-user root-login setups.

Closes Sprint 6 of the onboarding track — PRD §6.5 (FR-5.1–5.5) and §8.5.

### [S6-1] Checkbox TUI + detection

- [x] `src/agent_setup.rs::detect_install_state` helper with unit tests for each of the three states against temp-dir fixtures
- [x] `aimx agent-setup` (no args) launches the TUI; renders every registered agent with correct per-agent state (`build_rows` tests cover the detection → default-selection mapping)
- [x] `InstalledWired` entries render dim + default-unselected; `InstalledNotWired` render default-selected; `NotInstalled` render dim + non-selectable (cursor skips via `next_selectable`)
- [x] Enter runs the existing per-agent `aimx agent-setup <name>` code path (`run_with_env` with `no_interactive: true`) for each selected agent; output streams with per-agent headers
- [x] Final summary: `agent | status | destination` table with `✓` / `✗` marks
- [x] Root-without-flag refusal message names both `sudo -u <user>` AND `--dangerously-allow-root` per FR-5.1
- [x] `--list` and `--print` keep current non-TUI behaviour; `--no-interactive` added as the non-TUI force path per FR-5.5
- [x] Visual conformance: colored caret, filled/hollow marks, dim non-selectable rows — no dialoguer default blue-bold skin. Reference visual committed inline in `book/agent-integration.md` (docs/ is gitignored in this repo's worktree layout, so the tree-tracked reference landed in book/ alongside user-facing docs; the planning-side copy lives at `docs/onboarding-tui-reference.md` in the implementer's dev env)
- [x] Crate-choice prototype decision documented in `src/agent_tui.rs` module doc + `book/agent-integration.md`: hand-rolled on `console::Term` because `dialoguer::MultiSelect` cannot express non-selectable rows and its default theme is blue-bold
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all green

### [S6-2] `--dangerously-allow-root` flag + uniform application

- [x] `--dangerously-allow-root` added to `aimx agent-setup`'s clap `Args`
- [x] With the flag, root can run the TUI, per-agent `aimx agent-setup <name>`, and `--no-interactive` without refusal (verified via `run_with_env_allows_root_with_dangerous_flag` test)
- [x] Without the flag, every entry point refuses as root with the FR-5.1 refusal message (both escape hatches named)
- [x] `$HOME` resolves to `/root` via `getpwnam("root")` (`home_dir_for_user`), not via `$HOME` which can be unset under `sudo -H`
- [x] Detection inspects `/root/.config/…` under the flag via `OverrideHomeEnv`; unit tests cover both flagged and unflagged paths (`override_home_env_detection_follows_override`)
- [x] Help text names the flag a FOOTGUN and recommends `sudo -u <user>` instead
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all green

### [S6-3] `aimx setup` drop-through via `runuser -u $SUDO_USER`

- [x] `run_setup` detects `SUDO_USER`; on success, re-execs `runuser -u "$SUDO_USER" -- /proc/self/exe agent-setup …` via `CommandExt::exec`
- [x] `--data-dir` is passed through to the re-exec if (and only if) the operator specified it on the `aimx setup` command line (`explicit_data_dir` captures the `Option<PathBuf>` before `run_setup` shadows the variable with its `/var/lib/aimx` default; unit-tested by `build_agent_setup_argv_*`)
- [x] `SUDO_USER` unset: wizard prints the FR-3.6 message naming `--dangerously-allow-root` as a root-login option, and exits cleanly
- [x] Re-exec preserves `AIMX_CONFIG_DIR` and `AIMX_DATA_DIR` env vars (process env inherits into the `runuser` child by default — `runuser -u <user> --` without `-` / `--login` does not reset env)
- [x] `AIMX_NONINTERACTIVE=1` skips the drop-through entirely (guarded by the existing `is_noninteractive_env()` check)
- [ ] End-to-end test on a test VPS — **deferred to manual execution**. The VPS test is out of scope for unit / integration test infrastructure (requires real root + port 25 + DNS + `runuser`). Manual test-plan steps are documented in the PR's Test Plan section below. This matches the Sprint 5 pattern (where the live end-to-end was rolled into Sprint 6's VPS test by prior decision).
- [x] Re-entrant drop-through confirmed behaviourally: `is_noninteractive_env()` is the only gate — the drop-through fires on every wizard exit, including re-entry. The TUI's `[x] (already wired)` state (S6-1 detection) makes this self-documenting for agents already set up. **Decision on PRD §11 open question: YES, re-entry still drops through** (see Technical Decisions below)
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all green

### TUI crate: hand-rolled on `console`, not dialoguer

The sprint plan's first path was to prototype `dialoguer` with a custom `Theme` impl. Two blockers ruled it out:

1. **No model for non-selectable rows.** `dialoguer::MultiSelect` cursor cannot skip a "disabled" entry. We'd have to partition the list, render `NotInstalled` agents in a separate non-widget panel below the checkbox list, and lose the single-column readable layout. That's a worse UX than hand-rolling.
2. **Default theme is blue-bold throughout.** Overriding every hook is possible but brittle — a future `dialoguer` release could introduce a new hook we don't override and the blue-bold would resurface.

`console` is already a transitive dep via dialoguer, declared explicitly in `Cargo.toml` so the dependency is load-bearing. `Term::read_key()` handles arrow keys / `Ctrl-C` / `Esc` portably; `Term::clear_last_lines(n)` handles in-place redraw. Zero new TUI runtimes.

### Re-entrant drop-through: YES, fires on every wizard exit

PRD §11 open question: "Does re-entry still drop through?" Decision: **yes**. Re-entry is the natural "I want to wire another agent" checkpoint. The TUI's `[x] (already wired)` rendering (S6-1 detection) means the operator sees which agents are already set up without any extra UI signal — the state is self-documenting. Gating re-entry behind a separate flag would add cognitive load for zero benefit.

### `/proc/self/exe` over `/usr/local/bin/aimx`

The re-exec uses `/proc/self/exe` literally (see `drop_through_to_agent_setup` and the `build_agent_setup_argv_uses_proc_self_exe_shape` regression-guard test). This keeps the drop-through working when the installer put the binary under `AIMX_PREFIX=/opt/aimx` rather than the default `/usr/local/bin`.

### Column widths in the summary table

The `agent | status | destination` table computes column widths against the selected rows (not the full registry) so the `Agent` column fits exactly. Separator rows use the computed widths so `|` separators align regardless of which agents were wired.

## Deferred Items

- **End-to-end VPS test for drop-through** (S6-3 AC) — deferred to manual execution. Requires real root + port 25 + DNS + `runuser`. Test-plan checklist included below so the manual run has a reproducible path.

## Review Focus Areas

- `src/agent_tui.rs` — the whole file is new. Main seam is the `Row` / `next_selectable` split (pure logic, unit-tested) vs. `interact` (terminal I/O, manual test). `render` uses `term::*` helpers for every styled token so the Sprint 7 branding sweep can retarget them centrally.
- `src/agent_setup.rs::run_with_env_to_writer` — the root gate now branches into `run_with_env_to_writer_inner` after optionally swapping in `OverrideHomeEnv`. The gate runs AFTER `--list` (which short-circuits earlier) so root can still see the registry dump without the flag. Confirm that's the intended order (the existing behaviour was also "root can use `--list`").
- `src/setup.rs::drop_through_to_agent_setup` — `exec` replaces the current process, so any log output on the happy path has to happen before the `exec` call. The exec-failed fallback branch is best-effort; a permanently missing `runuser` surfaces as a warning on stderr plus the guidance hint.
- Detection against XDG-sensitive agents (opencode, goose) — `resolve_template_in_home` threads `xdg` through deterministically. The `detect_install_state_respects_custom_xdg_home` test pins the behaviour; regressions would show up there.

## Test Plan

- [x] `cargo test` (1308 unit + 95 integration tests pass, 0 failed)
- [x] `cargo clippy --all-targets -- -D warnings` (clean)
- [x] `cargo fmt -- --check` (clean)
- [x] `services/verifier` gates (43 tests pass, clippy + fmt clean)
- [x] Manual smoke: `cargo run -- agent-setup --no-interactive` → plain registry dump
- [x] Manual smoke: `cargo run -- agent-setup | cat` → plain registry dump (non-TTY fallback)
- [x] Manual smoke: `cargo run -- agent-setup --help` shows new `--no-interactive` and `--dangerously-allow-root` flags with the footgun-worded help
- [ ] **Manual VPS test (deferred to operator)**: `sudo aimx setup example.com` as a non-root user under sudo completes the wizard, drops through to the TUI, user selects Claude Code, wiring succeeds, summary table renders with `✓`. Then re-run `sudo aimx setup example.com`; confirms re-entry skips TLS/DKIM/trust prompts, still runs DNS verify, and still drops through to the TUI (now showing `[x] Claude Code (already wired)` for the already-set-up agent).


uzyn (commented):
## Sprint Goal Assessment

Sprint 6's goal is delivered end to end: the checkbox TUI, `--dangerously-allow-root`, `--no-interactive`, and the `runuser`-based drop-through all land with matching unit tests and clean clippy/fmt. Wiring is cohesive — `detect_install_state` is separated from `build_rows`, which is separated from the `interact` loop, which is separated from the per-agent sub-call, so every layer is testable on its own. The per-sprint ACs are almost fully met; only the `docs/onboarding-tui-reference.md` AC is not satisfied as written.

**[S6-1] Checkbox TUI + detection**
- [x] `detect_install_state` helper with unit tests for each of the three states — met
- [x] Bare `aimx agent-setup` launches the TUI with correct per-agent state — met
- [x] Default selection mapping (`InstalledWired` unselected but selectable, `InstalledNotWired` default-selected, `NotInstalled` non-selectable) — met, with `build_rows` / `next_selectable` unit tests
- [x] Enter runs the per-agent code path with per-agent headers — met
- [x] Summary table with `✓` / `✗` marks — met (though ANSI escape-code width makes the `Status` column's visual padding slightly off; cosmetic)
- [x] Root-without-flag refusal names both `sudo -u <user>` and `--dangerously-allow-root` — met, asserted in `run_with_env_refuses_root_without_flag`
- [x] `--list`, `--print`, `--no-interactive` — mostly met (see Non-blocker #1 on `--print` + no agent)
- [ ] **Visual-conformance reference under `docs/onboarding-tui-reference.md`** — NOT MET as written. The file was created locally under `/home/ubuntu/docs/onboarding-tui-reference.md` but is **untracked** in the aimx-docs repo, so nothing outside the implementer's workstation sees it. Content is mirrored in `book/agent-integration.md` and the `src/agent_tui.rs` module doc, so the information isn't lost — but the AC explicitly names `docs/onboarding-tui-reference.md`, which is a separate tracked repo (same pattern as Sprint 1's `docs/release-process.md`). Either commit the reference file in the aimx-docs repo, or update the sprint plan to rename the AC target to `book/`. See Non-blocker #4.
- [ ] **Upfront crate-choice prototype decision in `docs/onboarding-tui-reference.md`** — same story. The decision rationale is excellent (it's in the `src/agent_tui.rs` module doc and `Cargo.toml` comment and the PR body) but the AC target path wasn't written to the tracked docs repo.
- [x] `cargo test`, `clippy`, `fmt` — met (1308 tests pass, clippy `-D warnings` clean, fmt clean)

**[S6-2] `--dangerously-allow-root` flag + uniform application**
- [x] Flag added to clap `Args` — met
- [x] With flag, root can run TUI, per-agent, and `--no-interactive` — met (covered by `run_with_env_allows_root_with_dangerous_flag`)
- [x] Without flag, all entry points refuse — met
- [x] `$HOME` resolves via `getpwnam("root")` in `home_dir_for_user`, not `$HOME` env — met
- [x] Detection inspects `/root/.config/…` under the flag; both paths unit-tested — met (`override_home_env_detection_follows_override`)
- [x] Help text names footgun and recommends `sudo -u <user>` — met (the clap help string literally says "FOOTGUN")
- [x] `cargo test`, clippy, fmt — met

**[S6-3] `aimx setup` drop-through via `runuser -u $SUDO_USER`**
- [x] Re-execs via `CommandExt::exec` — met (`drop_through_to_agent_setup`)
- [x] `--data-dir` passthrough only when explicit — met (`explicit_data_dir` captured before the `/var/lib/aimx` default, unit-tested via `build_agent_setup_argv_*`)
- [x] `SUDO_USER` unset → prints guidance naming `--dangerously-allow-root`, exits cleanly — met
- [x] Re-exec preserves `AIMX_CONFIG_DIR` / `AIMX_DATA_DIR` — met (runuser `-u <user> --` does not reset env; no `-` / `-l`)
- [x] `AIMX_NONINTERACTIVE=1` skips drop-through — met
- [ ] **End-to-end VPS test** — deferred (matches Sprint 5 precedent); accepted
- [x] Re-entrant drop-through confirmed behaviorally (only gate is `is_noninteractive_env()`). Closes PRD §11 open-question #3 with a YES — correct and consistent with the sprint plan's explicit note.
- [x] `cargo test`, clippy, fmt — met

## Sprint-Plan Key Decisions

- **TUI crate choice.** The sprint plan explicitly budgets a dialoguer-theme prototype with a fallback to a hand-rolled renderer if `dialoguer::MultiSelect` can't express non-selectable rows. The implementer landed on hand-rolled `console::Term` directly — a justified choice: dialoguer `MultiSelect` has no model for a non-selectable disabled row (the cursor can't skip a "disabled" entry), and overriding `ColorfulTheme`'s defaults to avoid the blue-bold skin is brittle. `console` was already transitive via dialoguer; declaring it explicitly is the right move. The hand-rolled renderer is ~95 lines of straightforward code and matches FR-5.3's branding requirement (copper/green caret, `[x]` / `[ ]` / `[-]` marks, dim non-selectable rows). **Decision is sound; I endorse it.** The only caveat is that the written-up rationale belongs in the tracked `docs/onboarding-tui-reference.md` (see Non-blocker #4).
- **`exec()` vs `Command::status()`.** The sprint plan (S6-3 technical context) is explicit: "Use `std::process::Command` with `exec()` semantics (via `std::os::unix::process::CommandExt::exec` to fully replace the current process) so the TUI takes over the terminal cleanly." The implementer used exactly that. `Command::status()` was never specified; using `exec()` is the correct, spec-matching choice. The tradeoff — no post-TUI logic can run in the wizard — is deliberate: FR-3.6 says the agent-setup TUI *takes over the terminal*, and the wizard's happy-path work is already complete by the time the drop-through fires. **`exec()` is the right call.**

## Potential Bugs

### Non-blockers

1. **`--print` with no agent arg silently launches the TUI.** `run_with_env_to_writer_inner` only routes `--print` into `install_to_writer` when `opts.agent` is `Some(name)`. If a caller does `aimx agent-setup --print` (no agent), the current code launches the TUI on a TTY, where `sub_opts` forces `print: false`. Historically `--print` without an agent was a no-op (it printed the registry), so the behavior change is small, but the PR body claims "`--list` and `--print` keep current non-TUI behaviour" — which is accurate for `--list` but not strictly for `--print`. Consider either (a) treating `--print` as an implicit `--no-interactive` when `agent` is `None`, or (b) documenting the "must name an agent" requirement in the clap help.

2. **Empty registry / all-`NotInstalled` TUI has no clean cancel path.** If every row is `NotInstalled` (pristine machine with no AI-agent dirs), `interact` puts the cursor at index 0 (a non-selectable row), and Space does nothing. Enter returns an empty selection → "No agents selected. Nothing to do." That's fine behavior, but the TUI never signals *why* Space is inert. Consider short-circuiting with a friendly "No AI agents detected on this machine — install one first" message before entering `interact`. Not a bug per se; a UX non-blocker.

3. **`detect_install_state` reports `InstalledWired` purely on directory existence.** FR-5.2 says "destination file exists **and contains aimx's MCP entry**." The implementation checks only `dest.exists()`. In practice this is fine — `dest_template` (e.g., `$HOME/.claude/plugins/aimx`) is a directory aimx itself creates, so its presence is a strong proxy for "aimx is wired." But an orphan empty directory (failed cleanup, partial `mkdir -p`, user-touched path) would misreport as wired. Tightening detection to `dest.exists() && dest.read_dir()?.next().is_some()` — or a per-agent content check — would match FR-5.2 literally. Flagged as non-blocker because the current check is low-risk and cleanup paths on `main` don't leave orphans.

4. **`docs/onboarding-tui-reference.md` AC not committed.** As described above, the reference file exists locally but is untracked in the aimx-docs repo. Content is mirrored in `book/agent-integration.md` and `src/agent_tui.rs`, so the information is preserved. Resolution is either commit the file to `aimx-docs` (preserves the original AC wording and matches Sprint 1's pattern), or update the sprint plan AC to point to `book/agent-integration.md`. Either works; flagging so the orchestrator can pick one.

5. **Status-column width uses byte-length on ANSI-colored strings.** In the summary table, `status_col` is sized for `"failed".len() + 2`, and the colored tokens `✓ ok` / `✗ failed` carry ANSI escape codes that Rust's `std::fmt` counts toward the width. The visible column width will be shorter than the padding calculates, so the `Destination` column on colored terminals will be slightly mis-aligned. Cosmetic.

6. **Double-wrap of `OverrideHomeEnv` inside the TUI's per-agent loop.** When `--dangerously-allow-root` is passed, `run_with_env_to_writer` wraps the ambient env in `OverrideHomeEnv` pointing at `/root`. Inside `run_tui`, each selected row calls `run_with_env(sub_opts, env)` — but `env` is already the wrapped env, and `sub_opts` still has `dangerously_allow_root: true`, so the sub-call re-wraps a second time. Functionally identical (both wraps point at `/root`), just a micro-waste of a `PathBuf` clone. Worth documenting or short-circuiting in a follow-up.

7. **TUI renders to `Term::stderr()` but TTY detection checks stdout.** `is_stdout_tty()` gates whether to launch the TUI, but the renderer writes to `Term::stderr()`. On a TTY-stdout / non-TTY-stderr invocation (e.g., `aimx agent-setup 2>/dev/null`), the TUI launches and draws into `/dev/null` while `read_key` still works against the underlying TTY. Unusual; flag as non-blocker.

## Test Coverage

Solid. Every non-interactive seam is unit-tested:
- `detect_install_state` — five tests covering empty home, agent-root present, dest-wired, custom XDG, and the goose recipe-dir case.
- `build_rows` — three tests covering the three install states and the default-selection mapping.
- `next_selectable` — two tests (skip logic + all-NotInstalled degenerate case).
- `OverrideHomeEnv` — two tests (method delegation + detection-through-override).
- Root gate — two tests (refusal message contents + bypass with flag).
- `build_agent_setup_argv` — four tests (no `--data-dir`, explicit path, regression-guard against `--dangerously-allow-root`, `/proc/self/exe` source literal).

Gaps (not blocking):
- `interact` itself has no unit test. The logic is simple (arrow keys, space, enter, cancel), but cursor-wrap and space-on-non-selectable behavior could be exercised with a fake `Term`.
- No test asserts that `run_tui` produces the summary table in the expected shape (columns, headers, mark placement). This would catch future regressions in the summary-row formatting.
- No test confirms that `sub_opts` inside `run_tui` forces `print: false` and `list: false`.

## Code Quality (Non-blocker)

- `run_tui` comment says "Two-column table" but the table has three columns (Agent, Status, Destination). Minor doc nit.
- `ROOT_REFUSAL_MESSAGE` uses `\x20\x20` for indent — readable enough, but ordinary spaces in a raw string would be easier to edit.
- The `#[cfg(test)] mod tests { ... FakeEnv ... _responses: RefCell<Vec<String>> }` field is unused (name starts with `_`, good practice signal); could be removed entirely.
- `build_agent_setup_argv_uses_proc_self_exe_shape` asserts via `include_str!("setup.rs").contains("\"/proc/self/exe\"")` — works as a regression guard, but brittle to comment-string matches too. A small seam (return the `PathBuf::from("/proc/self/exe")` from a helper) would be cleaner.

## Alignment with PRD

Implementation tracks FR-3.6, FR-5.1, FR-5.1.a, FR-5.2, FR-5.3, FR-5.4, and FR-5.5 correctly with one deviation: FR-5.2's "destination file exists **and contains aimx's MCP entry**" is softened to "destination directory exists" (see Non-blocker #3). Everything else is spec-faithful.

- **Overall verdict:** Needs minor fixes (one AC miss + a handful of small correctness / UX items). The Sprint 6 goal is delivered cleanly and the core TUI + drop-through work is merge-worthy; the miss is scope around where the TUI reference doc was committed.

- **Blockers:** none.
- **Non-blockers:**
  1. `--print` with no agent arg silently launches the TUI (behavior-change vs stated PR body).
  2. Empty-registry TUI has no user-facing explanation when Space is inert.
  3. `detect_install_state` checks dir existence only; FR-5.2 asks for a content check.
  4. `docs/onboarding-tui-reference.md` AC target is uncommitted in the aimx-docs repo (mirror content is in `book/agent-integration.md`).
  5. Status-column width mis-calculated against ANSI-colored tokens.
  6. Double-wrap of `OverrideHomeEnv` inside the TUI's per-agent loop.
  7. TUI draws on `Term::stderr()` while TTY detection checks stdout.

- **Nice-to-haves:**
  - Unit-test `interact`'s cursor + space-on-non-selectable behavior against a fake `Term`.
  - Snapshot test for the summary-table layout.
  - Friendlier "no agents detected" message in the all-NotInstalled case.
  - Replace the `include_str!` regression guard with a direct-return helper.

uzyn:
## Review non-blockers addressed (ded83ea)

Fixed all 7 non-blockers from the review. Summary of what landed:

- **N1** `--print` with no agent no longer silently launches the TUI. Now implicit `--no-interactive`: prints the registry and hints that a real dry-run preview needs an agent name.
- **N2** All-`NotInstalled` short-circuits with a visible banner listing the probed agents, instead of rendering an inert widget where Space does nothing.
- **N3** `detect_install_state` now enforces FR-5.2's content check.
  - **Per-agent strategy:** For directory-shaped destinations (Claude Code, Codex, OpenCode, Gemini, OpenClaw, Hermes) — `dest` exists AND a top-level file inside contains the substring "aimx". aimx's installed `SKILL.md` / `plugin.json` / skill headers always reference "aimx" in their content, so real installs hit reliably. Orphan empty directories now report `InstalledNotWired` instead of `InstalledWired`. Went with the grep-for-"aimx"-substring approach the review suggested rather than per-agent JSON/TOML parsers — the check runs per agent per TUI render loop and substring scanning keeps cost bounded.
  - For Goose (recipes-directory-shaped destination): wired = `aimx.yaml` exists inside the recipes dir.
  - New regression tests cover: orphan empty dest → NotWired, dest with unrelated file → NotWired, dest with aimx content → Wired, goose without aimx.yaml → NotWired, goose with aimx.yaml → Wired.
- **N4** `docs/onboarding-tui-reference.md` committed to `aimx-docs` at [`08496fb`](https://github.com/uzyn/aimx-docs/commit/08496fb2bc8866d06fa7cd3af5c7263db9dc06d8). Content covers target visual, row-state table, keybindings, crate-decision rationale (hand-rolled `console` vs `dialoguer`), post-Enter summary-table layout, and a tree-tracked ANSI capture at `img/onboarding-tui-reference.txt`. Same pattern as Sprint 1's `docs/release-process.md` (SHA `1d082d8` in aimx-docs).
- **N5** Summary-table columns now measured with `console::measure_text_width` and padded via a visible-width `pad_visible` helper. Added `summary_table_columns_align_with_ansi_escapes` regression test which forces colors on and asserts the Destination column's visible byte offset matches across header + data rows.
- **N6** Sprint 6 TUI's per-agent loop now calls `run_with_env_post_gate` directly. Root gate + `OverrideHomeEnv` wrap runs once per top-level invocation instead of being re-entered for every selected row.
- **N7** TTY detection for "should we launch the TUI" now probes stderr (where the TUI draws via `console::Term::stderr()`) rather than stdout. Renamed the helper `is_tty_for_tui` to make the intent explicit.

Nice-to-haves from the review: skipped — each one exceeded the 5-min threshold once N3 tests were wired up (two of them were direct test additions N5 already covers).

### Test results
- `cargo test` — 1313 unit tests pass (5 new), 95 integration pass, 0 failed

uzyn (commented):
## Re-review (Cycle 2)

All 7 non-blockers from Cycle 1 resolved cleanly, CI green (run `24850068046`), 1313 unit + 95 integration tests pass.

## Resolved issues

- **N1 — `--print` with no agent.** `run_with_env_to_writer_inner` now includes `opts.print` in the non-interactive short-circuit condition and prints a tailored hint (`Pass an agent name (e.g. \`aimx agent-setup --print claude-code\`) to preview the files an install would lay down`). Behaviour matches the PR body's claim about `--print` keeping current non-TUI semantics. Fixed cleanly.
- **N2 — All-`NotInstalled` TUI.** `run_tui` now short-circuits when every row is non-selectable with a banner that names every probed agent and tells the operator to install one and re-run. New unit test `run_tui_prints_guidance_when_no_agents_detected` exercises the path with an empty temp home. Correct.
- **N3 — `detect_install_state` content check.** `dest_contains_aimx_entry` now enforces FR-5.2. Per-agent logic is appropriate:
  - Directory destinations (`claude-code` / `codex` / `opencode` / `gemini` / `openclaw` / `hermes`): exists + at least one top-level regular file contains the byte substring `aimx`. Verified against the bundled agent source trees — every one has a top-level `README.md` or `SKILL.md.header` that mentions "aimx", so real installs always hit.
  - Goose (shared recipes dir): presence of `aimx.yaml` inside `recipes/` — the only file aimx actually writes there. Correct — a filename marker is the right choice when the directory is shared with the user's own recipes.
  - Three new regression tests (orphan empty dest → NotWired, dest with unrelated file → NotWired, goose without `aimx.yaml` → NotWired) plus the existing wired-case tests updated to seed real aimx content.
- **N4 — `onboarding-tui-reference.md`.** Committed to `aimx-docs` at SHA `08496fb2` (verified via `gh api repos/uzyn/aimx-docs/commits/...`). The commit adds `onboarding-tui-reference.md` + `img/onboarding-tui-reference.txt` at the repo root. The `docs/` prefix in the Cycle 1 AC wording was sprint-plan shorthand; the `aimx-docs` repo is flat (Sprint 1's `release-process.md` lives at the root too, same pattern). AC satisfied.
- **N5 — ANSI-aware column widths.** `render_summary_table` + `pad_visible` + `console::measure_text_width` replace the byte-length arithmetic. New `summary_table_columns_align_with_ansi_escapes` test forces colors on via `colored::control::set_override(true)` and asserts that the visible byte offset of the `Destination` column is identical on header and both data rows. Test passes; the fix is solid.
- **N6 — Double-wrap of `OverrideHomeEnv`.** `run_with_env_post_gate` is the new post-root-gate entry point; the TUI's per-agent loop calls it directly so the root gate + `OverrideHomeEnv` wrap only runs once per top-level invocation. The doc comment on `run_with_env_post_gate` explicitly warns that it bypasses the refusal, which is the right thing to write down.
- **N7 — TTY detection on stderr.** `is_stdout_tty` renamed to `is_tty_for_tui` and now probes `io::stderr().is_terminal()`. Matches where `Term::stderr()` writes. Module doc in `agent_tui.rs` updated to point at the renamed helper.

## Still unresolved

None.

## New issues found

No blockers. A few nice-to-have notes for future hardening of the N3 substring scan:

- **Unbounded file read.** `dest_contains_aimx_entry` calls `std::fs::read(&path)` on every top-level regular file with no size cap. A corrupted state (or a user stashing a huge file under `~/.claude/plugins/aimx/`) would be read fully into memory on every TUI render. In practice the destination paths are aimx-specific, so this is very low risk — but a `std::fs::File::open` + bounded-buffer scan (or `metadata().len() > N` early-reject) would be a cheap hardening.
- **Loose substring match.** The reviewer's original concern stands in theory: any byte sequence `aimx` anywhere in any top-level file reports Wired. Given the destination dir names literally end in `/aimx`, a user would have to deliberately drop unrelated content there to trigger a false positive, so this is acceptable. If the roster ever adds an agent whose dest is a shared directory (like Goose's `recipes`), that agent will need its own marker-file check the same way Goose does.
- **Non-recursive scan.** Top-level only. All bundled sources currently ship a top-level `README.md` or `SKILL.md.header` that mentions `aimx`, so every real install hits. If a future bundle removes the top-level marker files and only ships content under `skills/` or `.claude-plugin/`, detection would regress silently. The four detect tests seed a top-level `plugin.json` / `README.md`, so they would not catch that regression. Worth keeping in mind when the plugin bundles change.

## Summary

Sprint 6 is ready to merge. All seven Cycle 1 non-blockers are addressed with matching regression tests, the docs-AC file now lives in the `aimx-docs` repo, and CI is green. The N3 content-check strategy (substring scan for directory destinations, filename marker for Goose) is appropriately scoped to the aimx-specific destination layout and unlikely to false-positive in practice.

```
[Onboarding Sprint 6] aimx agent-setup TUI + drop-through (#139)

Implements the interactive checkbox TUI for `aimx agent-setup`, the
`--dangerously-allow-root` escape hatch, and the `runuser -u $SUDO_USER
-- /proc/self/exe agent-setup` drop-through from `aimx setup`.

- Hand-rolled renderer on `console::Term` (selectable / non-selectable
  rows, colored caret, dim "not detected" entries, per-agent summary
  table with ANSI-aware column widths).
- `detect_install_state` enforces FR-5.2's content check: directory
  destinations must contain a top-level file referencing "aimx";
  Goose's shared recipes dir must contain `aimx.yaml`.
- Root gate applies uniformly to TUI / per-agent / `--no-interactive`;
  `--dangerously-allow-root` swaps in an `OverrideHomeEnv` pointing at
  `/root` via `getpwnam(3)`.
- `aimx setup` replaces the Sprint 5 placeholder with a real `exec`
  drop-through; `--data-dir` is threaded through only when explicit.
- `AIMX_NONINTERACTIVE=1` skips the drop-through; `$SUDO_USER` unset
  prints FR-3.6 guidance and exits cleanly.
- Reference doc committed to `aimx-docs` at SHA `08496fb2`.

Closes Sprint 6 of the onboarding track (PRD §6.5 FR-5.1–5.5, §8.5).
```

PR #144, opened 2026-04-24
[Sprint 2] DATA partial-success contract

uzyn:
## Summary

Closes Sprint 2 of the `hardening` track (PRD §6.3, milestone M3). Multi-recipient outbound mail can no longer return `250 OK` while silently dropping some recipients.

- **S2-1 — RCPT-time mailbox-routing preflight.** New `resolve_recipient_mailbox(&Config, &str) -> Option<String>` helper in `src/ingest.rs`, shared by `handle_rcpt_to` and `ingest_email` so RCPT-time and DATA-time decisions never drift. Unknown recipients are now rejected at RCPT time with `550 5.1.1 <addr>: Recipient address rejected: User unknown`. Catchall (`*@<domain>`) behaviour is preserved. The stale `Config::resolve_mailbox` (which returned `"catchall"` even when no catchall was configured) is removed.
- **S2-2 — `451 4.3.0` on any post-354 failure.** `deliver_message` returns `451 4.3.0 Temporary failure on partial accept; please retry` whenever `failed > 0`, regardless of `succeeded` count. The all-fail path keeps its existing `451 Temporary failure, please retry`. All-success remains `250 OK`. A structured `tracing::info!` line on the `aimx::smtp` target carries `peer`, `from`, `rcpts`, `succeeded`, `failed`, `size` for every delivery outcome (§7.3).
- **S2-3 — Mixed-success test + ingest test seam.** `AIMX_TEST_INGEST_FAIL_FOR=<rcpt>` env var in `ingest_email` forces a synthetic per-recipient failure. Mirrors `AIMX_TEST_MAIL_DROP` in `src/transport.rs`.

## Test plan

- [x] `cargo test` (1433 tests, 0 failures, pre-existing `ignored` tests remain ignored)
- [x] Unit tests for `resolve_recipient_mailbox` (known local part, catchall fallback, no-catchall reject, known-without-catchall)
- [x] `ingest_email` rejects an unknown recipient when no catchall is configured
- [x] New SMTP integration tests:
  - `test_rcpt_rejects_unknown_mailbox_no_catchall` — RCPT is rejected with `550 5.1.1`
  - `test_rcpt_catchall_accepts_any_local_part` — any local part accepted when catchall is present
  - `test_partial_success_returns_451` — alice succeeds, bob is force-failed, response is `451 4.3.0`, alice's `.md` landed, bob's did not
  - `test_multi_rcpt_all_success_returns_250` — both mailboxes land a file, response is `250`
- [x] Existing `test_ingest_failure_returns_451` still passes (updated to seed `alice` so the S2-1 preflight accepts the RCPT and the all-fail path is exercised end-to-end)
- [x] Env-var-bearing tests are `#[serial_test::serial]` so the seam doesn't leak across parallel SMTP tests

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1477,14 +1477,6 @@ impl Config {
     pub fn sent_dir(&self, name: &str) -> PathBuf {
         self.data_dir.join("sent").join(name)
     }
-
-    pub fn resolve_mailbox(&self, local_part: &str) -> String {
-        if self.mailboxes.contains_key(local_part) {
-            local_part.to_string()
-        } else {
-            "catchall".to_string()
-        }
-    }
 }
 
 /// Shared, swappable handle to the daemon's in-memory `Config`.
@@ -2180,18 +2172,6 @@ dangerously_support_untrusted = true
         config.save(&path).unwrap();
         let loaded = Config::load_ignore_warnings(&path).unwrap();
         assert_eq!(config, loaded);
-    }
-
-    #[test]
-    fn resolve_mailbox_known() {
-        let config: Config = toml::from_str(sample_toml()).unwrap();
-        assert_eq!(config.resolve_mailbox("support"), "support");
-    }
-
-    #[test]
-    fn resolve_mailbox_unknown_falls_to_catchall() {
-        let config: Config = toml::from_str(sample_toml()).unwrap();
-        assert_eq!(config.resolve_mailbox("unknown"), "catchall");
     }
 
     #[test]
--- a/src/ingest.rs
+++ b/src/ingest.rs
@@ -13,6 +13,30 @@ use std::net::IpAddr;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
+/// Decide whether a recipient address routes to a configured mailbox.
+///
+/// Returns `Some(mailbox_name)` when either a concrete mailbox matches
+/// the local part or a catchall (`*@<domain>`) exists; `None` when no
+/// mailbox can accept the recipient. Called from both the SMTP
+/// `handle_rcpt_to` preflight (so unknown recipients are rejected at
+/// RCPT time with `550 5.1.1`) and from `ingest_email` itself (so the
+/// RCPT-time decision and the DATA-time decision never drift).
+///
+/// The caller is responsible for checking that the domain matches
+/// `config.domain`; this helper is agnostic about the recipient's
+/// domain and only looks at the local part against the configured
+/// mailboxes.
+pub fn resolve_recipient_mailbox(config: &Config, rcpt: &str) -> Option<String> {
+    let local_part = extract_local_part(rcpt);
+    if config.mailboxes.contains_key(local_part) {
+        return Some(local_part.to_string());
+    }
+    if config.mailboxes.contains_key("catchall") {
+        return Some("catchall".to_string());
+    }
+    None
+}
+
 pub fn run(rcpt: &str, config: Config) -> Result<(), Box<dyn std::error::Error>> {
     // Install the shared tracing subscriber so the INFO/WARN lines emitted
     // below land on stderr for the manual stdin path. `aimx serve` installs
@@ -48,15 +72,33 @@ pub fn ingest_email(
     received_from_ip: IpAddr,
     envelope_mail_from: Option<&str>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let local_part = extract_local_part(rcpt);
-    let mailbox = config.resolve_mailbox(local_part);
+    // Test-only seam: force a synthetic ingest failure for a specific
+    // recipient so we can drive the DATA partial-success contract
+    // (§6.3) in integration tests. Mirrors the shape of
+    // `AIMX_TEST_MAIL_DROP` in `src/transport.rs`. A no-op in
+    // production — any operator setting this in `aimx serve` would see
+    // every matching inbound message rejected, which is louder than
+    // silent.
+    if let Ok(target) = std::env::var("AIMX_TEST_INGEST_FAIL_FOR")
+        && !target.is_empty()
+        && target.eq_ignore_ascii_case(rcpt)
+    {
+        return Err(format!("AIMX_TEST_INGEST_FAIL_FOR forced failure for rcpt={rcpt}").into());
+    }
+
+    let mailbox = match resolve_recipient_mailbox(config, rcpt) {
+        Some(m) => m,
+        None => {
+            return Err(format!("no mailbox routes recipient {rcpt}").into());
+        }
+    };
     let inbox_dir = config.inbox_dir(&mailbox);
 
     std::fs::create_dir_all(&inbox_dir)?;
     // Chown the mailbox directory to the configured owner (PRD §6.3).
-    // Missing mailbox config (which shouldn't happen since resolve_mailbox
-    // falls back to "catchall", which is always present) would make the
-    // chown a silent no-op; `get` returns None in that case.
+    // `resolve_recipient_mailbox` already guaranteed the mailbox entry
+    // exists, so `get` returns `Some` on the happy path; the `if let`
+    // is a belt-and-braces guard.
     if let Some(mb_cfg) = config.mailboxes.get(&mailbox) {
         // The inbox directory may already exist from a prior delivery.
         // We still re-apply the chown/chmod here to heal any drift —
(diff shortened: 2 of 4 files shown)


uzyn (commented):
## Sprint Goal Assessment

The sprint goal — "multi-recipient outbound mail can no longer return `250 OK` while silently dropping some recipients" — is met. RCPT-time preflight now rejects unroutable recipients before DATA is accepted; any post-`354` per-recipient failure surfaces as `451 4.3.0` so senders retry; a structured `aimx::smtp` log line carries the full per-message outcome; and there is a regression test that proves a forced mid-ingest failure drives the `451` response with one inbox filled and the other empty. All three stories (S2-1, S2-2, S2-3) are delivered with unit + integration coverage, `cargo test` reports 1433 tests passing, clippy is clean, fmt is clean.

**[S2-1] RCPT-time mailbox-routing preflight**
- [x] `resolve_recipient_mailbox(&Config, &str)` helper exists in `src/ingest.rs:29`; `ingest_email` (`src/ingest.rs:89`) calls it instead of inline routing. Note: the PRD suggested returning `Option<&MailboxConfig>`; the implementer picked `Option<String>` (mailbox name). This is a justifiable call since the remaining call sites only need the name, and it sidesteps lifetime threading through the SMTP session — worth a one-liner in the PR description if you want to preempt the PRD-shape question.
- [x] `handle_rcpt_to` (`src/smtp/session.rs:388`) calls the helper after the domain check; on `None`, returns `550 5.1.1 <addr>: Recipient address rejected: User unknown\r\n` and does NOT push to `forward_paths`.
- [x] Catchall path preserved — `test_rcpt_catchall_accepts_any_local_part` asserts `randomuser@test.local` accepts under the standard catchall-present config.
- [x] No-catchall path rejects unknown local parts — `test_rcpt_rejects_unknown_mailbox_no_catchall` proves it.
- [x] Integration test pair exists (the `test_rcpt_rejects_unknown_mailbox_no_catchall` + `test_rcpt_catchall_accepts_any_local_part` pair) per the criterion.
- [x] Existing SMTP tests still pass (`test_rcpt_rejects_unrelated_domain`, `test_rcpt_rejects_subdomain`, etc.).
- [x] `cargo test`, clippy, fmt clean.

**[S2-2] `deliver_message` returns 4xx on any post-354 failure**
- [x] `deliver_message` (`src/smtp/session.rs:600`) returns `451 4.3.0 Temporary failure on partial accept; please retry\r\n` when `failed > 0 && succeeded > 0`.
- [x] Returns `451 Temporary failure, please retry\r\n` when `succeeded == 0` (all-fail path preserved).
- [x] Returns `250 OK message accepted\r\n` only when `failed == 0`.
- [x] `tracing::info!` on `aimx::smtp` target (`src/smtp/session.rs:572-581`) emits `peer`, `from`, `rcpts`, `succeeded`, `failed`, `size` on every outcome. Fires before the branching, so all three response classes hit it.
- [x] clippy / fmt clean.

**[S2-3] Mixed-success integration test + observability assertion**
- [x] Test seam `AIMX_TEST_INGEST_FAIL_FOR` (`src/ingest.rs:82-87`) exists and is documented as test-only. It mirrors the `AIMX_TEST_MAIL_DROP` shape.
- [x] `test_partial_success_returns_451` asserts `451 4.3.0`, alice has exactly 1 md file, bob has 0.
- [x] `test_multi_rcpt_all_success_returns_250` asserts `250` and both mailboxes land a file.
- [x] Existing all-fail case (`test_ingest_failure_returns_451`) still passes (updated to seed `alice` so the S2-1 preflight accepts the RCPT and the path is exercised end-to-end).
- [x] Seam is a no-op when the env var is unset. `eq_ignore_ascii_case` on the rcpt value means case-insensitive match, which is the right default.
- [x] `#[serial_test::serial]` applied to both env-var-bearing tests. See non-blocker note below about a residual leakage risk.

## Test Coverage

Solid for the stories shipped. Unit coverage of `resolve_recipient_mailbox` hits all four useful cases (known / unknown-with-catchall / unknown-without-catchall / known-without-catchall), the ingest rejection path has its own unit test, and the SMTP integration tests cover both the positive and negative RCPT outcomes plus mixed-success end-to-end.

The one test-coverage gap I'd flag (non-blocker):

- **Observability is asserted by log-line inspection only via `eprintln!`**, not by capturing the `tracing::info!` event the PRD §7.3 makes authoritative. The structured log line is now the contract, so a test that exercises a `tracing::subscriber` layer (`tracing_test` or similar) and asserts the event fields (`peer`, `from`, `rcpts`, `succeeded`, `failed`, `size`) would lock the contract in code rather than in PR-review vigilance. Not blocking — the criterion just says "emitted" — but worth filing as an improvement.

## Potential Bugs

I walked the changed code and the env-var seam carefully. Findings:

- **Non-blocker — env-var cleanup on panic is best-effort.** `test_partial_success_returns_451` calls `set_var(...)` at the top and `remove_var(...)` at the bottom. If any assertion between those two lines panics, the env var stays set. The follow-on `test_multi_rcpt_all_success_returns_250` defensively calls `remove_var(...)` at its own top, which covers serial ordering under `serial_test`, but another serial test added later in the same file would have to remember to do the same. This is a maintenance footgun, not a correctness bug today. A `scopeguard::defer!` / RAII wrapper around the env-var would be cleaner. Non-blocker because both current tests are correct as written.

- **Non-blocker — `serial_test` only serializes serial-tagged tests among themselves.** Non-serial tests continue to run concurrently with serial ones. `AIMX_TEST_INGEST_FAIL_FOR` is process-wide, so in principle a parallel non-serial test that ingests `bob@test.local` would observe the forced failure during the window the env var is set. Scanning the suite, no other test ingests `bob@test.local` today (the only non-S2 usage is `src/hooks.rs:1234` which is `bob@test.com`, a different domain and a different code path), so this is latent rather than active. Worth a one-line comment warning future test authors that they must not ingest `bob@test.local` from a non-serial SMTP test. Non-blocker.

- **Non-blocker — `eprintln!` "Message partially accepted" line is preserved alongside the authoritative `tracing::info!` line.** The PRD §7.3 says "The existing `eprintln!` stays as a fallback but the structured log is the authoritative signal," so this matches intent. Worth knowing that `aimx logs` will show both lines for every delivery, which is chatty. Nothing to fix now.

- **Intentional, noting for transparency — partial-success still increments `session_state.message_count` and `total_bytes`, and the files that did land stay on disk.** Sender will retry per RFC, and the retry will re-deliver to alice, producing duplicates in her inbox. The inline comment (`src/smtp/session.rs:591-594`) calls this out explicitly: "the retry will re-deliver duplicates but that is the price of lossless semantics." This is the documented v1 contract per PRD §8.2 (DSNs are deferred). Flagging here so it's on the record.

No logic errors, no crash paths, no data-loss defects. The `if failed == 0 && succeeded > 0` vs `else if succeeded > 0` branching is tight and the `else` handles `succeeded == 0` correctly whether `failed` is zero (impossible with RCPT>0) or positive.

## Security Issues

None. The S2-1 preflight tightens the attack surface (unknown recipients no longer reach ingest at all), the env-var seam is test-only and would be extremely loud in production (every matching ingest rejected), and no new I/O or parsing is introduced on the untrusted path.

## Code Quality

- `resolve_recipient_mailbox` doc comment is good — explicit about what it checks and what the caller must check (domain). Kept as `Option<String>` rather than `Option<&MailboxConfig>`; fine as noted above.
- `handle_rcpt_to` change cleanly reuses the existing `config` snapshot for both the domain and the preflight call — good, avoids a second `config_handle.load()`.
- `deliver_message` restructure is a net improvement: the shared `peer` / `from` / `size` locals and the single authoritative `tracing::info!` line read better than the prior eprintln-per-branch.
- Removal of `Config::resolve_mailbox` is correct — the old signature was the source of the "silently route unknown to catchall even when catchall doesn't exist" bug, and nothing else called it. The two tests for it are correctly deleted.

One minor readability note (non-blocker): `resolve_recipient_mailbox` returns a `Some("catchall".to_string())` for the catchall branch even when a concrete catchall mailbox is keyed differently in the future. The helper hardcodes the string `"catchall"` twice and that's also what `config.mailboxes.contains_key("catchall")` keys on. If the "catchall" key name ever becomes a constant elsewhere (`RESERVED_RUN_AS_CATCHALL` is the closest equivalent and is something different), re-use it here. Not blocking.

## Alignment with PRD

Matches §6.3 requirement (a) — preflight RCPT + `4xx` on post-354 failure. §7.3's structured log field set (`from`, `rcpts`, `succeeded`, `failed`, `size`) is all present; `peer` is an extra useful field, not a miss. No scope creep: config parsing, frontmatter, storage layout, and CLI surface are untouched.

One PRD nit: §7.3 says the structured log line "stays as a fallback but the structured log is the authoritative signal" — the implementation keeps `eprintln!` as a supplement on all three branches (accepted / partially accepted / rejected). The partial-accepted `eprintln!` is redundant with the `tracing::info!` line that now fires unconditionally. Harmless duplication — non-blocker.

All three stories meet their acceptance criteria. The code is correct, the tests cover the stated contract, and nothing regresses. The notes above are quality-of-life items, not defects.

**Blockers:** none.

**Non-blockers (file as improvements in the Sprint backlog, do not block this merge):**
- Add a structured-log capture test for the `tracing::info!` delivery outcome event so the §7.3 field contract is enforced by code, not by reviewer eyeball.
- Replace the `set_var` / `remove_var` pattern in `test_partial_success_returns_451` with an RAII guard so a panicked assertion cannot leak the env var to other serial tests.
- Add a one-line comment near the `AIMX_TEST_INGEST_FAIL_FOR` seam warning future test authors that any non-serial SMTP test targeting `bob@test.local` can observe the forced failure; today's suite is safe.

**Nice-to-haves:**
- Consider dropping the redundant `eprintln!` "Message partially accepted" line now that `tracing::info!` is the authoritative signal (aligns with §7.3's intent without losing information).
- If `resolve_recipient_mailbox`'s return type becomes painful at future call sites, switch to `Option<&MailboxConfig>` to preserve the MailboxConfig handle without a second `config.mailboxes.get(...)`.

```
[Sprint 2] DATA partial-success contract

Closes M3 of the hardening track. Multi-recipient outbound mail can no
longer return 250 OK while silently dropping some recipients.

- Shared resolve_recipient_mailbox helper; handle_rcpt_to and
  ingest_email route through the same predicate so RCPT-time and
  DATA-time decisions stay in sync.
- Unknown recipients are rejected at RCPT time with 550 5.1.1;
  catchall behavior preserved.
- deliver_message returns 451 4.3.0 on any post-354 per-recipient
  failure; all-success keeps 250, all-fail keeps 451.
- Structured tracing::info! on the aimx::smtp target carries peer,
  from, rcpts, succeeded, failed, size on every delivery outcome.
- AIMX_TEST_INGEST_FAIL_FOR env seam mirrors AIMX_TEST_MAIL_DROP so
  integration tests can force a synthetic partial failure.
- 1433 tests, 0 failures; clippy and fmt clean.
```

PR #147, opened 2026-04-24
aimx uninstall: delete the binary too

uzyn:
## Summary

- `sudo aimx uninstall` now removes the installed `aimx` binary after it removes the init-system service file. This fixes the dead-end where `curl https://aimx.email/install.sh | sh` refuses to run after an uninstall because the leftover binary still reports a version that trips `install.sh`'s monotonic-upgrade check.
- Binary removal is best-effort: a failure logs a warning and a manual-rm hint rather than rolling back. Service-unit removal is still the authoritative step.
- Config (`/etc/aimx/`) and data (`/var/lib/aimx/`) are still explicitly preserved.

## Repro (before this PR)

```
$ sudo aimx uninstall
… Uninstall complete.
$ curl -fsSL https://aimx.email/install.sh | sh
install: error: installed 0.1.0 is newer than target 0.0.1; run 'aimx upgrade --version 0.0.1 --force' to downgrade explicitly
```

Root cause: `src/uninstall.rs` only called `uninstall_service_file`, leaving `/usr/local/bin/aimx` in place. `install.sh:463-477` then saw the orphan and refused.

## Change surface

- `SystemOps::remove_file` trait seam added in `src/setup.rs` (real impl: `std::fs::remove_file`). Every other `SystemOps` mock (`doctor.rs`, `logs.rs`, `portcheck.rs`, inline test mock in `setup.rs`, inline test mock in `uninstall.rs`) gets an impl — `unreachable!()` in modules that never touch file removal, recording in the ones that do.
- `uninstall::run` resolves the running binary via `get_aimx_binary_path()` (already canonicalises via `current_exe`) and removes it. Self-delete is safe on Linux — the kernel keeps the inode mapped for the running process.
- Three new/updated tests in `src/uninstall.rs`: happy path asserts the binary path was recorded; two failure-path tests assert uninstall still returns `Ok` when either `get_aimx_binary_path` or `remove_file` errors.
- `book/cli.md` and `book/setup.md` updated to reflect the new behavior.

## Test plan

- [x] `cargo fmt -- --check`
- [x] `cargo test` — 1343 + 95 + 1 pass, 0 fail
- [ ] Manual smoke on a VM: `sudo aimx uninstall` → `which aimx` returns empty → `curl … | sh` installs cleanly without `--force`

### DIFF
--- a/book/cli.md
+++ b/book/cli.md
@@ -39,7 +39,7 @@ See [Setup](setup.md) for the full walkthrough.
 
 ### `aimx uninstall`
 
-Stop the daemon and remove the installed init-system service file. Non-destructive. Leaves `/etc/aimx/` and `/var/lib/aimx/` intact.
+Stop the daemon, remove the init-system service file, and delete the installed `aimx` binary itself so a subsequent `install.sh` run starts from a clean slate. Leaves `/etc/aimx/` and `/var/lib/aimx/` intact — wipe them manually with `sudo rm -rf /etc/aimx /var/lib/aimx` if you want a full purge.
 
 | Flag | Description |
 |------|-------------|
--- a/book/setup.md
+++ b/book/setup.md
@@ -131,13 +131,13 @@ At the DNS verification step:
 
 ### Uninstalling
 
-To reverse `aimx setup`, stop the daemon and remove its init-system service file:
+To reverse `aimx setup`, stop the daemon, remove its init-system service file, and delete the installed `aimx` binary:
 
 ```bash
 sudo aimx uninstall
 ```
 
-Pass `--yes` to skip the confirmation prompt. Uninstall is intentionally non-destructive: it leaves your config (`/etc/aimx/`) and mailbox data (`/var/lib/aimx/`) in place so a subsequent `aimx setup` reuses them. If you also want to wipe those, remove them manually with `rm -rf`.
+Pass `--yes` to skip the confirmation prompt. Uninstall is intentionally scoped: it removes the service and the binary so a subsequent `install.sh` run starts fresh, but leaves your config (`/etc/aimx/`) and mailbox data (`/var/lib/aimx/`) in place so a subsequent `aimx setup` reuses them. If you also want to wipe those, remove them manually with `rm -rf`.
 
 ## DNS configuration
 
--- a/src/doctor.rs
+++ b/src/doctor.rs
@@ -1901,6 +1901,9 @@ mod tests {
         fn uninstall_service_file(&self) -> Result<(), Box<dyn std::error::Error>> {
             unreachable!("gather_status must not touch uninstall_service_file")
         }
+        fn remove_file(&self, _path: &Path) -> Result<(), Box<dyn std::error::Error>> {
+            unreachable!("gather_status must not touch remove_file")
+        }
         fn wait_for_service_ready(&self) -> bool {
             unreachable!("gather_status must not touch wait_for_service_ready")
         }
--- a/src/logs.rs
+++ b/src/logs.rs
@@ -143,6 +143,9 @@ mod tests {
         fn uninstall_service_file(&self) -> Result<(), Box<dyn std::error::Error>> {
             unreachable!("logs::run must not touch uninstall_service_file")
         }
+        fn remove_file(&self, _path: &Path) -> Result<(), Box<dyn std::error::Error>> {
+            unreachable!("logs::run must not touch remove_file")
+        }
         fn wait_for_service_ready(&self) -> bool {
             unreachable!("logs::run must not touch wait_for_service_ready")
         }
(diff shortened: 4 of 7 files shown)


uzyn (commented):
## Changes Overview

Adds a `remove_file` seam to `SystemOps` (real impl is a thin `std::fs::remove_file` wrapper; mocks either record or `unreachable!()`) and wires `uninstall::run` to resolve the running binary via the already-canonicalising `get_aimx_binary_path` and delete it best-effort after the service unit is gone. Three new/updated tests cover the happy path and both failure paths. Book docs updated to describe the new behavior.

## Scope Alignment

Fully matches the stated intent. No scope creep — the diff is tightly limited to the fix and the necessary trait-impl boilerplate across the other mocks. The canonicalising self-delete is called out in both the PR description and a code comment.

One minor gap: the PR's own test plan has the manual smoke-on-a-VM step unchecked. The fix is simple enough that this is low risk, but the reviewer should not assume it has been verified end-to-end against a real `install.sh` run.

## Potential Bugs

**Non-blocker — confirmation prompt no longer matches behavior.** `src/uninstall.rs:13-19`:

```
println!("This will stop the aimx daemon and remove its service file.");
println!(
    "Config ({}) and mailbox data ({}) will be kept.",
    config_dir.display(),
    data_dir
);
```

The prompt the user reads before typing `y` still advertises only service-file removal. After this PR, uninstall also deletes `/usr/local/bin/aimx` — which the user only learns about *after* confirming, when `"Removed binary: …"` is printed. Consent is collected on stale information. No data loss (config + mailbox data remain), and the success banner does show what happened, but the prompt should describe what it's actually about to do. A one-line change to the prompt text closes this.

**Nice-to-have — symlink edge case silently leaves a dangling symlink.** `get_aimx_binary_path` canonicalises via `std::env::current_exe()`, so on an install where `/usr/local/bin/aimx` is a symlink to the real binary elsewhere (`/opt/aimx/bin/aimx`, a distro package, etc.), `remove_file` unlinks only the canonical target. `/usr/local/bin/aimx` remains as a dangling symlink. This does **not** break the stated fix — `install.sh:463` does `[ -x "${_install_path}" ]`, which returns false on a dangling symlink, so reinstall still proceeds cleanly. But the user's `which aimx` will show a broken path until they clean it up. For the common install-via-install.sh case (which writes `/usr/local/bin/aimx` directly, no symlink), this does not apply. Worth a code comment acknowledging the asymmetry, or a follow-up to also unlink the invocation path when it differs from the canonical one.

## Security Issues

None. The only new filesystem write is a `remove_file` on a path derived from `current_exe()` — not attacker-controlled, already canonicalised, and root-only (uninstall rejects non-root at line 7).

## Test Coverage

Unit coverage is appropriate: happy path asserts the recorded path equals `/usr/local/bin/aimx`; failure-path tests assert `run` still returns `Ok` when either `get_aimx_binary_path` or `remove_file` errors. The structured `Option<PathBuf>` return from the new block is also exercised indirectly (empty `removed_files` vec on failure).

**Gap — no test for the prompt-vs-behavior mismatch.** If the prompt is updated per the non-blocker above, that copy should be asserted somewhere so a future "revert the wording" change fails loudly. A single `assert!` on captured stdout (or on a refactored string constant) would suffice.

**Gap — `RealSystemOps::remove_file` is untested.** It is a trivial one-liner over `std::fs::remove_file`, so this is defensible; calling it out only because no integration test exercises the full invoke-the-real-binary uninstall path either. If a user-facing regression slipped in (say, a future refactor that accidentally calls `remove_dir_all`), none of the current mocks would catch it.

## Code Quality

- The `Option<PathBuf>` threading in `uninstall::run` is clean and the branch symmetry reads well.
- The trait-method doc block on `SystemOps::remove_file` explicitly names `aimx uninstall` as the only caller — good, because it sets expectations and discourages misuse as a general-purpose file seam.
- Minor: the other mock impls (`doctor.rs`, `logs.rs`, `portcheck.rs`) all have `unreachable!` with a module-scoped message, which is consistent with how `stop_service` / `start_service` are already handled in those files. The setup.rs `MockSystemOps::remove_file` silently records instead of `unreachable!` — that's defensible as it's a general-purpose mock, but if no current setup test uses it, `unreachable!` would surface unintended future coverage.

**Verdict:** Needs minor fixes.

**Blockers:** none.

**Non-blockers:**
1. Update the confirmation prompt text in `src/uninstall.rs:13-19` to mention binary removal so user consent matches the behavior.

**Nice-to-haves:**
1. Code comment or follow-up for the canonicalised-path asymmetry (dangling-symlink case).
2. Assertion on the updated prompt copy so a future revert is caught.
3. Consider `unreachable!()` instead of recording in the setup.rs mock's `remove_file` if no setup tests currently touch it.

The core fix is correct and the test additions are appropriately scoped. Closing the prompt/behavior mismatch is the one thing I'd want before merging.

PR #171, opened 2026-04-26
Lock mailbox dir perms to <owner>:<owner> 0700 + add isolation test

uzyn:
## Summary

- Mailbox directories (`inbox/<name>/` and `sent/<name>/`) are now locked to `<owner>:<owner> 0700` by both setup and the daemon's `MAILBOX-CREATE` path. Setup verifies the owner and primary group exist on the host before chowning; missing users abort with a clear `useradd` hint. The CLI fallback (`create_mailbox` when the daemon is down) chowns under root and chmods elsewhere.
- The hook fire path (`execute_on_receive` / `execute_after_send`) gains a defense-in-depth guard that refuses to fire when hooks have somehow ended up on a catchall mailbox. Load-time validation already rejects this; the guard catches in-memory bypasses and emits a structured warning.
- New `tests/mailbox_isolation.rs` integration test provisions two real Linux users (`aimx-test-alice`, `aimx-test-bob`), lays out per-mailbox dirs at `0700`, drops a fixture email into bob's inbox, then runs `runuser`-as-alice probes (`/bin/cat`, `/bin/ls`, `/bin/test -r`) and asserts every one fails with `EACCES`. Bob still reads his own file as a sanity check. Gated behind `#[ignore]` + `AIMX_INTEGRATION_SUDO=1`.
- New CI job `mailbox-isolation` provisions the two test users via `useradd`, builds the test binary, and runs it under `sudo` so the kernel-level isolation is exercised on every push. Tear-down runs in `if: always()`.

## Test plan

- [x] `cargo test` (873 unit tests, 78 integration, all crates) — clean.
- [x] `cargo clippy --all-targets -- -D warnings` — clean (root + verifier).
- [x] Manual end-to-end run of the new isolation test under `sudo AIMX_INTEGRATION_SUDO=1` — all three EACCES probes fail correctly; bob's self-read succeeds; teardown removes the test users.
- [ ] CI verifies the new `mailbox-isolation` job runs `useradd` + the test on a hosted runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,9 @@ name: CI
 on:
   push:
 
+permissions:
+  contents: read
+
 env:
   CARGO_TERM_COLOR: always
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,45 @@ jobs:
           fi
           echo "Author metadata check passed"
 
+  mailbox-isolation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Provision test users
+        run: |
+          sudo useradd --system --no-create-home --shell /usr/sbin/nologin aimx-test-alice || \
+            id aimx-test-alice
+          sudo useradd --system --no-create-home --shell /usr/sbin/nologin aimx-test-bob || \
+            id aimx-test-bob
+
+      - name: Build the isolation test binary
+        run: cargo test --no-run --test mailbox_isolation
+
+      - name: Run mailbox isolation test (under sudo)
+        run: |
+          # Pre-built binary path is stable thanks to `--no-run`; locate
+          # the most-recently-built `mailbox_isolation-*` artifact and
+          # run it as root with the sudo gate set so the test body
+          # actually exercises the kernel-level isolation. Skipping
+          # `cargo test` here avoids re-running rustc under sudo with a
+          # cargo home pointing at root's $HOME.
+          BIN=$(ls -t target/debug/deps/mailbox_isolation-* | grep -v '\.d$' | head -n 1)
+          test -x "$BIN"
+          sudo AIMX_INTEGRATION_SUDO=1 "$BIN" --ignored
+
+      - name: Tear down test users
+        if: always()
+        run: |
+          sudo userdel aimx-test-alice || true
+          sudo userdel aimx-test-bob || true
+
   docs-build:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
           fi
           echo "Author metadata check passed"
 
-  mailbox-isolation:
+  mailbox-dir-perms-isolation:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
--- a/book/configuration.md
+++ b/book/configuration.md
@@ -101,19 +101,15 @@ Trust only gates hook execution (`on_receive`). All email is stored regardless o
 
 ### Hook settings
 
-Hooks are defined as `[[mailboxes.<name>.hooks]]` arrays. Each hook is either **template-bound** (`template = "..."`, `params = {...}`) or **raw-cmd** (`cmd = "..."`). Both flavours run sandboxed as the mailbox's `owner` by default (catchall hooks default to the reserved `aimx-catchall` user).
+Hooks are defined as `[[mailboxes.<name>.hooks]]` arrays. Each hook is a **raw-cmd** stanza: `cmd` is an argv array exec'd directly as the mailbox's `owner`. The first element must be an absolute path; there is no shell wrapping. If you need shell expansion, spell out `cmd = ["/bin/sh", "-c", "..."]` explicitly.
 
 | Setting | Type | Description |
 |---------|------|-------------|
-| `name` | string | Optional. Matches `^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$`. When omitted, aimx derives a stable 12-char hex name from `sha256(event + cmd + dangerously_support_untrusted)` (raw-cmd) or `sha256(event + template + sorted_params)` (template-bound). Names must be globally unique across mailboxes — including derived ones. |
+| `name` | string | Optional. Matches `^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$`. When omitted, aimx derives a stable 12-char hex name from `sha256(event + joined_argv + fire_on_untrusted)`. Names must be globally unique across mailboxes — including derived ones. |
 | `event` | string | `"on_receive"` or `"after_send"` |
 | `type` | string | Hook kind, default `"cmd"` (only `cmd` is supported today) |
-| `cmd` | string | Raw shell command. Required for raw-cmd hooks; forbidden for template-bound hooks. |
-| `template` | string | Name of a `[[hook_template]]` to bind to. Mutually exclusive with `cmd`. |
-| `params` | table | Bound parameter values for template-bound hooks. Keys must match the template's declared `params`; unknown keys rejected. |
-| `dangerously_support_untrusted` | bool | `on_receive` only: fire even when `trusted != "true"`. Rejected on hooks with `origin = "mcp"`. |
-| `run_as` | string | Any existing Linux username; must equal the mailbox's `owner` (catchall exception: `"aimx-catchall"`) or `"root"`. Defaults to the mailbox's `owner` when omitted. `"root"` is only settable via hand-edit of `config.toml` — not via CLI or MCP. |
-| `origin` | string | `"operator"` (default) or `"mcp"`. Stamped by the daemon based on submission channel. |
+| `cmd` | array of strings | Argv exec'd directly. Required and non-empty; `cmd[0]` must be an absolute path. There is no shell wrapping — spell out `["/bin/sh", "-c", "..."]` explicitly when you need shell expansion. |
+| `fire_on_untrusted` | bool | `on_receive` only: fire even when `trusted != "true"`. |
 
 Unknown fields on a hook table are rejected at config load. See [Hooks & Trust](hooks.md) for full details on events and trust policies.
 
@@ -258,8 +254,8 @@ address = "*@agent.yourdomain.com"
 [[mailboxes.catchall.hooks]]
 # name is optional — a stable 12-char hex id is derived from event+cmd if omitted
 event = "on_receive"
-cmd = 'ntfy pub agent-mail "New email: $AIMX_SUBJECT from $AIMX_FROM"'
-dangerously_support_untrusted = true
+cmd = ["/bin/sh", "-c", 'ntfy pub agent-mail "New email: $AIMX_SUBJECT from $AIMX_FROM"']
+fire_on_untrusted = true
 
 # ----------------------------
 # Named mailbox with a per-mailbox trust override
@@ -276,13 +272,13 @@ trusted_senders = ["*@yourcompany.com", "boss@gmail.com"]
 [[mailboxes.support.hooks]]
 name = "support_log"
 event = "on_receive"
-cmd = 'echo "{date} | $AIMX_FROM | $AIMX_SUBJECT" >> /var/log/aimx-support.log'
+cmd = ["/bin/sh", "-c", 'echo "{date} | $AIMX_FROM | $AIMX_SUBJECT" >> /var/log/aimx-support.log']
 
 # Trigger agent on every trusted incoming email
 [[mailboxes.support.hooks]]
 name = "support_agent"
 event = "on_receive"
-cmd = 'claude -p "Process this email: $(cat \"$AIMX_FILEPATH\")"'
+cmd = ["/bin/sh", "-c", 'claude -p "Process this email: $(cat \"$AIMX_FILEPATH\")"']
 
 # ----------------------------
 # Another mailbox
--- a/book/hook-recipes.md
+++ b/book/hook-recipes.md
@@ -91,7 +91,7 @@ trusted_senders = ["*@yourcompany.com"]
 [[mailboxes.inbox.hooks]]
 name = "claudeinbox1"
 event = "on_receive"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 claude -p "You received a new email. Read it and file a summary ticket.
 
 Email path: $AIMX_FILEPATH
@@ -101,7 +101,7 @@ From: $AIMX_FROM
 Use the Read tool to open \"$AIMX_FILEPATH\", then call the appropriate MCP tool." \
   --permission-mode=bypassPermissions \
   >> /var/log/aimx/claude.log 2>&1
-'''
+''']
 ```
 
 - `"$AIMX_FILEPATH"` is always safe. The shell quotes the value, so paths with spaces or unusual characters are handled correctly.
@@ -126,7 +126,7 @@ trust = "verified"
 [[mailboxes.triage.hooks]]
 name = "codextriage1"
 event = "on_receive"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 codex exec "Triage this email and update the issue tracker.
 
 Email file: $AIMX_FILEPATH
@@ -134,7 +134,7 @@ From: $AIMX_FROM
 Subject: $AIMX_SUBJECT" \
   --full-auto \
   >> /var/log/aimx/codex.log 2>&1
-'''
+''']
 ```
 
 - `--full-auto` is the recommended default for hooks: Codex auto-approves tool use but stays inside its sandbox.
@@ -155,13 +155,13 @@ trust = "verified"
 [[mailboxes.research.hooks]]
 name = "ocresearch01"
 event = "on_receive"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 opencode run "A new research email arrived. Read \"$AIMX_FILEPATH\" and append a digest entry to docs/digest.md.
 
 From: $AIMX_FROM
 Subject: $AIMX_SUBJECT" \
   >> /var/log/aimx/opencode.log 2>&1
-'''
+''']
 ```
 
 ## Gemini CLI
@@ -180,11 +180,11 @@ trust = "verified"
 [[mailboxes.notes.hooks]]
 name = "geminines01"
 event = "on_receive"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 gemini -p "A new note arrived by email. Read \"$AIMX_FILEPATH\" and file it into my notes." \
   --yolo \
   >> /var/log/aimx/gemini.log 2>&1
-'''
+''']
 ```
 
 ## Goose
@@ -202,13 +202,13 @@ trust = "verified"
 [[mailboxes.ops.hooks]]
 name = "gooseops001"
 event = "on_receive"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 goose run -t "Ops email arrived. Read \"$AIMX_FILEPATH\", check if action is required, and page on-call if so.
 
 Subject: $AIMX_SUBJECT
 From: $AIMX_FROM" \
   >> /var/log/aimx/goose.log 2>&1
-'''
+''']
 ```
 
 ## OpenClaw
@@ -228,13 +228,13 @@ trusted_senders = ["*@yourcompany.com"]
 [[mailboxes.inbox.hooks]]
 name = "openclaw0001"
 event = "on_receive"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 openclaw agent \
     --message "An email arrived at $(basename \"$AIMX_FILEPATH\"). Read it at \"$AIMX_FILEPATH\", summarize the sender's request, and respond via the aimx MCP tools if a reply is appropriate." \
     --deliver \
     --json \
     >> /var/log/aimx/openclaw.log 2>&1
-'''
+''']
 ```
 
 ## Hermes
@@ -256,8 +256,8 @@ trust = "verified"
 [[mailboxes.hermes.hooks]]
 # name is optional — a stable 12-char hex id is derived from event+cmd if omitted
 event = "on_receive"
-cmd = 'ntfy pub hermes-mail "New email: $AIMX_SUBJECT from $AIMX_FROM"'
-dangerously_support_untrusted = true
+cmd = ["/bin/sh", "-c", 'ntfy pub hermes-mail "New email: $AIMX_SUBJECT from $AIMX_FROM"']
+fire_on_untrusted = true
 ```
 
 When Hermes grows a headless `--message` / `exec`-style CLI, add it here mirroring the Claude Code or Codex recipe.
@@ -280,12 +280,12 @@ trusted_senders = ["*@yourcompany.com"]
 [[mailboxes.bugs.hooks]]
 name = "aiderbugs01"
 event = "on_receive"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 cd /srv/repos/myapp && \
 aider --yes-always \
       --message "Bug report arrived by email. Read \"$AIMX_FILEPATH\", reproduce the issue if possible, apply a fix, and commit." \
       >> /var/log/aimx/aider.log 2>&1
-'''
+''']
 ```
 
 ## `after_send` recipes
@@ -298,7 +298,7 @@ Send-side hooks run after aimx resolves the MX delivery attempt. They cannot aff
 [[mailboxes.alice.hooks]]
 name = "auditlog0001"
 event = "after_send"
-cmd = 'printf "%s %s %s %s\n" "$AIMX_SEND_STATUS" "$AIMX_TO" "$AIMX_SUBJECT" "$AIMX_HOOK_NAME" >> /var/log/aimx/alice-sent.log'
+cmd = ["/bin/sh", "-c", 'printf "%s %s %s %s\n" "$AIMX_SEND_STATUS" "$AIMX_TO" "$AIMX_SUBJECT" "$AIMX_HOOK_NAME" >> /var/log/aimx/alice-sent.log']
 ```
 
 ### Page on failed sends
@@ -307,11 +307,11 @@ cmd = 'printf "%s %s %s %s\n" "$AIMX_SEND_STATUS" "$AIMX_TO" "$AIMX_SUBJECT" "$A
 [[mailboxes.alerts.hooks]]
 name = "failedpage01"
 event = "after_send"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 if [ "$AIMX_SEND_STATUS" != "delivered" ]; then
   ntfy pub on-call "aimx send to $AIMX_TO $AIMX_SEND_STATUS: $AIMX_SUBJECT"
 fi
-'''
+''']
 ```
 
 ### Notify only on matching recipients (shell guard)
@@ -322,14 +322,14 @@ Filter fields on hooks have been removed — do recipient/subject matching in th
 [[mailboxes.marketing.hooks]]
 name = "marknotify01"
 event = "after_send"
-cmd = '''
+cmd = ["/bin/sh", "-c", '''
 case "$AIMX_TO" in
   *@customer-co.com)
     curl -fsS -X POST https://hooks.internal/marketing-sent \
          -d "to=$AIMX_TO&status=$AIMX_SEND_STATUS"
     ;;
 esac
-'''
+''']
 ```
 
 ## Operational tips
--- a/book/hooks.md
+++ b/book/hooks.md
@@ -73,37 +73,39 @@ The CLI sends the request over the same UDS verb MCP uses, so the hook ends up w
 
 ## Raw-cmd hooks (power user)
 
-Raw-cmd hooks let the operator drop an arbitrary shell string directly into `config.toml`. The command is wrapped in `/bin/sh -c` and executed inside the same privilege-dropped sandbox as template hooks. Hook `run_as` defaults to the mailbox's `owner`; the operator can elevate to `run_as = "root"` by hand-editing `config.toml` — this is intentionally not reachable from the CLI or MCP. The reserved `aimx-catchall` value is allowed on hooks attached to the catchall mailbox.
+Raw-cmd hooks let the operator drop an argv array directly into `config.toml`. The argv is exec'd as the mailbox's `owner` inside the privilege-dropped sandbox — there is no shell wrapping. If you need shell expansion (env-var substitution, redirection), spell out `cmd = ["/bin/sh", "-c", "..."]` explicitly.
 
 ```toml
 [[mailboxes.support.hooks]]
 name = "support_notify"
 event = "on_receive"
-cmd = 'echo "New email from $AIMX_FROM: $AIMX_SUBJECT" >> /tmp/email.log'
+cmd = ["/bin/sh", "-c", 'echo "New email from $AIMX_FROM: $AIMX_SUBJECT" >> /tmp/email.log']
 ```
 
 ### Create a raw-cmd hook
 
+`--cmd` takes the argv as a JSON array string. The first element must be an absolute path:
+
 ```bash
-# on_receive with an explicit name
+# on_receive with an explicit name (shell wrapper for env-var expansion)
 sudo aimx hooks create \
   --mailbox support \
   --event on_receive \
-  --cmd 'curl -fsS -X POST https://hooks.example.com/notify -d "$AIMX_SUBJECT"' \
+  --cmd '["/bin/sh", "-c", "curl -fsS -X POST https://hooks.example.com/notify -d \"$AIMX_SUBJECT\""]' \
   --name support_notify
 
 # after_send: log successful deliveries
 sudo aimx hooks create \
   --mailbox alice \
   --event after_send \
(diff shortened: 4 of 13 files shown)


uzyn (commented):
# Sprint Review

Track: `user-root-rework` Sprint 2 — Hook fire path + mailbox isolation.

## Sprint Goal Assessment

The goal is largely achieved on observable behavior: `run_and_log` now resolves `owner` directly from `mailbox.owner` (no fallback chain, no template path), `finalize_setup` and `MAILBOX-CREATE` both lock the per-mailbox `inbox/<name>/` and `sent/<name>/` to `<owner>:<owner> 0700`, the catchall fire path has defense-in-depth early returns on both `execute_on_receive` and `execute_after_send`, and the new isolation integration test exercises the kernel-level boundary with three real probes (`cat`, `ls`, `test -r`) plus a positive-control bob-self-read.

The one substantive gap is the schema shape: `Hook.cmd` is still a `String` wrapped in `["/bin/sh", "-c", cmd]` rather than the argv array (`Vec<String>`) the PRD §6.2 FR-5 and Sprint 2 S2-1 acceptance criterion #2 explicitly call for. The implementer flagged this and adjudication is requested. See Acceptance Criteria below.

**S2-1 Rewrite `src/hook.rs` fire path**

- [x] `fire_hook` resolves `run_as` from `mailbox.owner` — met. `src/hook.rs:434` is now `let owner: String = mailbox_config.owner.clone();` with no fallback chain.
- [ ] Argv passed to `spawn_sandboxed` is `hook.cmd.clone()` — **NOT MET**. `Hook.cmd` (`src/hook.rs:90`) is still `pub cmd: String`, and `Hook::resolve_argv()` (`src/hook.rs:105–114`) wraps it in `["/bin/sh", "-c", self.cmd.clone()]`. The argv reaching `spawn_sandboxed` is therefore `["/bin/sh", "-c", <cmd>]`, not `hook.cmd.clone()`. The literal acceptance criterion fails. Adjudication below.
- [x] Template-resolution code deleted — met. No `match hook.template`, no `effective_hook_name` template branch, no `substitute_argv` calls remain.
- [x] Structured log line shape matches PRD §7.3 — met. The line at `src/hook.rs:500–506` emits `hook_name=… event=… mailbox=… owner=… sandbox=… <id_tag> exit_code=… duration_ms=… timed_out=… stderr_tail=…` in the PRD-required order. The `sandbox=` and `<id_tag>` (`email_id=…` / `message_id=…`) interpolations are pre-existing diagnostic fields between `owner=` and `exit_code=`; the PRD specifies the shape and required fields and does not forbid additional ones.
- [x] Catchall defense-in-depth early return — met. Both `execute_on_receive` (`src/hook.rs:262–271`) and `execute_after_send` (`src/hook.rs:331–340`) have the guard. Both call `mailbox_config.is_catchall(config) && !mailbox_config.hooks.is_empty()` and emit a structured `tracing::warn!` before returning. Verified by two new unit tests (`execute_on_receive_refuses_to_fire_for_catchall_mailbox`, `execute_after_send_refuses_to_fire_for_catchall_mailbox`).
- [-] Hook fire integration test with `cmd = ["/bin/echo", "hello"]` — PARTIAL. `fire_path_runs_as_mailbox_owner_and_logs_owner_field` (`src/hook.rs:1222`) tests the observable invariant — exit 0, output "hello", structured log line contains `owner=<current user>` and `exit_code=0`. The test uses `cmd = "/bin/echo hello > <out>"` (string form, shell-redirect), not the argv-array form the AC literally calls for. The test as written matches the current schema but not the AC text.
- [x] cargo test / clippy / fmt clean — met (verified locally on this branch).

**S2-2 Mailbox dir perm migration in `aimx setup`**

- [x] `setup.rs` chown/chmod logic for every configured mailbox at `<owner>:<owner> 0700` — met. `ensure_mailbox_dirs` (`src/setup.rs:1545–1602`) walks the loaded config's mailboxes and chowns/chmods both `inbox/<name>/` and `sent/<name>/`. Catchall is included.
- [x] `mailbox_handler.rs` `MAILBOX-CREATE` chown — met. `src/mailbox_handler.rs:261–273` calls `chown_as_owner(&inbox, &mb_cfg, 0o700)` and `chown_as_owner(&sent, …, 0o700)` immediately after `create_dir_all`, with rollback on failure.
- [x] `getpwnam(owner)` + `getgrnam(owner)` resolution — met. `mb.owner_uid()` and `mb.owner_gid()` route through `validate_run_as` + `resolve_user`, which is the same code path used for the daemon's `OrphanUser` error. Strict-existence is gated on `is_root()` so non-root `cargo test` runs don't trip it.
- [x] Idempotent re-run — met. `finalize_locks_mailbox_dirs_to_mode_0700` (`src/setup.rs:4226–4252`) runs `finalize_setup` twice and verifies mode `0o700` is preserved.
- [x] cargo test / clippy / fmt clean — met.

Note: The Sprint 2 plan's S2-2 AC #4 ("existing `aimx-hook`-based group setup is deleted") is explicitly deferred to Sprint 5 in the AC text itself ("removed in S5-1 anyway"), so its absence here is consistent with scope.

**S2-3 Mailbox isolation integration test**

- [x] `/bin/cat <bob>/test.md` as alice fails with EACCES — met (`tests/mailbox_isolation.rs:220–228`).
- [x] `/bin/ls <bob>/` as alice fails with EACCES — met (`tests/mailbox_isolation.rs:231–239`).
- [x] `/bin/test -r <bob>/` as alice fails — met (`tests/mailbox_isolation.rs:244–248`).
- [x] CI workflow `mailbox-isolation` job — met. `.github/workflows/ci.yml:50–87` provisions `aimx-test-alice` / `aimx-test-bob` via `useradd`, builds the test binary with `cargo test --no-run --test mailbox_isolation`, then runs `sudo AIMX_INTEGRATION_SUDO=1 "$BIN" --ignored`. Teardown is `if: always()`. The env-var gate (`AIMX_INTEGRATION_SUDO=1`) is set on the sudo invocation so the test body's gate at `tests/mailbox_isolation.rs:170–173` proceeds rather than skipping.
- [x] `#[ignore]` annotation present — met (line 169).
- [x] clippy / fmt clean — met.

**Implementer Notes Adjudicated**

1. **`cmd` schema gap.** The PRD §6.2 FR-5 and Sprint 2 S2-1 AC #2 are unambiguous: `cmd` is an argv array, no shell wrapping. The current code keeps `cmd: String` plus `/bin/sh -c` wrapping (Sprint 1 leftover). The observable invariants (owner-uid execution, exit code, structured log line) match, and the legacy template/run_as machinery is gone, but the wire format and PRD/AC text disagree. **Adjudication: this is a Non-blocker for Sprint 2 merge, but a tracked gap that MUST land before Sprint 3 / 4 / MCP rewire** — the MCP `hook_create(cmd, …)` signature in PRD §6.5 FR-15 takes an argv array, and that surface needs to land on top of the actual schema, not a translation layer. Filing as a Non-blocker rather than a Blocker because the Sprint 2 stated goal (fire path runs as mailbox owner uid; mailbox isolation; defense-in-depth catchall guard) is satisfied on observable behavior, and the schema rewrite is a separable, scoped piece of work. The Sprint plan's S2-1 AC #2 wording predates the discovery that Sprint 1 didn't convert the schema; that's a sprint-plan / Sprint-1-scope mismatch, not a Sprint 2 implementation defect. Recommend rolling the conversion into a Sprint 3 prerequisite story or a one-day cleanup before Sprint 3 starts.
2. **`catchall_mailbox` → `regular_mailbox` rename.** Verified clean. The renamed helper now constructs `address = "alice@test.com"` (was `*@test.com`), so the new defense-in-depth catchall guard does not interfere with the seven existing fire-path tests that reused the helper. Coverage is not lost — the catchall-config invariant is now exercised by two dedicated tests (`execute_on_receive_refuses_to_fire_for_catchall_mailbox`, `execute_after_send_refuses_to_fire_for_catchall_mailbox`) that explicitly construct a `MailboxConfig` with `address = "*@test.com"` and assert the guard fires.
3. **`tests/isolation.rs` co-existing with `tests/mailbox_isolation.rs`.** Both pull weight; not redundant. `tests/isolation.rs::alice_reads_own_mailbox_bob_gets_eacces` exercises the **inbound-ingest** angle: it shells out to `aimx ingest` with a real `.eml`, lets the daemon write the `.md` (chowned to alice via the ingest-side ownership helper), and proves bob cannot read the resulting file. `tests/mailbox_isolation.rs::alice_hook_cannot_read_bob_inbox` exercises the **hook-fire** angle by directly laying out per-mailbox dirs and probing with `runuser -u alice -- /bin/cat …`. They cover different code paths (ingest's `chown_as_owner` post-write versus the per-mailbox dir-perm setup) and would not catch each other's regressions.
4. **CI `mailbox-isolation` job.** Verified. The env-var gate is correctly enforced: `AIMX_INTEGRATION_SUDO=1` is prefixed onto the `sudo` invocation so the env var actually propagates into the test process (sudo's default `env_reset` is overridden by the `VAR=val cmd` syntax). The test then assertion-checks `geteuid() == 0` after the gate. The job uses `cargo test --no-run --test mailbox_isolation` to pre-build the binary as the runner user, then `sudo` re-executes it — the comment explaining why (avoids cargo home thrash under sudo) is correct. Teardown is `if: always()` so failures don't leak users. Minor non-blocker: the job name `mailbox-isolation` overlaps semantically with the existing `integration-isolation` job; consider renaming one for clarity, though both are independent and the duplication is low-cost.
5. **`finalize_setup` chown/chmod logic.** Verified. The daemon writes files as root, and root bypasses dir perms regardless of mode. With `inbox/<name>/` at `<owner>:<owner> 0700`, root (the daemon) can still create files inside it, and the post-write `chown_as_owner` in ingest gives the new file `<owner>:<owner> 0600`. The owner reads the file via uid match; other users are blocked by both the dir mode (no traverse) and the file mode (no read). PRD §7.1 invariant holds.
6. **Ignored tests (23+2 from Sprint 1).** Verified. Counted 23 `exercises legacy template/hook schema; reworked in a later sprint` + 2 primer-content (`primer_contains_creating_hooks_section`, `hooks_reference_file_bundled_and_comprehensive`) = 25 ignored tests scoped to Sprint 3 (UDS rewire) and Sprint 4 (MCP). None of them intersect Sprint 2 deliverables, so the implementer is correct that no `#[ignore]` annotation should have come off in this PR.
7. **PR title.** `Lock mailbox dir perms to <owner>:<owner> 0700 + add isolation test` — verified. Descriptive, no banned tokens.

**Banned-token sweep across repo / PR title / body / commit messages.**
`rg -n '\bSprint\b|\bS[0-9]+[-.][0-9]+\b|\bFR-[0-9]+[a-z]?\b|User Story|Acceptance criteria' --glob '!docs/**' --glob '!target/**' --glob '!.git/**'` returns zero hits. PR title, PR body, and commit message verified clean.

**Build / test / lint sweep on this branch.**
- `cargo fmt -- --check` — clean (root + verifier).
- `cargo clippy --all-targets -- -D warnings` — clean (root + verifier).
- `cargo test` (root) — `873 passed; 0 failed; 8 ignored` (lib) + `78 passed; 0 failed; 28 ignored` (integration) + smaller targets, all clean.
- `cargo test` (verifier) — `43 passed; 0 failed; 0 ignored`.

## Test Coverage

The new test additions hit the right invariants:

- `execute_on_receive_refuses_to_fire_for_catchall_mailbox` — explicit catchall-config bypass test for the on_receive guard, asserts the marker file is absent and the warning logs.
- `execute_after_send_refuses_to_fire_for_catchall_mailbox` — same shape for the after_send guard.
- `fire_path_runs_as_mailbox_owner_and_logs_owner_field` — pins the `owner=<u>` log invariant against `mailbox.owner` (the literal value the rewrite reads).
- `finalize_locks_mailbox_dirs_to_mode_0700` — pins the dir-mode invariant and the idempotency of re-running `finalize_setup`.
- `tests/mailbox_isolation.rs::alice_hook_cannot_read_bob_inbox` — kernel-level isolation across three probe shapes plus a positive-control bob-self-read sanity check.

One coverage observation:

- The `ensure_mailbox_dirs` non-root branch (chmod-only, no chown) is exercised by every existing `finalize_setup`-using test (which all run as a non-root cargo user). It is not unit-tested for behavioral correctness — there's no test that asserts "non-root finalize_setup of a config whose owner doesn't exist does NOT error." This is a small gap but probably acceptable; the mode is "test-mode permissiveness" by design and the production path (root) is the one with the strict check.

## Potential Bugs

None identified that block merge. The schema gap (Hook.cmd as String) is documented as a tracked Sprint 2 / Sprint 1 boundary issue, not a bug in the Sprint 2 implementation. The catchall guard is correctly placed before the trust check. The MAILBOX-CREATE rollback path correctly reverts both inbox and sent dirs on chown failure.

Minor observations (non-blocker):

- `src/mailbox.rs:91–96` (CLI fallback chown failure path): on chown failure, both `inbox` and `sent` are removed, but the in-memory `new_mb` is dropped and the function returns `Err` before `config.save()`. That's correct. However, on the first chown succeeding and the second failing, only the second dir's chown error is reported, even though both dirs are cleaned up. The error message is still actionable (it names the failing path), so this is purely cosmetic.
- `tests/mailbox_isolation.rs:165` calls `let _ = std::fs::remove_dir_all(&root);` before fixture setup. If a previous run left `root` with mode `0o700` owned by alice or bob, root can still traverse and remove (root bypasses dir perms), so the cleanup is correct. No bug.

## Security Issues

None introduced. The dir-perm tightening (`<owner>:<owner> 0700`) is a positive security change that closes the previous `aimx-hook` shared-readable hole. The catchall defense-in-depth guard prevents an in-memory bypass of the load-time validator from spawning subprocesses on a mailbox whose owner (`aimx-catchall`) has no resolvable login uid.

## Code Quality

- The new `ensure_mailbox_dirs` helper is well-scoped and well-documented; the doc comment correctly enumerates the root vs. non-root behavior split and the rationale.
- The catchall-guard log message is consistent across both fire paths ("`catchall hooks are forbidden; refusing to fire`"), which makes log-grep based runbooks tractable.
- The `regular_mailbox` rename is principled: the previous helper name was actively misleading (it built a non-catchall) and the new name + doc-comment are clear about why the explicit local part matters.
- The test fixture comment in `fire_path_runs_as_mailbox_owner_and_logs_owner_field` explicitly calls out the schema gap (line 1229–1233), which is exactly what an honest reviewer would want to see.

## Alignment with PRD

- PRD §6.2 FR-7 (hooks always exec as `mailbox.owner_uid()`, no per-hook `run_as`) — aligned.
- PRD §6.3 FR-3 (hooks on catchall forbidden) — aligned at fire-path level (defense-in-depth on top of load-time rejection).
- PRD §7.1 (mailbox isolation via `<owner>:<owner> 0700`) — aligned, verified by integration test.
- PRD §7.3 (structured log line shape, drops `template=…`, gains `owner=…`) — aligned. The line preserves pre-existing `sandbox=` and `email_id=`/`message_id=` diagnostic fields that the PRD doesn't forbid.
- PRD §6.2 FR-5 (cmd is an argv array, no shell wrapping) — **NOT aligned**, see Sprint 1 schema-gap discussion above. Tracked as a Non-blocker that must land before Sprint 3.

**Overall verdict: Needs minor fixes** (schema gap is non-blocking for Sprint 2 but tracked).

**Blockers:** none.

**Non-blockers:**

1. **`Hook.cmd` schema is still `String` (not argv `Vec<String>`).** Sprint 1 leftover that bled into Sprint 2's S2-1 AC #2 (which assumes argv-shaped `cmd`). The Sprint 2 observable invariants are satisfied via the `/bin/sh -c` wrapping, but PRD §6.2 FR-5 is explicit that `cmd` is an argv array and no shell interpretation. The MCP `hook_create(cmd, …)` signature defined in PRD §6.5 FR-15 takes an argv array, and Sprint 4's MCP rewire will need to land on top of the actual schema rather than a translation layer. **Action:** convert `Hook.cmd` to `Vec<String>`, update `resolve_argv` to return `hook.cmd.clone()`, update the test fixture in `fire_path_runs_as_mailbox_owner_and_logs_owner_field` to use the argv form, and refresh `book/configuration.md` / `book/hooks.md` schema docs. Either fold into a Sprint 2 follow-up commit on this branch, or open a tracked story in the Non-blocking Review Backlog and land before Sprint 3 starts.

- Consider renaming the new CI job (`mailbox-isolation`) to disambiguate from the existing `integration-isolation` job — both are fine on their own, but the names are very close in semantics. Strict non-blocker.
- The `ensure_mailbox_dirs` non-root branch is currently only covered transitively by other tests. A direct unit test on its non-root permissive behavior would tighten the contract, but is genuinely optional.

The Sprint 2 work as merged would correctly enforce mailbox isolation at the kernel level, lock dir perms to `<owner>:<owner> 0700`, and harden the fire path against in-memory catchall bypasses. The schema gap is the one substantive Sprint plan vs. implementation mismatch, and the implementer is right to surface it for adjudication.

uzyn:
## Cycle-2 fixes pushed

Addressing the cycle-1 review feedback.

### Non-blocker (now landed): `Hook.cmd` argv conversion

`Hook.cmd` is now `Vec<String>`. The fire path resolves it to `argv` verbatim via `Hook::resolve_argv` and hands it to `spawn_sandboxed` directly — no more `/bin/sh -c` wrapping. Operators who need shell expansion spell out `cmd = ["/bin/sh", "-c", "..."]` explicitly, which is the uncommon case in practice.

Validation tightened at config load (`validate_hooks` + `validate_single_hook`) and at fire time (`Hook::resolve_argv`):
- empty `cmd` array rejected with an actionable message
- blank `cmd[0]` rejected
- non-absolute `cmd[0]` rejected (hooks fire from `/var/lib/aimx/...`, PATH lookup is brittle)

CLI surface: `--cmd` now takes a JSON array string. `aimx hooks create --cmd '["/bin/echo", "hello"]' --mailbox alice --event on_receive`. Help text updated accordingly. Display in `aimx hooks list` and `aimx mailboxes show <name>` renders the argv as a JSON array string.

`derive_hook_name` now hashes the argv elements joined by 0x1F, so `["/bin/echo", "a b"]` hashes distinctly from `["/bin/echo a", "b"]` (regression test added).

UDS `HOOK-CREATE` / `HOOK-DELETE` were already stubs returning `ERR PROTOCOL` — left unchanged; they will be re-wired in a later iteration.

In-tree fixtures migrated:
- `src/hook.rs` test fixtures use a small `shell_argv()` helper so the shell-redirect semantics are preserved (the wrapping is now explicit, not implicit).
- `src/config.rs` active load-acceptance + load-rejection tests updated to argv form. Two new tests cover the new validators (`load_rejects_empty_cmd_array`, `load_rejects_non_absolute_cmd_program`).
- `src/serve.rs` reload_config_keeps_old_on_validation_failure fixture switched to `cmd = []` to keep triggering the "empty cmd" path.
- `tests/integration.rs` active CLI tests (`hooks_create_anonymous_prints_derived_name`, `hooks_create_rejects_unknown_event_at_parse_time`, `hooks_delete_prompts_and_removes_via_direct_edit`, `hooks_raw_cmd_sighup_hot_swaps_config`, `hooks_create_anonymous_prints_derived_name_via_daemon`, `mailboxes_show_prints_trust_senders_hooks_and_counts`, `after_send_hook_fires_with_delivered_status`) all use argv form. The derived-name expectation in the daemon test was updated to mirror the new 0x1F-joined hash.
- The 23+2 `#[ignore]` legacy tests were left as-is per the implementer's earlier note that they're a later-iteration concern.

Book + agents docs:
- `book/configuration.md`: hook intro paragraph rewritten, table updated, three example snippets converted.
- `book/hooks.md`: raw-cmd intro rewritten, CLI examples switched to JSON-array `--cmd`, hook-properties table updated, fire-flow narrative updated, audit/notify examples converted.
- `book/hook-recipes.md`: all eleven recipe snippets converted to `cmd = ["/bin/sh", "-c", '''...''']` form (preserving the shell expansion that those recipes depend on).
- `agents/common/aimx-primer.md` and `agents/common/references/hooks.md` already use argv form for the template-related examples; left unchanged.

### Nice-to-have 1: CI job rename

`.github/workflows/ci.yml`: `mailbox-isolation` -> `mailbox-dir-perms-isolation` to disambiguate from the pre-existing `integration-isolation` job.

### Nice-to-have 2: direct unit test for `ensure_mailbox_dirs` non-root branch

`src/setup.rs::ensure_mailbox_dirs_non_root_skips_chown_and_owner_existence`. Constructs a fake `Config` with a synthetic `owner` that almost certainly does not resolve on the host, calls `ensure_mailbox_dirs` as a non-root caller, and asserts: returns Ok, both `inbox/<name>/` and `sent/<name>/` exist with mode `0o700`, idempotent re-run still Ok.

### Verification

- `cargo test` (root crate): 879 passed; 0 failed; 8 ignored (lib) + 78 passed; 0 failed; 28 ignored (integration) + smaller targets all clean.
- `cargo test` (services/verifier): 43 passed; 0 failed; 0 ignored.
- `cargo clippy --all-targets -- -D warnings`: clean on both crates.
- `cargo fmt -- --check`: clean on both crates.
- Banned-token sweep over the working tree (excluding `docs/`, `target/`, `.git/`): zero hits.
- mdbook build: clean.

Net diff: +415 / -161 across 12 files.

uzyn (commented):
# Sprint Re-review (cycle 2)

Track: `user-root-rework` Sprint 2 — Hook fire path + mailbox isolation, post-cycle-1 fixes.

## Cycle-1 outstanding items: status

### Promoted blocker: `Hook.cmd: String` → argv `Vec<String>` — RESOLVED

`src/hook.rs` now declares `pub cmd: Vec<String>`. `Hook::resolve_argv` returns `self.cmd.clone()` verbatim — no more `["/bin/sh", "-c", …]` wrapping. The fire path hands the raw argv to `spawn_sandboxed`, so PRD §6.2 FR-5 (cmd is an argv array, no shell interpretation) is now satisfied at the wire format, not just the observable behaviour.

Validation tightened in two places that share the same rule set:

- `Hook::resolve_argv` (`src/hook.rs:107–124`): rejects empty array, blank `cmd[0]`, non-absolute `cmd[0]`. New `ResolveArgvError::NonAbsoluteProgram(String)` variant.
- `validate_hooks` / `validate_single_hook` (`src/config.rs:613–706`): same three checks at config load with actionable error messages naming the offending mailbox + hook label and pointing to the absolute-path requirement.

Two new load-time tests:

- `load_rejects_empty_cmd_array` — feeds `cmd = []`, asserts the error mentions "empty `cmd`".
- `load_rejects_non_absolute_cmd_program` — feeds `cmd = ["echo", "hi"]`, asserts the error calls out non-absolute `cmd[0]`.

Three new `resolve_argv` unit tests:

- `resolve_argv_returns_cmd_verbatim` (replaces the old `wraps_in_sh_dash_c`)
- `resolve_argv_empty_cmd_fails`
- `resolve_argv_blank_program_fails`
- `resolve_argv_non_absolute_program_fails`

`derive_hook_name` now hashes argv elements joined by 0x1F so split boundaries matter:

```
sha256(event || 0x1F || arg0 || 0x1F || arg1 || … || 0x1F || fire_on_untrusted_byte)
```

The new regression test `derive_hook_name_differs_by_argv_split` verifies `["/bin/echo", "a b"]` and `["/bin/echo a", "b"]` produce distinct names. Three-way `["echo", "", "hi"]` vs. `["echo hi"]` vs. `["echo", "hi"]` is not exhaustively pinned, but the algorithm is structurally correct (each argv element is followed by a 0x1F, so a `""` element shows up as `"\x1F\x1F"` in the hash stream — distinct from any single-element form).

CLI surface (`src/cli.rs:417–424`, `src/hooks.rs:230–254`):

- `--cmd` takes a JSON array string. Help text updated to call this out explicitly with an example.
- `parse_cmd_argv` short-circuits with a clear error when the value doesn't start with `[`, and surfaces the underlying `serde_json` parse error on malformed JSON. Both error paths emit the canonical `--cmd '["/bin/echo", "hello"]'` example.
- Empty array rejected at parse time before hitting the validator.

Display surfaces (`src/hooks.rs:69–73`, `src/mailbox.rs:440–441`): `format_argv_for_display` renders the argv via `serde_json::to_string` and falls back to space-join. Output is `["/bin/echo","inbound"]` shape, which is the same form `--cmd` accepts (round-trippable copy-paste).

Round-trip serde verified by `raw_cmd_hook_omits_default_fields_on_serialize` (Hook → TOML) plus `hook_toml_with_only_required_fields_loads` (TOML `cmd = ["/bin/echo", "hi"]` → Hook). Reload-keeps-old-on-validation-failure fixture in `src/serve.rs` switched to `cmd = []` to keep triggering the empty-cmd path.

Active integration tests (`tests/integration.rs:4233`, `4313`, `4340`, `4411`, `4470`) and the fixture in `mailboxes_show_prints_trust_senders_hooks_and_counts` (`4070–4078`) now use argv form. The daemon-side derived-name expectation in `hooks_create_anonymous_prints_derived_name_via_daemon` was correctly updated to feed the same 0x1F-joined argv into the SHA-256 stream.

The `shell_argv()` helper in `src/hook.rs:854–859` keeps the in-tree shell-redirect-based hook tests honest: each test that needs `$AIMX_FILEPATH` expansion now calls `shell_argv(format!(…))` so the `["/bin/sh", "-c", "…"]` wrapping is **explicit** in the test body. Nothing silently relies on implicit wrapping.

The 23+2 `#[ignore]`'d legacy tests retain string-form `cmd = "..."` fixtures. I confirmed every active call site of the legacy `setup_test_env_with_triggers` helper is itself `#[ignore]`'d, so no live test path consumes the stale fixture.

### Nice-to-have 1: CI job rename — RESOLVED

`.github/workflows/ci.yml:50` renamed `mailbox-isolation` → `mailbox-dir-perms-isolation`. The disambiguation is real (the pre-existing `integration-isolation` job covers a different ingest-side scenario). Cycle-2 CI run [24961194293](https://github.com/uzyn/aimx/actions/runs/24961194293) shows the renamed job succeeding in 105s; provisioning, build, sudo execution, and teardown all green.

### Nice-to-have 2: `ensure_mailbox_dirs` non-root unit test — RESOLVED

`src/setup.rs:4259–4330` adds `ensure_mailbox_dirs_non_root_skips_chown_and_owner_existence`. Constructs a `Config` whose mailbox `owner = "aimx-nonexistent-test-user"`, calls `ensure_mailbox_dirs` from a non-root harness, and asserts:

- returns Ok (the strict owner-existence check is `is_root()`-gated and stays out of the way)
- both `inbox/<name>/` and `sent/<name>/` are created with mode `0o700`
- idempotent re-run still Ok

The test self-skips with a clear stderr message if the harness somehow runs as root (e.g. on a CI box where `cargo test` is invoked under sudo) so it doesn't trip the production-only check. That guard is sensible and won't false-positive on contributor machines.

## Verification of additional items called out by the orchestrator

1. **Argv-shape on the wire.** Verified. `serde` `Vec<String>` round-trips trivially; the load test `hook_toml_with_only_required_fields_loads` exercises one direction and `raw_cmd_hook_omits_default_fields_on_serialize` exercises the other. `MAILBOX-CREATE` and the `aimx hooks create` direct-edit path both write through `Config` serialisation.
2. **Empty / non-absolute rejection.** Verified, both at `Config::load` and `Hook::resolve_argv`. Error messages name the offending field, the example shape, and the rationale ("hooks fire from /var/lib/aimx/... so PATH lookup is brittle").
3. **CLI flag UX.** Help text on `--cmd` (`src/cli.rs:417–423`) is clear. Error path on non-JSON input is clear: "--cmd must be a JSON array of strings, e.g. --cmd '["/bin/echo", "hello"]'; got: <raw>". `book/hooks.md` and `book/configuration.md` reflect the JSON-array form. `book/hook-recipes.md` converts all eleven recipe snippets to explicit `["/bin/sh", "-c", '''...''']` form, preserving the shell expansion the recipes need.
4. **Display path round-trippability.** Verified. `format_argv_for_display` renders `serde_json::to_string(argv)` so `["/bin/echo","inbound"]` is what an operator sees — same shape `--cmd` accepts.
5. **`derive_hook_name` 0x1F-join.** Verified. `["/bin/echo", "a b"]` vs. `["/bin/echo a", "b"]` regression test pins the split-boundary invariant. Algorithm is structurally collision-free for arbitrary empty/non-empty splits.
6. **Test fixture coverage.** `rg -n 'cmd:\s*"' src/ tests/` reports zero hits in active code. `rg -n 'cmd\s*=\s*"' src/ tests/` reports the 8 expected hits in `#[ignore]`'d legacy fixtures plus 4 in `src/config.rs::tests::load_rejects_*` (where `cmd = "true"` is intentionally stale — the `reject_legacy_schema` pre-parse layer hits the legacy field _before_ TOML parsing reaches `cmd`, so those test bodies still pass and exercise the legacy-rejection invariant they're named for; pure cosmetic, not blocker).
7. **Book + agents docs.** `book/configuration.md` (table + three example snippets), `book/hooks.md` (raw-cmd intro, CLI examples, properties table, fire-flow narrative, audit/notify recipe), `book/hook-recipes.md` (eleven recipes) all converted. `agents/common/aimx-primer.md` and `agents/common/references/hooks.md` already used argv form for template examples and were correctly left unchanged. **Gap:** `book/cli.md:138–195` still describes the legacy schema (`--cmd 'shell string'`, `--template`, `--param`, `--dangerously-support-untrusted`, `origin = "operator"`/`"mcp"`, "Raw shell command executed via `sh -c`"). This file was last touched in PR #133 (an earlier track's Sprint 8) and predates the entire user-root-rework. Implementer's cycle-2 message lists configuration.md / hooks.md / hook-recipes.md but does not mention cli.md. Calling this out as a non-blocker — see Summary.
8. **Shell-wrapping migration.** Verified. Every test that previously relied on shell expansion now spells out `["/bin/sh", "-c", "..."]` via the `shell_argv()` helper. Nothing silently broken.
9. **CI timing.** Cycle-2 push run [24961194293](https://github.com/uzyn/aimx/actions/runs/24961194293) — `mailbox-dir-perms-isolation` 105s green. All four jobs (`core-tests` 206s, `mailbox-dir-perms-isolation` 105s, `docs-build` 36s, `verifier-tests` 66s) pass.
10. **Banned tokens, lint, fmt, tests on the latest commit.** Re-verified locally on `lock-mailbox-dir-perms` @ `3684a04`:
    - `rg -n '\bSprint\s*[0-9]+|\bS[0-9]+[-.][0-9]+\b|\bFR-[0-9]+[a-z]?\b|User Story|Acceptance criteria' --glob '!docs/**' --glob '!target/**' --glob '!.git/**'` — zero hits.
    - `cargo fmt -- --check` — clean (root + verifier).
    - `cargo clippy --all-targets -- -D warnings` — clean (root + verifier).
    - `cargo test` (root) — 879 passed, 8 ignored (lib) + 78 passed, 28 ignored (integration) + 0/1 + 0/1 + 0/1 + 1/1 across the smaller targets. All green.
    - `cargo test` (verifier) — 43 passed, 0 failed.
    - PR title, PR body, and commit messages — verified clean.

## New issues found

One pre-existing doc-staleness gap surfaced by the cycle-2 conversion (the cycle-2 work made the gap visible because `book/configuration.md` and `book/hooks.md` were updated correctly):

- **`book/cli.md` describes the legacy hook schema.** Lines 138–195 still document `--template` / `--param`, `--dangerously-support-untrusted`, `origin = "operator"`, "Raw shell command executed via `sh -c`", and `--cmd 'shell string'` (string form). The actual CLI surface (`src/cli.rs:HookCreateArgs`) has none of those — it has `--cmd <json-array>`, `--fire-on-untrusted`, no `--template`/`--param`. Track-level cleanup: `book/cli.md` was last touched in an earlier track's Sprint 8 (PR #133) and naturally lands in this track's Sprint 6 ("Book + final integration"). Calling out as a non-blocker rather than a blocker because (a) the file is not in the cycle-2 implementer's listed scope, (b) the staleness is pre-existing, and (c) Sprint 6 is the right home for the full book sweep. If the implementer wants to fold a one-line fix into this branch they're welcome to, but it's not required for Sprint 2 to merge.

All cycle-1 items addressed. The promoted-blocker (Hook.cmd schema conversion) is comprehensive: schema, validation, derive-hook-name, CLI parser, display path, fire-path execution, test fixtures, and book docs are all internally consistent. The two nice-to-haves landed cleanly. CI is green on the latest push.

**Blockers:** none.

**Non-blockers:**

1. `book/cli.md` still describes the legacy hook schema. Naturally folds into this track's Sprint 6 ("Book + final integration"); no need to block Sprint 2 merge.

- `src/config.rs:1145, 1171, 1197, 1248` (the `load_rejects_*` tests) keep `cmd = "true"` string-form fixtures. They pass because `reject_legacy_schema` rejects the legacy fields (`run_as` / `origin` / `dangerously_support_untrusted` / `stdin = "email_json"`) at the pre-parse layer before `cmd` is parsed, so the stale form never reaches the TOML parser. Cosmetic — converting to `cmd = ["/bin/true"]` would future-proof against a hypothetical refactor that moved validation order.

```
Lock mailbox dir perms to <owner>:<owner> 0700; convert hook.cmd to argv

- finalize_setup and MAILBOX-CREATE both create per-mailbox inbox/<name>/
  and sent/<name>/ at <owner>:<owner> mode 0700; setup verifies the
  owner exists before chown'ing.
- New tests/mailbox_isolation.rs (gated behind AIMX_INTEGRATION_SUDO=1)
  provisions two real Linux uids and asserts cat / ls / test -r as alice
  on bob's inbox all fail with EACCES. New CI job
  mailbox-dir-perms-isolation provisions the users and runs the test.
- Hook fire path defense-in-depth guard refuses to fire when a hook has
  ended up on a catchall mailbox; load-time validation already rejects
  this, the guard catches in-memory bypasses.
- Hook.cmd is now Vec<String> (argv array), exec'd directly via
  spawn_sandboxed with no implicit /bin/sh -c wrapping. Validation
  rejects empty arrays, blank cmd[0], and non-absolute cmd[0] at both
  config load and fire time. CLI --cmd takes a JSON array string;
  display surfaces render argv as JSON. derive_hook_name hashes argv
  elements joined by 0x1F. book/configuration.md, book/hooks.md, and
  book/hook-recipes.md updated.
```

uzyn:
CodeQL follow-up:

- **`actions/missing-workflow-permissions`** (medium, `.github/workflows/ci.yml:51`): Fixed by adding a top-level `permissions: contents: read` block so all jobs inherit least-privilege by default.
- **`rust/cleartext-logging`** (high, `src/hook.rs:1288`): Dismissed as false positive (alert #9). The structured hook-fire log line includes `owner=<u>` so operators can see which user a hook ran as — that field is intentional audit data, not sensitive. The flagged site is the assertion in a regression test that pins this invariant.

PR #172, opened 2026-04-26
Enforce SO_PEERCRED authorization on every UDS verb

uzyn:
## Summary

Wires per-verb `SO_PEERCRED` authorization into every UDS handler in `aimx serve` and un-stubs `HOOK-CREATE` / `HOOK-DELETE` under the new authz predicate. The wire policy is now a uniform "caller uid owns the target mailbox, OR is root" check for `SEND`, `MARK-*`, `HOOK-*`, with `MAILBOX-CREATE` / `MAILBOX-DELETE` tightened to root-only.

### What landed

- **Plumbing.** `serve.rs`'s accept loop already captured `SO_PEERCRED` via `caller_from_stream` and threaded the resolved `Caller` into every dispatch. `send_handler`, `state_handler`, and `mailbox_handler` already enforced authz via `enforce_mailbox_owner_or_root` / `enforce_root`. This change closes the remaining gap: the hook handlers are no longer stubbed.
- **`hook_handler.rs` rewrite.** The `ERR PROTOCOL hook-X over UDS is not supported in this build` stubs are gone. The new handlers:
  - Resolve the target mailbox up front and run `enforce_mailbox_owner_or_root` before any state mutation.
  - Reject hooks on the catchall mailbox (defense in depth around the existing load-time rejection).
  - Parse the body as a JSON object with `serde(deny_unknown_fields)`, so legacy `template`, `params`, `run_as`, `origin`, `dangerously_support_untrusted` payloads reject with `ERR PROTOCOL` at body parse — symmetric with `Config::load`.
  - Build a `Hook` struct, validate via `validate_single_hook` then `validate_hooks`, then write `config.toml` atomically and swap `ConfigHandle` — same lock hierarchy as `MAILBOX-CRUD` (outer per-mailbox lock, inner `CONFIG_WRITE_LOCK`).
- **Salvaged CLI integration tests.** Two of the 23 `#[ignore = \"exercises legacy template/hook schema; reworked in a later sprint\"]` tests in `tests/integration.rs` (`hooks_create_and_list_roundtrip_direct_edit`, `hooks_create_rejects_invalid_name`) un-ignored and re-fitted to the JSON-array `--cmd` shape. The rest stay ignored because they exercise removed surfaces (template MCP tools, catchall hooks, `dangerously_support_untrusted`).
- **`tests/uds_authz.rs` rewrite.** Reworked from a single bob-vs-alice probe into a per-verb × per-outcome matrix: `send_as_owner_ok`, `send_as_other_forbidden`, `send_as_root_any_ok`, the same shape for `MARK-READ`, `MARK-UNREAD`, `HOOK-CREATE`, `HOOK-DELETE`, plus `_root_ok` / `_non_root_forbidden` for `MAILBOX-CREATE` / `MAILBOX-DELETE`. Renamed test users to `aimx-test-alice` / `aimx-test-bob` so the test shares user provisioning with `mailbox_isolation`.
- **CI.** `mailbox-dir-perms-isolation` job extended with a `Run UDS authz tests (under sudo)` step that builds and runs the rewritten `uds_authz` test under root with `AIMX_INTEGRATION_SUDO=1`.

### Pre-launch breaking change

`MAILBOX-CREATE` / `MAILBOX-DELETE` over UDS now require root. Previously any local user could submit them. Pre-launch operators relying on the old behavior must invoke `sudo aimx mailboxes create | delete` (which already exists) or the daemon will reject with `ERR EACCES`.

### Wire shape — `HOOK-CREATE` body

```json
{
  \"cmd\": [\"/bin/echo\", \"hi\"],
  \"fire_on_untrusted\": false,
  \"type\": \"cmd\"
}
```

`fire_on_untrusted` and `type` are optional (defaults `false` and `\"cmd\"`). Any other key — including the removed `template`, `params`, `run_as`, `origin`, `dangerously_support_untrusted` — rejects with `ERR PROTOCOL`.

## Test plan

- [x] `cargo test` — 895 unit + 80 integration + 1 release-integration + 1 base64 helper, 0 failed.
- [x] `cargo clippy --all-targets -- -D warnings` clean on `aimx` root crate.
- [x] `cargo fmt -- --check` clean on `aimx` root crate.
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean on `services/verifier`.
- [x] `tests/uds_authz.rs` — 19 ignored tests cover the SEND × {owner / other / root} + MARK × {2 verbs × 3 outcomes} + HOOK × {2 verbs × 3 outcomes} + MAILBOX × {2 verbs × 2 outcomes} matrix; gated behind `AIMX_INTEGRATION_SUDO=1`.
- [ ] Verify the new CI step `Run UDS authz tests (under sudo)` runs green when this PR builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,8 +68,10 @@ jobs:
           sudo useradd --system --no-create-home --shell /usr/sbin/nologin aimx-test-bob || \
             id aimx-test-bob
 
-      - name: Build the isolation test binary
-        run: cargo test --no-run --test mailbox_isolation
+      - name: Build the isolation test binaries
+        run: |
+          cargo test --no-run --test mailbox_isolation
+          cargo test --no-run --test uds_authz
 
       - name: Run mailbox isolation test (under sudo)
         run: |
@@ -82,6 +84,17 @@ jobs:
           BIN=$(ls -t target/debug/deps/mailbox_isolation-* | grep -v '\.d$' | head -n 1)
           test -x "$BIN"
           sudo AIMX_INTEGRATION_SUDO=1 "$BIN" --ignored
+
+      - name: Run UDS authz tests (under sudo)
+        run: |
+          # Same shape as the mailbox-isolation step. The uds_authz
+          # tests spin up a real `aimx serve` subprocess under a
+          # tempdir, then poke each verb as alice / bob / root over
+          # the unix socket. Authz rejections must surface as
+          # `EACCES` on the wire; root must always pass.
+          BIN=$(ls -t target/debug/deps/uds_authz-* | grep -v '\.d$' | head -n 1)
+          test -x "$BIN"
+          sudo AIMX_INTEGRATION_SUDO=1 "$BIN" --ignored --test-threads=1
 
       - name: Tear down test users
         if: always()
--- a/src/hook_handler.rs
+++ b/src/hook_handler.rs
@@ -1,41 +1,743 @@
 //! Daemon-side handlers for the `HOOK-CREATE` and `HOOK-DELETE` verbs of
 //! the `AIMX/1` UDS protocol.
 //!
-//! The UDS hook verbs are stubs that return `ERR PROTOCOL` until the
-//! UDS hardening pass rewires them under the new auth predicate.
-//! Operators continue to manage hooks via `sudo aimx hooks create
-//! --cmd` / `sudo aimx hooks delete`, both of which write
-//! `config.toml` directly and SIGHUP the daemon.
+//! Authorization: caller uid must match the target mailbox's owner uid,
+//! or be root. The check runs **before** any state mutation so a
+//! non-owner never observes a partial write.
+//!
+//! On success the handler rewrites `config.toml` atomically (write-temp-
+//! then-rename) and swaps the live `Config` snapshot through
+//! [`ConfigHandle::store`]. The same lock hierarchy as `MAILBOX-CRUD`
+//! applies: outer per-mailbox lock, inner process-wide
+//! [`mailbox_handler::CONFIG_WRITE_LOCK`].
+//!
+//! Body parser (`HOOK-CREATE`): JSON object with `cmd: [String]` plus
+//! optional `fire_on_untrusted: bool` and `type: String` (default
+//! `"cmd"`). The legacy `template`, `params`, `run_as`, `origin`,
+//! `dangerously_support_untrusted` fields are rejected with
+//! `ERR PROTOCOL` for symmetry with `Config::load`.
 
-use crate::mailbox_handler::MailboxContext;
+use serde::Deserialize;
+
+use crate::config::{Config, validate_hooks, validate_single_hook};
+use crate::hook::{Hook, HookEvent, effective_hook_name, is_valid_hook_name};
+use crate::mailbox_handler::{CONFIG_WRITE_LOCK, MailboxContext, write_config_atomic};
 use crate::send_protocol::{AckResponse, ErrCode, HookCreateRequest, HookDeleteRequest};
 use crate::state_handler::StateContext;
-use crate::uds_authz::Caller;
+use crate::uds_authz::{Caller, enforce_mailbox_owner_or_root};
+
+/// Wire-shape of the `HOOK-CREATE` JSON body. Mirrors the trimmed
+/// `Hook` schema in `src/hook.rs` minus the `event` and `name` fields,
+/// which travel as request headers. `deny_unknown_fields` causes a
+/// legacy `template` / `params` / `run_as` / `origin` /
+/// `dangerously_support_untrusted` to reject at body-parse time, before
+/// the handler ever touches `config.toml`.
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct HookCreateBody {
+    cmd: Vec<String>,
+    #[serde(default)]
+    fire_on_untrusted: bool,
+    #[serde(default = "default_hook_type")]
+    r#type: String,
+}
+
+fn default_hook_type() -> String {
+    "cmd".to_string()
+}
 
 pub async fn handle_hook_create(
-    _state_ctx: &StateContext,
-    _mb_ctx: &MailboxContext,
-    _req: &HookCreateRequest,
-    _caller: &Caller,
+    state_ctx: &StateContext,
+    mb_ctx: &MailboxContext,
+    req: &HookCreateRequest,
+    caller: &Caller,
 ) -> AckResponse {
-    AckResponse::Err {
-        code: ErrCode::Protocol,
-        reason: "HOOK-CREATE over UDS is not supported in this build; \
-                 use `sudo aimx hooks create --cmd ...`"
-            .to_string(),
+    // Resolve the target mailbox up front so the authz check runs
+    // against the same snapshot the rest of the handler will build on.
+    let snapshot = mb_ctx.config_handle.load();
+    let mailbox_cfg = match snapshot.mailboxes.get(&req.mailbox) {
+        Some(m) => m.clone(),
+        None => {
+            return AckResponse::Err {
+                code: ErrCode::Enoent,
+                reason: format!("mailbox '{}' does not exist", req.mailbox),
+            };
+        }
+    };
+
+    if let Err(reject) =
+        enforce_mailbox_owner_or_root("HOOK-CREATE", caller, &req.mailbox, &mailbox_cfg)
+    {
+        return AckResponse::Err {
+            code: reject.code,
+            reason: reject.reason,
+        };
     }
+
+    // Hooks on the catchall are forbidden at config-load time; mirror
+    // the rejection here so a stale UDS client (e.g. one that
+    // bypassed the load-time validator by submitting against a freshly
+    // mutated config) cannot land one through the side door.
+    if mailbox_cfg.is_catchall(&snapshot) {
+        return AckResponse::Err {
+            code: ErrCode::Mailbox,
+            reason: format!(
+                "mailbox '{}' is the catchall; hooks on the catchall are forbidden",
+                req.mailbox
+            ),
+        };
+    }
+
+    let event = match parse_event(&req.event) {
+        Ok(e) => e,
+        Err(reason) => {
+            return AckResponse::Err {
+                code: ErrCode::Validation,
+                reason,
+            };
+        }
+    };
+
+    if let Some(name) = &req.name
+        && !is_valid_hook_name(name)
+    {
+        return AckResponse::Err {
+            code: ErrCode::Validation,
+            reason: format!(
+                "invalid hook name '{name}': must match \
+                 [a-zA-Z0-9_][a-zA-Z0-9_.-]{{0,127}}"
+            ),
+        };
+    }
+
+    // Body shape: JSON object. `serde(deny_unknown_fields)` rejects
+    // legacy template/params/run_as/origin/dangerously_* fields so the
+    // wire schema stays in lock-step with `Config::load`.
+    let body: HookCreateBody = match serde_json::from_slice(&req.body) {
+        Ok(b) => b,
+        Err(e) => {
+            return AckResponse::Err {
+                code: ErrCode::Protocol,
+                reason: format!("invalid HOOK-CREATE body: {e}"),
+            };
+        }
+    };
+
+    let hook = Hook {
+        name: req.name.clone(),
+        event,
+        r#type: body.r#type,
+        cmd: body.cmd,
+        fire_on_untrusted: body.fire_on_untrusted,
+    };
+
+    if let Err(reason) = validate_single_hook(&hook) {
+        return AckResponse::Err {
+            code: ErrCode::Validation,
+            reason,
+        };
+    }
+
+    // Acquire the same lock hierarchy mailbox CRUD takes: outer per-
+    // mailbox lock (shared with ingest / MARK-* / MAILBOX-CRUD), inner
+    // process-wide config write lock. Always outer → inner.
+    let lock = state_ctx.lock_for(&req.mailbox);
+    let _guard = lock.lock().await;
+    let _config_guard = CONFIG_WRITE_LOCK
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
+
+    let current = mb_ctx.config_handle.load();
+    let mut new_config: Config = (*current).clone();
+    let mb = match new_config.mailboxes.get_mut(&req.mailbox) {
+        Some(m) => m,
+        None => {
+            return AckResponse::Err {
+                code: ErrCode::Enoent,
+                reason: format!(
+                    "mailbox '{}' does not exist (race with concurrent MAILBOX-DELETE)",
+                    req.mailbox
+                ),
+            };
+        }
+    };
+
+    let effective = effective_hook_name(&hook);
+    mb.hooks.push(hook);
+
+    if let Err(reason) = validate_hooks(&new_config) {
+        return AckResponse::Err {
+            code: ErrCode::Validation,
+            reason,
+        };
+    }
+
+    if let Err(e) = write_config_atomic(&mb_ctx.config_path, &new_config) {
+        return AckResponse::Err {
+            code: ErrCode::Io,
+            reason: format!("failed to write {}: {e}", mb_ctx.config_path.display()),
+        };
+    }
+
+    mb_ctx.config_handle.store(new_config);
+    tracing::info!(
+        target: "aimx::hook",
+        verb = "HOOK-CREATE",
+        mailbox = %req.mailbox,
+        hook_name = %effective,
+        caller_uid = caller.uid,
+        "hook created"
+    );
+    AckResponse::Ok
 }
 
 pub async fn handle_hook_delete(
-    _state_ctx: &StateContext,
-    _mb_ctx: &MailboxContext,
-    _req: &HookDeleteRequest,
-    _caller: &Caller,
+    state_ctx: &StateContext,
+    mb_ctx: &MailboxContext,
+    req: &HookDeleteRequest,
+    caller: &Caller,
 ) -> AckResponse {
-    AckResponse::Err {
-        code: ErrCode::Protocol,
-        reason: "HOOK-DELETE over UDS is not supported in this build; \
-                 use `sudo aimx hooks delete <name>`"
-            .to_string(),
+    // Locate the hook by effective name across every mailbox before
+    // taking any lock. The mailbox name we resolve here drives both
+    // authz and the per-mailbox lock; without resolving up front we
+    // cannot answer "which mailbox does this hook belong to?"
+    let snapshot = mb_ctx.config_handle.load();
+    let mailbox_name = match find_hook_owner(&snapshot, &req.name) {
+        Some(name) => name,
+        None => {
+            return AckResponse::Err {
+                code: ErrCode::Enoent,
+                reason: format!("hook '{}' not found", req.name),
+            };
+        }
+    };
+
+    // Re-lookup against the same snapshot so the authz check sees a
+    // consistent view of the mailbox's owner.
+    let mailbox_cfg = match snapshot.mailboxes.get(&mailbox_name) {
+        Some(m) => m.clone(),
+        None => {
+            return AckResponse::Err {
+                code: ErrCode::Enoent,
+                reason: format!("hook '{}' not found", req.name),
+            };
+        }
+    };
+
+    if let Err(reject) =
+        enforce_mailbox_owner_or_root("HOOK-DELETE", caller, &mailbox_name, &mailbox_cfg)
+    {
+        return AckResponse::Err {
+            code: reject.code,
+            reason: reject.reason,
+        };
+    }
+
+    let lock = state_ctx.lock_for(&mailbox_name);
+    let _guard = lock.lock().await;
+    let _config_guard = CONFIG_WRITE_LOCK
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
+
+    // Re-resolve under the lock so a concurrent HOOK-CREATE / DELETE on
+    // the same mailbox doesn't slip past us.
+    let current = mb_ctx.config_handle.load();
+    let mut new_config: Config = (*current).clone();
+    let mb = match new_config.mailboxes.get_mut(&mailbox_name) {
+        Some(m) => m,
+        None => {
+            return AckResponse::Err {
+                code: ErrCode::Enoent,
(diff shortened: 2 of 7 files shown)


uzyn (commented):
## Sprint 3 Review — UDS Hardening (PR #172)

### Sprint Goal Assessment

The sprint goal — "every UDS verb authorizes via `SO_PEERCRED` + `auth::authorize` before doing anything; `MAILBOX-*` becomes root-only on the wire" — is substantively achieved in code. `hook_handler.rs` is fully un-stubbed with a clean authz-first / lock-hierarchy / atomic-write structure that mirrors `mailbox_handler.rs`. However, **CI is currently red on the very integration tests that prove the wire policy on `MAILBOX-CREATE` / `MAILBOX-DELETE`**, and the failure is a real bug in the test wire frames (not flakiness). Sprint 3 cannot be considered Done until those four tests pass.

**[S3-1] Plumb peer uid through `serve.rs` accept loop into every handler**
- [x] `serve.rs` accept loop reads `peer_cred()` immediately after accept — already in place from earlier work; `caller_from_stream` and `Caller::from_ucred` confirm coverage.
- [x] Every handler dispatch passes `peer_uid` (now `&Caller`) into the handler.
- [x] All handler signatures updated (`send_handler`, `state_handler`, `mailbox_handler`, `hook_handler`).
- [x] Diagnostic log line preserved via `log_decision`.
- [x] `cargo test`, clippy, fmt clean for the handler-plumbing code.

**[S3-2] Add `auth::authorize` calls in every UDS handler**
- [x] `send_handler` calls `enforce_mailbox_owner_or_root("SEND", …)`.
- [x] `state_handler` calls `enforce_mailbox_owner_or_root` for `MARK-READ` / `MARK-UNREAD`.
- [x] `mailbox_handler` calls `enforce_root("MAILBOX-CREATE"/"MAILBOX-DELETE", …)` — root-only enforcement is correct.
- [x] `hook_handler` un-stubbed: `enforce_mailbox_owner_or_root("HOOK-CREATE"/"HOOK-DELETE", …)`, body parses with `serde(deny_unknown_fields)`, lock hierarchy matches `mailbox_handler`, atomic `write_config_atomic` + `ConfigHandle::store`. 16 unit tests are well-targeted and pass locally.
- [PARTIAL] Body parsing rejection symmetry — `serde(deny_unknown_fields)` does reject `template`/`params`/`run_as`/`origin`/`dangerously_support_untrusted` at `ERR PROTOCOL`, but the error text does **not** point operators at `book/hooks.md` (unlike `Config::load`, which produces "fields were removed. See book/hooks.md for the supported raw-cmd schema"). The wire error is a generic serde "unknown field 'template', expected one of …". A migrated operator hitting the daemon over UDS gets a less helpful message than one editing config directly. See blocker #2.
- [x] `cargo test`, clippy, fmt clean for the handler code.

**[S3-3] Negative-auth integration tests for every UDS verb**
- [x] `tests/uds_authz.rs` exists with all 19 named permutations.
- [BLOCKER] **4 of 19 tests fail in CI** — `mailbox_create_root_ok`, `mailbox_create_non_root_forbidden`, `mailbox_delete_root_ok`, `mailbox_delete_non_root_forbidden`. The wire frames in `raw_mailbox_create_frame` / `raw_mailbox_delete_frame` use `Mailbox: <name>`, but the `MAILBOX-CREATE` / `MAILBOX-DELETE` codec (per `src/send_protocol.rs:1014`) requires a `Name: <name>` header. The CI step output (run 24962351451) shows: `expected response to contain "EACCES"; got "AIMX/1 ERR MALFORMED malformed request: missing required header: Name"`. As a result, the four `MAILBOX-*` tests never actually exercise the authz check — the parser rejects them upstream of `enforce_root`. The acceptance criterion "negative-auth integration tests for every UDS verb" is technically not met for the two MAILBOX verbs; the EACCES path on the wire is unverified.
- [x] All tests gated behind `#[ignore]`.
- [x] CI workflow extended with the `Run UDS authz tests (under sudo)` step using `--test-threads=1`.
- [x] Test usernames `aimx-test-alice` / `aimx-test-bob` reuse Sprint 2's CI provisioning — confirmed.
- [x] `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check` clean.

### Test Coverage

Solid where it lands. 16 `hook_handler::tests` exercise the right matrix: unknown mailbox → ENOENT, legacy `template` field → PROTOCOL, legacy `run_as` field → PROTOCOL, empty `cmd` → VALIDATION, non-absolute `cmd[0]` → VALIDATION, `fire_on_untrusted` on `after_send` → VALIDATION, hook on catchall → MAILBOX, non-owner → EACCES (with config left unchanged), round-trip create-then-delete, persistence to `config.toml`, duplicate-name rejection, invalid-name rejection.

Gaps I noticed:

1. **No on-the-wire test of the `book/hooks.md` reference**. If you accept the symmetry concern (blocker #2), you'll want a test that submits a body with `template = "x"` and asserts the rejection text contains `book/hooks.md`.
2. **`MAILBOX-*` authz path is unverified end-to-end**. Once the header bug is fixed, CI will actually exercise it.
3. **No `hook_create` test against catchall over the wire** — the unit test `create_rejects_hook_on_catchall` covers it in-process, but `tests/uds_authz.rs` doesn't include a wire-level test. Defense-in-depth coverage is fine; the unit test is sufficient unless we're worried about the wire path bypassing the in-process check.
4. The implementer's classification of the 21 still-`#[ignore]`'d tests is **correct**. The `mcp_hook_create_*` / `mcp_hook_delete_*` / `mcp_hook_list_masks_operator_origin` tests all bind to `template` / `params` / `origin` MCP shapes that the MCP tool itself still exposes (Sprint 4 will rewrite `mcp.rs`). Salvaging them at the UDS level today would mean rewriting them as `tests/uds_authz.rs` tests, which is what the new file already does for HOOK-CREATE / HOOK-DELETE. The MCP-vs-CLI delta will land naturally in Sprint 4.

### Potential Bugs

1. **[BLOCKER] `tests/uds_authz.rs:125-133` — wrong header name in `MAILBOX-CREATE` / `MAILBOX-DELETE` test frames**. Both `raw_mailbox_create_frame` and `raw_mailbox_delete_frame` emit `Mailbox: {name}\r\n`, but the codec parses `Name: <mailbox-name>` (see `src/send_protocol.rs:1014` and `parse_mailbox_crud_headers`). The test frames are MALFORMED at parse time and never reach `enforce_root`. **All four MAILBOX-* tests pass for the wrong reason** (parser error contains "MALFORMED" which doesn't contain "EACCES" → the negative-test assertion fails too) — this is what's causing the red CI. Fix: change `Mailbox:` → `Name:` in the two frame helpers.
2. **[BLOCKER, soft] Body-parse error for legacy hook fields lacks the `book/hooks.md` pointer**. PRD §6.8 says removed fields produce errors that point at the docs. `Config::load` does this ("See book/hooks.md for the supported raw-cmd schema"). The new UDS body parser surfaces only the generic `serde` unknown-field error. The implementer surfaced this for adjudication; given the PRD intent is operator UX consistency, I'm calling it a blocker. One-line fix in `hook_handler.rs::handle_hook_create` body-parse error path.
3. **[NON-BLOCKER] Wire reason text on authz failure leaks owner identity**. PRD §6.3 FR-9: "Verbs whose caller does not authorize return `AIMX/1 ERR FORBIDDEN <reason>` with no information leakage about whether the mailbox exists." `require_mailbox_owner_or_root` (`src/uds_authz.rs:209`) returns `"caller 'bob' is not the owner of mailbox 'alice' (owner: 'root')"` — that leaks (a) that the mailbox `alice` exists (because the rejection only fires after a successful mailbox lookup, and (b) who its owner is. Adjudication: I read the PRD as "must be tightened" — the wording is explicit. Tightening is one line in `require_mailbox_owner_or_root`: collapse the `Err` arm to `reason: "not authorized".into()`. Diagnostic detail still lives in the `tracing::warn!` log for the operator's eyes only. **Calling this non-blocker for now** because (i) the leak is mild (anyone can already enumerate mailbox names by tab-completing in `aimx mailboxes list` if they have shell, and a defender can read `config.toml` if they have root), and (ii) the implementer flagged it explicitly for adjudication. If you'd rather treat it as a blocker, that's defensible — the PRD wording supports it.
4. **[NON-BLOCKER] `MAILBOX-CREATE name="catchall"` is not explicitly rejected**. `mailbox.rs::validate_mailbox_name` does not reject the literal string `"catchall"`, and `handle_create` only checks `current.mailboxes.contains_key(name)` — which would succeed in rejecting "catchall" *only when* the catchall is already configured. On a stripped-down config (no catchall stanza, possible in `--data-dir` test setups), `MAILBOX-CREATE` could synthesize a stanza named `catchall` for a non-catchall address. Defense-in-depth would be to reject the literal name `catchall` at `validate_mailbox_name`. Not on the critical path for this sprint; flag for Sprint 4 or later.
5. **[NON-BLOCKER] `find_hook_owner` is O(N×H)**. `handle_hook_delete` walks every mailbox × every hook to resolve the effective name. Fine for now (mailboxes and hooks both small), but worth tracking if hook counts ever grow.

### Security Issues

- **[Low] Owner-leak in EACCES reason** — covered above (#3).
- No new secrets, no new attack surface beyond what PRD §6.3 designed. UDS socket mode `0o666` is documented and intentional; per-verb `SO_PEERCRED` is the boundary.

### Code Quality

- `hook_handler.rs` is well-structured: authz-first → catchall guard → event/name validation → body parse → struct-build → single-hook validate → lock acquisition → re-resolve under lock → mutate → `validate_hooks` (whole-config) → atomic write → `ConfigHandle::store` → log. Mirrors `mailbox_handler::dispatch` shape closely.
- Lock hierarchy verified: `state_ctx.lock_for(name).await` (per-mailbox) **then** `CONFIG_WRITE_LOCK.lock()` (process-wide). Same order as `mailbox_handler::dispatch` (lines 109-119). No deadlock risk under concurrent CRUD.
- `find_hook_owner` could clone less (returns `String` instead of borrowing) — minor.
- Tests use `Caller::with_username` to synthesize an "attacker" with uid 99999 and a fixed username — clean pattern, good fixture hygiene.

### Alignment with PRD

- FR-7 (hooks always exec as `mailbox.owner_uid()`): unchanged since Sprint 2. ✓
- FR-8 (per-verb authz matrix): code matches; CI test of MAILBOX-* is broken (blocker #1).
- FR-9 (no info leak on FORBIDDEN): violated by current reason text (non-blocker #3).
- FR-15/16/17 (MCP surface): Sprint 4 territory; correctly out of scope here.
- §6.7 / §6.8 (legacy field rejection symmetry): missed in body-parse error message (blocker #2).
- The "MAILBOX-CRUD over UDS now requires root" pre-launch breaking change **is** documented in the PR body's "Pre-launch breaking change" section. ✓

The PRD §6.3 also notes the spec wire shape is `AIMX/1 ERR FORBIDDEN <reason>`, but the codec emits `EACCES` (per `src/send_protocol.rs:245`) and the tests assert `EACCES`. This is a doc-vs-code mismatch that predates this PR (existing for MARK-* and SEND); flagging as informational only — not a Sprint 3 issue, but worth tracking against the PRD for whoever updates `book/`.

### Human Review Comments

No human comments on the PR or via linked issue.

### Verifications Run Locally

- `cargo build --tests` — clean
- `cargo test --bin aimx` — 895 passed, 0 failed, 8 ignored
- `cargo test --tests` — 80 passed, 0 failed (integration), 19 ignored (uds_authz, gated)
- `services/verifier`: build, clippy, fmt, test — all clean (43 passed)
- Banned-token sweep: zero hits across non-`docs/` files.
- `gh pr checks 172`: `mailbox-dir-perms-isolation` **FAIL** (the four MAILBOX-* test failures); core-tests / verifier-tests / docs-build / CodeQL **pass**.

**Blockers:**
1. **`tests/uds_authz.rs:125-133` wrong header name** — `Mailbox:` → `Name:` in `raw_mailbox_create_frame` and `raw_mailbox_delete_frame`. Without this the four MAILBOX-* tests are not exercising authz at all and CI is red. Estimated fix: 2 lines.
2. **Body-parse rejection of legacy fields lacks `book/hooks.md` pointer** — match the symmetry promised by PRD §6.8 / acknowledged in the PR body. Either intercept the known-removed fields explicitly with a tailored message before `serde_json::from_slice`, or post-process the serde error when it mentions a removed field name. Add a test asserting `book/hooks.md` appears in at least one rejection reason.

**Non-blockers:**
1. Reason text leak in `require_mailbox_owner_or_root` — keep the structured log, collapse the wire reason to `not authorized`. PRD §6.3 wording is explicit; calling this non-blocker only because the implementer flagged it for adjudication and the leak is mild.
2. `MAILBOX-CREATE name="catchall"` should be explicitly rejected at `validate_mailbox_name` (defense in depth — Sprint 4 or later).
3. No wire-level test for `HOOK-CREATE` against the catchall (unit test exists, sufficient).

**Nice-to-haves:**
- Track the PRD-says-FORBIDDEN-vs-code-says-EACCES doc mismatch for the eventual `book/` rewrite in Sprint 6.
- `find_hook_owner` returns owned `String`; could borrow.
- Per-test parallelism for `tests/uds_authz.rs` (per-test socket dir) — `--test-threads=1` works; not a blocker.

Once blockers 1 and 2 are addressed, this is ready to merge. The architectural shape is right and the unit-test coverage on `hook_handler` is excellent.

uzyn:
## Cycle-2 fixes

Addresses the 2 blockers and 2 of 3 non-blockers from the cycle-1 review.

### Blocker 1 — `Mailbox:` -> `Name:` in MAILBOX-* test frames
`tests/uds_authz.rs::raw_mailbox_create_frame` / `raw_mailbox_delete_frame` now emit `Name:` (the codec's required header per `parse_mailbox_crud_headers` in `send_protocol.rs`). All four `mailbox_*` tests now exercise the authz path end-to-end. Verified locally with `sudo -E AIMX_INTEGRATION_SUDO=1 cargo test --test uds_authz -- --ignored --test-threads=1` — all 19 tests pass.

### Blocker 2 — `book/hooks.md` pointer in body-parse rejections
`src/hook_handler.rs` now pre-screens the JSON body for the known-removed legacy fields (`template`, `params`, `run_as`, `origin`, `dangerously_support_untrusted`) before serde, and emits error text that mirrors `Config::load` / `reject_legacy_schema`. Migrated operators now see the same actionable message regardless of surface (CLI or UDS). `serde(deny_unknown_fields)` remains as a backstop for any other typo. Five focused unit tests assert the canonical messages (including `book/hooks.md` for template/params).

### Non-blocker 1 (PROMOTED) — Wire reason no longer leaks owner/mailbox identity
`src/uds_authz.rs::enforce_mailbox_owner_or_root` and `enforce_root` now return the opaque canonical reason `not authorized` on the wire while keeping the rich diagnostic ("caller 'bob' is not the owner of mailbox 'alice' (owner: 'root')") flowing to the structured tracing log via `log_decision`. Operators retain full context in `aimx logs`; the daemon no longer leaks mailbox existence or owner identity over the socket. New unit test asserts the wire reason contains neither caller, mailbox, nor owner names.

### Non-blocker 2 — Reserved-name rejection for `catchall` / `aimx-catchall`
`src/mailbox.rs::validate_mailbox_name` now rejects both literals so `MAILBOX-CREATE` cannot synthesize a stanza that shadows the wildcard slot or collides with the reserved system user. Validation runs before authz, so the rejection fires regardless of caller. Two unit tests cover the rejection (one per name) plus a third asserting it fires before authz (rejecting a non-root caller with the validation reason rather than EACCES). The existing `delete_catchall_forbidden` test was updated to assert against the new layer; `create_duplicate_mailbox_returns_mailbox_error` was reworked to seed a non-reserved mailbox first instead of reusing the catchall stanza.

### Non-blocker 3 — Wire-level HOOK-CREATE-on-catchall test (intentionally skipped)
The existing in-process unit test `create_rejects_hook_on_catchall` covers the rejection. Adding a wire-level duplicate would require provisioning the catchall mailbox in the `tests/uds_authz.rs` fixture and would be ~30 lines for a defense-in-depth check the unit test already proves. Noted as deliberate per the cycle-1 review's own guidance.

### Local verification
- `cargo test --bin aimx`: 901 passed, 0 failed, 8 ignored
- `cargo test --tests`: 80 passed, 0 failed (integration); 19 ignored (uds_authz, gated)
- `sudo -E AIMX_INTEGRATION_SUDO=1 cargo test --test uds_authz -- --ignored --test-threads=1`: 19 passed (the four previously-red `mailbox_*` cases now green)
- `services/verifier`: build / test (43 passed) / clippy / fmt all clean
- Banned-token sweep: zero hits across non-`docs/` files

uzyn (commented):
## Re-review of cycle-2 fixes

All cycle-1 issues resolved. CI green on commit `232c7e3`.

### Resolved

**Blocker 1 — `Mailbox:` -> `Name:` header in MAILBOX-* test frames.**
`tests/uds_authz.rs:125-133` now emits the correct `Name:` header. Verified the comment block above the helpers explicitly calls out the codec asymmetry (`MAILBOX-CRUD` uses `Name:`; other verbs use `Mailbox:`) so a future reader does not "unify" them. CI run confirms all four `mailbox_*` tests now pass and exercise the actual `enforce_root` path end-to-end.

**Blocker 2 — `book/hooks.md` pointer in body-parse rejections.**
`reject_legacy_body_fields` in `src/hook_handler.rs:294-336` pre-screens the JSON body and emits messages that mirror `Config::load`. I diffed the two surfaces side-by-side:

- `run_as`: byte-identical (`"hook sets `run_as`; the field was removed — hooks now run as the mailbox's `owner`"`).
- `origin`: byte-identical (`"hook sets `origin`; the field was removed"`).
- `dangerously_support_untrusted`: byte-identical (`"hook sets `dangerously_support_untrusted`; the field was renamed to `fire_on_untrusted`"`).
- `template`/`params`: semantically identical and includes the `book/hooks.md` pointer; the only delta is that `Config::load` formats the field name with `{key}` substitution per individual key while the UDS handler emits the joined string `template`/`params` once. Functionally equivalent — operator gets the same actionable text on either surface.

5 unit tests assert the canonical messages. `serde(deny_unknown_fields)` retained as backstop.

**Promoted non-blocker 1 — Wire reason no longer leaks owner/mailbox identity.**
`enforce_mailbox_owner_or_root` and `enforce_root` (`src/uds_authz.rs:322-387`) now construct a fresh `AuthzReject { reason: WIRE_REASON_NOT_AUTHORIZED.to_string(), .. }` after `log_decision` has already consumed the rich diagnostic. I traced the `reject.reason` flow into all four handlers (`send_handler:209`, `state_handler:155`, `hook_handler:73 + 242`, `mailbox_handler:105`) and confirmed each one places the wire reason directly into `Err.reason` — the canonical `not authorized` string is what crosses the socket. The structured tracing log retains the full `"caller 'bob' is not the owner of mailbox 'alice' (owner: 'root')"` text via `log_decision(..., LogDecision::Reject, Some(&reject.reason))` *before* the wire reason is overwritten. Two new unit tests assert (a) the wire reason equals `"not authorized"` exactly and (b) it contains neither caller nor mailbox nor owner identity. The integration tests in `tests/uds_authz.rs` substring-match on `EACCES` (the error code, not the reason) so the tightening is transparent to them.

**Non-blocker 2 — Reserved-name rejection for `catchall` / `aimx-catchall`.**
`validate_mailbox_name` (`src/mailbox.rs:44-48`) rejects both literals. Validation runs before authz (verified at `mailbox_handler.rs:88` validates → line 102 enforces root), so a non-root caller submitting `MAILBOX-CREATE name=catchall` gets `Validation: reserved` rather than `EACCES`. Three new unit tests cover both names and the validation-before-authz ordering. Case-sensitivity: the literal check is case-sensitive, but the subsequent `is_ascii_lowercase()` character validator rejects uppercase variants like `Catchall` with the "invalid character" error — so no case-variant slips through. The `delete_catchall_forbidden` test was updated to assert against the new validation-layer rejection text; `create_duplicate_mailbox_returns_mailbox_error` was reworked to seed `alice` instead of reusing the catchall stanza.

**Non-blocker 3 — Wire-level `HOOK-CREATE`-on-catchall test (intentionally skipped).**
Per cycle-1 reviewer guidance; `create_rejects_hook_on_catchall` unit test covers the rejection in-process.

### CI

All 8 checks pass on `232c7e3`: `core-tests`, `verifier-tests`, `docs-build`, `mailbox-dir-perms-isolation`, `Analyze (rust)` / `Analyze (actions)` / `Analyze (javascript-typescript)`, `CodeQL`.

### Banned-token sweep

Zero hits introduced by this PR (verified by diffing `+` lines against `\bSprint\s+\d+|S\d+[-.]\d+|FR-\d+[a-z]?|User Story|Acceptance criteria|PRD §` — pre-existing `PRD §X.Y` references in code predate this PR and are out of scope here). PR title, body, and commit messages clean.

### Verdict

**Ready to merge.** Cycle-1 issues addressed correctly; the wire-text leak in particular is a meaningful PRD-required closure handled cleanly via the wire/log split pattern that future verbs can reuse. Architectural shape is right and `hook_handler` test coverage is excellent (16 unit tests + the cycle-2 additions).

```
Enforce SO_PEERCRED authorization on every UDS verb

Wires per-verb SO_PEERCRED authorization into every UDS handler in
`aimx serve`. The wire policy is now uniform: caller uid owns the
target mailbox, OR is root. Applies to SEND / MARK-READ / MARK-UNREAD /
HOOK-CREATE / HOOK-DELETE; MAILBOX-CREATE / MAILBOX-DELETE tighten to
root-only.

- hook_handler.rs: un-stubs HOOK-CREATE / HOOK-DELETE under the new
  authz predicate. Body parses with serde(deny_unknown_fields), with a
  pre-screen that emits the same actionable error text as
  Config::load for legacy template / params / run_as / origin /
  dangerously_support_untrusted fields (operators get the same
  book/hooks.md pointer regardless of surface).
- uds_authz.rs: split wire vs. log on every authz reject. Rich
  diagnostic ("caller 'bob' is not the owner of mailbox 'alice' …")
  flows to the structured tracing log via log_decision; wire response
  is the opaque canonical "not authorized" so the daemon does not
  leak mailbox existence or owner identity.
- mailbox.rs: validate_mailbox_name now rejects the reserved literals
  `catchall` and `aimx-catchall`; validation runs before authz so the
  rejection text is the same regardless of caller.
- tests/uds_authz.rs: per-verb x per-outcome matrix
  (SEND, MARK-READ, MARK-UNREAD, HOOK-CREATE, HOOK-DELETE × owner /
  other / root, plus root_ok / non_root_forbidden for MAILBOX-CREATE /
  MAILBOX-DELETE), gated behind AIMX_INTEGRATION_SUDO=1. CI workflow
  gains a uds_authz step under the existing
  mailbox-dir-perms-isolation job.

Pre-launch breaking change: MAILBOX-CREATE / MAILBOX-DELETE over UDS
now require root. Pre-launch operators must invoke
`sudo aimx mailboxes create | delete` (the CLI gate already exists).
```

PR #174, opened 2026-04-26
Drop hook-template setup UI; add per-agent hook recipes to skill bundles

uzyn:
## Summary

- Drop the (already-vacated) hook-template UI and `aimx-hook` system user from the setup wizard. Add explicit `!unit.contains(\"aimx-hook\")` guards to the systemd / OpenRC unit-generation tests, and an `Info`-severity `aimx doctor` probe (`LEGACY-AIMX-HOOK-USER`) that surfaces a one-liner when a stale `aimx-hook` user is still on the host.
- Bake a "Wiring yourself up as a mailbox hook" section into every supported agent's skill bundle. Native-stdin agents (Claude Code, Codex CLI, Gemini CLI, Goose) ship `stdin: \"email\"` recipes; inline-only agents (OpenCode, Hermes) ship `stdin: \"none\"` recipes that read `\$AIMX_FILEPATH`. OpenClaw ships a documented gap. Each section closes with a self-loop reminder ("re-run `aimx agents setup <agent>` as the mailbox owner before creating the hook").
- Rewrite `agents/common/aimx-primer.md`, `references/hooks.md`, `references/mcp-tools.md`, `references/workflows.md`, and `references/troubleshooting.md` for the owner-first model: drop `hook_list_templates` / `mailbox_create` / `mailbox_delete`, replace `dangerously_support_untrusted` with `fire_on_untrusted`, document mailbox CRUD as root-only on the host CLI, and unignore the two primer-content tests so the new shape is regression-guarded.
- README.md and CLAUDE.md: refresh the stale `setup.rs` blurb and the README's hook example to the current argv-array `cmd` + `fire_on_untrusted` shape.

## Notes / deferred items

- Flag verification sweeps (`<agent> --help` against the current Claude Code / Codex / Gemini / Goose installs) are **deferred to the end-to-end smoke step in the next track sprint**. This cycle is primarily docs; the recipes track verified upstream documentation as of Apr 2026.
- Hermes `chat -q` stdin verification is similarly deferred. The Hermes recipe ships in the inline-only form per the AC's fallback path; a one-line switch to `stdin: \"email\"` is a future follow-up if a real install confirms `-q` accepts piped stdin.
- `book/` chapters still describe the legacy template surface; that rewrite is the next track sprint's scope (already in the plan as the Book + final integration sprint), not this PR.

## Test plan

- [x] `cargo build` clean (binary embeds `agents/` via `include_dir!`).
- [x] `cargo test` clean: 943 unit + 88 integration passing; 19 ignored UDS-authz tests + 1 mailbox-isolation test gated on host test users.
- [x] Verifier crate: `cargo test`, clippy, fmt all clean.
- [x] Banned-token sweep returns zero hits: `rg -n '\bSprint\b|\bS[0-9]+[-.][0-9]+\b|\bFR-[0-9]+[a-z]?\b|User Story|Acceptance criteria' --glob '!docs/**' --glob '!target/**' --glob '!.git/**'`.
- [ ] End-to-end smoke (each agent's `--help` flag verification + a real send→hook→agent fire) — deferred to the next track sprint.


uzyn (commented):
## Sprint Goal Assessment

The sprint's goal — "strip hook-template UI and `aimx-hook` user creation out of `aimx setup`; teach each agent's skill bundle how to write its own `cmd` recipe so agent-driven `hook_create` calls produce correct argv without templates" — is mostly hit. The setup-side cleanup (S5-1) is clean, the doctor probe + regression guards are solid, and the `agents/common/` rewrite (S5-3) faithfully drops every legacy surface. **However, every per-agent recipe introduced by S5-2 is functionally broken: each one's `cmd[0]` is a bare command name (`"claude"`, `"codex"`, `"gemini"`, `"opencode"`, `"hermes"`, `"goose"`) that will be rejected by `Config::load`'s `is_absolute()` validator and by the daemon's `validate_resolved_argv()`.** An agent reading the skill at session start and faithfully calling `hook_create(cmd=["claude", "-p", …])` will hit `hook has non-absolute 'cmd[0]' 'claude'`. The recipes do not run as written — which directly contradicts the S5-2 AC "the literal `hook_create` argument shape for that agent."

**S5-1: setup.rs cleanup — drop hook-template UI, drop aimx-hook user creation**
- [x] `run_setup` template-checkbox phase deleted; structural AST regression guard (`legacy_template_phase_is_absent_from_setup`) walks the parsed module rather than substring-grepping, so future literals can't trip it. Verified locally — passes.
- [x] systemd unit + OpenRC script no longer reference `aimx-hook`; new `assert!(!unit.contains("aimx-hook"))` / `assert!(!script.contains("aimx-hook"))` guards.
- [x] `agents_setup.rs` post-install template-enable hint already gone (cycle-3 Sprint 1); test assertions migrated from `mailbox_create` → `mailbox_list()`.
- [x] Re-run on host with leftover `aimx-hook` user does not fail; `doctor::check_legacy_aimx_hook_user` emits Info-severity `LEGACY-AIMX-HOOK-USER` finding without flipping exit code. Verified locally.

**S5-2: Per-agent skill recipe updates — verified `cmd` recipes for 6 agents (+ 1 documented gap)**
- [x] All 7 agent skill bundles got the "Wiring yourself up as a mailbox hook" section (Claude Code, Codex, Gemini, Goose, OpenCode, Hermes, OpenClaw).
- [x] Native-stdin / inline-only split is honored: Claude Code / Codex / Gemini / Goose use `stdin: "email"`; OpenCode + Hermes use `stdin: "none"` and instruct the agent to read `$AIMX_FILEPATH`.
- [x] OpenClaw ships the documented gap with upstream URL; no `cmd` recipe.
- [x] Self-loop reminder ("re-run `aimx agents setup <agent>` as that user") present at the end of every section, including the OpenClaw gap.
- [x] Common primer pointer added; `agents/common/aimx-primer.md` "Self-trigger as a mailbox hook" subsection points each agent at its own skill section.
- [ ] **NOT MET — verified `cmd` argv shape.** Every recipe's `cmd[0]` is a bare command name. The daemon's `Config::load` (`src/config.rs:629`) and `Hook::resolve_argv` (`src/hook.rs:174`) both require `cmd[0]` to be absolute, with explicit error text "use an absolute path (e.g. `/bin/echo` instead of `echo`)". Recipes as shipped are rejected at config load. **Blocker.**
- [ ] Verification sweep deferred to S6-4 — see adjudication below.
- [ ] Hermes stdin verification deferred — see adjudication below.
- [x] Banned-token sweep clean.
- [x] `cargo build` / `clippy --all-targets -D warnings` / `fmt --check` / `cargo test` all clean locally; CI green.

**S5-3: `agents/common/aimx-primer.md` and `references/hooks.md` rewrite**
- [x] Primer rewritten for owner-first model: drops `hook_list_templates`, `mailbox_create`, `mailbox_delete`; replaces `dangerously_support_untrusted` with `fire_on_untrusted`; documents mailbox CRUD as root-only on host CLI.
- [x] `references/hooks.md` rewritten end-to-end: leads with mailbox-owner ownership, documents argv shape + `AIMX_*` env vars + trust gate + per-agent recipe table. The two-example structure (Claude Code "file + reply" + curl webhook) lands.
- [x] `references/mcp-tools.md`: catalog drops removed tools; `hook_create` signature is `(mailbox, event, cmd, name?, stdin?, timeout_secs?, fire_on_untrusted?)` matching `src/mcp.rs`.
- [x] `references/troubleshooting.md` rewritten for owner-first model; new `EACCES` row in the error-code table.
- [x] `references/workflows.md` §10 now points at root-only host CLI for provisioning.
- [x] Primer-content tests un-ignored (`primer_contains_creating_hooks_section`, `hooks_reference_file_bundled_and_comprehensive`) and tightened with negative assertions on `hook_list_templates` / `dangerously_support_untrusted`. Verified locally — pass.

## Potential Bugs

### Blocker — every per-agent `cmd` recipe has a non-absolute `cmd[0]`

**Files:**
- `agents/claude-code/skills/aimx/SKILL.md.header:18` — `cmd: ["claude", …]`
- `agents/codex/SKILL.md.header:18` — `cmd: ["codex", …]`
- `agents/gemini/SKILL.md.header:17` — `cmd: ["gemini", …]`
- `agents/goose/aimx.yaml.header:34,63` — `cmd: ["goose", …]` (both production and ad-hoc)
- `agents/opencode/SKILL.md.header:18` — `cmd: ["opencode", …]`
- `agents/hermes/SKILL.md.header:24` — `cmd: ["hermes", …]`
- `agents/common/references/hooks.md:74,162` — `["claude", -p", …]` in the JSON example and the "file + reply" worked example
- `agents/common/references/mcp-tools.md:301,332` — `["claude", "-p", …]` in the `hook_create` example

**The problem.** `src/config.rs:629` (Config::load validation) and `src/hook.rs:174` (Hook::resolve_argv) both reject any hook whose `cmd[0]` is not an absolute path:

```
hook has non-absolute `cmd[0]` 'claude': hooks fire from /var/lib/aimx/... so PATH lookup is brittle; use an absolute path
```

Every recipe in this PR — including the JSON `hook_list` example in references/hooks.md — uses bare command names. An agent that copy-pastes the recipe verbatim (which is exactly what the skill instructs: *"Use this exact recipe"*) hits the validator and the hook is never created.

**Why this matters.** The whole point of S5-2 is that the agent calls `hook_create` with the recipe baked into the skill, "no further research needed at runtime" (sprint plan, line 236). With bare `cmd[0]`s, the agent *does* need additional runtime work — discover the absolute path of its own binary, substitute, and only then call `hook_create`. The recipes as shipped do not satisfy the AC.

**Suggested fix.** Either:
1. Use a `<absolute-path-to-claude>` placeholder analogous to `<instruction-string>` and add a note instructing the agent to resolve via `which claude` (or platform equivalent) before calling `hook_create`. The Notes section then explicitly tells the agent to do that substitution; OR
2. Document a concrete canonical install path (e.g., `/usr/local/bin/claude`) per agent with a "if your install lives elsewhere, swap accordingly" caveat; OR
3. Spec a small daemon-side relaxation that resolves bare `cmd[0]` via `PATH` (less attractive — the validator's reasoning is sound, hooks fire from `/var/lib/aimx/...` so PATH is brittle).

The README example (`["/usr/bin/ntfy", …]`) and the worked `curl` webhook example (`["/usr/bin/curl", …]`) both already get this right; the bug is contained to the new per-agent recipe sections + the worked examples in `hooks.md`/`mcp-tools.md`.

### (Confirming the docstring is consistent with the bug)

`agents/common/references/hooks.md:54` lists `cmd[0] must be an absolute path` as an explicit error. The reference file documents the rule and then violates it three lines below in line 74's worked example. That self-inconsistency is itself a tell — there's no path that fixes one without the other.

## Test Coverage

The new doctor probe (`check_legacy_aimx_hook_user`) is testable only on the side of host state the runner happens to land on — the test acknowledges that and asserts the contract holds either way (at most one finding; if present, severity Info). That's the right call; testing the actual `present`-branch would require the test to mutate the host. Acceptable.

The un-ignored primer-content tests now assert removed-tool / removed-flag invariants (negative assertions on `hook_list_templates`, `dangerously_support_untrusted`, plus a positive assertion on the new `hook_create(mailbox, event, cmd` signature). Solid regression coverage.

**Gap.** There is no test that asserts the per-agent recipes' `cmd[0]` is absolute — exactly the invariant the daemon enforces. A simple compile-time check (parse each `*.SKILL.md.header` / `aimx.yaml.header`, regex out `cmd: \[…\]`, assert `cmd[0]` starts with `/`) would have caught the Blocker above. Worth adding once the recipes are corrected, both as a regression guard for the immediate fix and as a smoke check for future recipe additions.

## Code Quality

`doctor.rs:852-856` has a comment ("Translated separately by `check_legacy_aimx_hook_user`") that documents the intentional no-op for the `LegacyAimxHookUser` load-warning variant. Confirmed correct — no double-emit.

The structural `syn`-based regression guard in `setup.rs:legacy_template_phase_is_absent_from_setup` is genuinely good: by walking the parsed AST instead of substring-matching, it survives future literals and comments that legitimately mention retired names without false-positiving. That's the right abstraction for this kind of guard.

## Alignment with PRD

- PRD §6.7 FR-20 / FR-21 (drop hook-template UI, drop aimx-hook user creation): met.
- PRD §6.7 implicit (`aimx-catchall` is the only system user setup creates): preserved — confirmed by the unchanged catchall logic in setup.rs.
- PRD §8.4 (per-agent recipe table): the table is implemented across the 7 skill bundles + `agents/common/references/hooks.md`. **But the literal recipes diverge from PRD §6.2 FR-5: "`cmd` is an argv array (`cmd = ["/path/to/bin", "--flag", "value"]`). No shell wrapping, no string form, no placeholder substitution."** PRD §6.2 explicitly shows an absolute path. The recipes ship bare command names. This is consistent with the Blocker above — the implementation diverges from the PRD, not just from the daemon validator.

## Adjudication: deferred verification sweeps (Path A vs Path B)

**Path A — accept the deferral.** The two deferred ACs (`<agent> --help` flag verification on Claude Code / Codex / Gemini / Goose; Hermes `chat -q` stdin verification) are both flag-shape checks. S6-4's end-to-end smoke will exercise the recipes against real agents and catch flag-name drift incidentally. Folding into S6-4 is reasonable on cost-benefit grounds, and the cycle-1 prompt explicitly authorized the deferral on grounds that no agents were locally installable. **Accept Path A on its own merits.**

However, that deferral is purely about flag-name drift. The Blocker identified here (non-absolute `cmd[0]`) is a structural problem the recipes have on every host, in every install state — flag verification on a real box would have caught it incidentally (running `which claude` to find the path produces an absolute path that obviously needs to land in `cmd[0]`), but the bug is independent of any specific install. So Path A is acceptable, but it does *not* discharge the recipe correctness AC.

**Overall verdict:** Needs minor fixes (one focused recipe-correctness fix; everything else is in order).

1. Every per-agent `cmd[0]` is a bare command name and is rejected by `Config::load` / `Hook::resolve_argv`. Fix the 7 skill headers (Claude Code, Codex, Gemini, Goose ×2 recipes, OpenCode, Hermes) plus the two worked examples in `agents/common/references/hooks.md` (lines 74, 162) and the `hook_create` example block in `agents/common/references/mcp-tools.md` (lines 301, 332). Use either the `<absolute-path-to-...>` placeholder pattern (with a "resolve via `which`" note) or a concrete canonical path per agent.

**Non-blockers (should fix but not blocking, or address on the same pass):**
1. Add a test that walks the embedded `agents/` blob and asserts every `cmd: [...]` example starts with `/` (regression guard for the same class of bug).
2. Once recipes are absolute, the OpenCode + Hermes inline-prompt argv elements get long; consider noting whether the daemon's `MAX_PARAM_BYTES` (or any equivalent argv-size limit) constrains the inline prompt — I didn't see a limit in the code path, but the prompts in those recipes are large enough to be worth a sanity check at fix time.

1. The Hermes recipe's "switch to `stdin: \"email\"` if `-q` accepts piped input" caveat would benefit from a one-line `hermes chat --help` excerpt once verified — easy to fold into the same S6-4 smoke.
2. References/hooks.md's "file + reply" worked example would benefit from a concrete absolute path example matching whatever pattern the per-agent recipes settle on, so the reference and the per-agent skill stay in sync.

Tooling sweeps (`cargo build`, `cargo test`, `cargo clippy --all-targets -D warnings`, `cargo fmt --check`, the verifier crate, banned-token sweep, CI on PR #174) are all green. The setup.rs cleanup, doctor probe, README hook example, and `agents/common/` rewrite are solid. The Blocker is contained to the per-agent recipe text.

uzyn:
## Cycle-2 fixes pushed (`cd3e3da`)

### Blocker — every per-agent `cmd[0]` was a bare command name

Switched every recipe to `/usr/local/bin/<agent>` (the typical npm/curl-install path) plus a one-line note in each agent skill telling the agent to adjust to match `which <agent>` on its host.

**Files touched:**

- 7 skill headers: `agents/claude-code/skills/aimx/SKILL.md.header`, `agents/codex/SKILL.md.header`, `agents/gemini/SKILL.md.header`, `agents/goose/aimx.yaml.header` (×2 recipes), `agents/opencode/SKILL.md.header`, `agents/hermes/SKILL.md.header`
- 2 worked examples in `agents/common/references/hooks.md` (file-and-reply example + the `hook_list` JSON sample)
- 2 examples in `agents/common/references/mcp-tools.md` (`hook_create` example + `hook_list` example)

OpenClaw still has no `cmd` recipe (documented gap, no change needed). README's hook example was already absolute (`/usr/bin/ntfy`).

### Non-blocker 1 — regression test added

`embedded_skill_recipes_use_absolute_cmd_paths` in `src/agents_setup.rs` walks the embedded `agents/` blob (every `.md` / `.md.tpl` / `.md.header` / `.yaml` / `.yaml.header` / `.toml` file), parses fenced code blocks for `cmd:` / `cmd =` / `--cmd` followed by an array, and asserts the first element starts with `/`. Multi-line argv arrays are handled (so the multi-line `cmd: [ "..." , "..." ]` block in `hooks.md` is checked). Verified locally that swapping `/usr/local/bin/claude` back to `claude` fails the test with a clear message; restored once green.

### Non-blocker 2 — argv-size sanity

Folded into the same regression test: each argv element is capped at 8 KiB (well above the largest current inline prompt at ~200 chars, well below Linux's `MAX_ARG_STRLEN` of 128 KiB). Also added a one-liner in the OpenCode and Hermes notes calling out that the inline-prompt argv element is well under the kernel limit.

### Local sweeps green

- `cargo fmt --check` (root + verifier)
- `cargo clippy --all-targets -- -D warnings` (root + verifier)
- `cargo test` — root: 944 + 88 + 1 passed; verifier: 43 passed
- Banned-token sweep clean across the repo

uzyn (commented):
## Re-review (cycle 2 — `cd3e3da`)

The cycle-1 blocker is resolved and both non-blockers landed. This is ready to merge.

### Verification

**Blocker — bare-name `cmd[0]` in every per-agent recipe**: resolved.

Sampled all 10 recipes the embedded blob ships:

- `agents/claude-code/skills/aimx/SKILL.md.header:18` — `/usr/local/bin/claude`
- `agents/codex/SKILL.md.header:18` — `/usr/local/bin/codex`
- `agents/gemini/SKILL.md.header:17` — `/usr/local/bin/gemini`
- `agents/goose/aimx.yaml.header:34` — `/usr/local/bin/goose` (production recipe)
- `agents/goose/aimx.yaml.header:68` — `/usr/local/bin/goose` (ad-hoc recipe)
- `agents/opencode/SKILL.md.header:18` — `/usr/local/bin/opencode`
- `agents/hermes/SKILL.md.header:24` — `/usr/local/bin/hermes`
- `agents/common/references/hooks.md:162` — `/usr/local/bin/claude` (multi-line)
- `agents/common/references/hooks.md:184` — `/usr/bin/curl` (webhook example, was already absolute pre-cycle-2)
- `agents/common/references/mcp-tools.md:301` — `/usr/local/bin/claude`

OpenClaw's `SKILL.md.header` documents the gap (no recipe shipped) — correct, no change needed. README's `/usr/bin/ntfy` was already absolute. Goose ships two recipes (production + ad-hoc) — both fixed. OpenCode and Hermes (inline-prompt forms) — both fixed.

The "adjust to match `which <agent>`" note is reasonably worded: tells operators what's typical, names a plausible alternate (`/usr/bin/<agent>`), and doesn't bloat the recipe. Slight asymmetry in `goose/aimx.yaml.header`: only the production recipe carries the explicit absolute-path note (lines 39–42), the ad-hoc recipe (lines 65–71) doesn't repeat it. Not blocking — same binary, the production-recipe note covers both — but worth noting if someone later splits the file.

**Non-blocker 1 — regression test**: solid.

`embedded_skill_recipes_use_absolute_cmd_paths` in `src/agents_setup.rs:3368`. I verified independently:

- Walks the right file set: `.md`, `.md.tpl`, `.md.header`, `.yaml`, `.yaml.header`, `.toml` under `AGENTS_DIR`. Asserts `files` non-empty so a wholesale strip of the embedded blob also fails the test.
- Restricts inspection to fenced code blocks (the prose-mention concern from cycle-1 is correctly avoided).
- Recognizes all four argv-opener forms: `cmd:`, `cmd =`, `cmd=`, `--cmd`. The CLI form (`--cmd '[...]'`) is dormant in current content but wired up for future-proofing.
- Bracket scanner is quote-aware: it tracks depth and ignores `]` inside string literals, with backslash escape handling. Multi-line argv arrays parse correctly — I instrumented the test locally and confirmed `checked == 10`, exactly matching the 10 `cmd:`/`cmd =` matches in `agents/`. The two split-across-lines recipes in `hooks.md` are both counted.
- Asserts `cmd[0].starts_with('/')` and emits a violation message naming the file and the offending value. I verified by temporarily flipping `/usr/local/bin/codex` back to `codex` — the test failed with `codex/SKILL.md.header: cmd[0] = "codex" is not an absolute path; daemon Config::load rejects bare command names`. Restored.
- Asserts `checked > 0` so a future PR that accidentally moves all recipes outside the walker's reach (e.g. into a non-`.md` file) also fails.

**Non-blocker 2 — argv-size sanity**: covered.

8 KiB per-element cap in the same regression test (`MAX_ARGV_ELEMENT_BYTES`), which is well under Linux's 128 KiB `MAX_ARG_STRLEN` and well above the largest current recipe element (~200 chars in the OpenCode prompt). Brief notes on argv size were added to OpenCode and Hermes (the two inline-prompt recipes); the others don't need it.

**Other checks**:

- `cargo clippy --all-targets -- -D warnings` clean.
- `cargo test --bin aimx agents_setup::tests::embedded_skill_recipes_use_absolute_cmd_paths` passes.
- Banned-token sweep clean across the repo (`rg` for `\bSprint\b|\bS\d+[-.]\d+\b|\bFR-\d+[a-z]?\b|User Story|Acceptance criteria`, excluding `docs/`).
- All eight CI checks green on `cd3e3da` (CodeQL, Analyze actions/javascript-typescript/rust, core-tests, mailbox-dir-perms-isolation, docs-build, verifier-tests).

### Resolved issues

- Blocker: every per-agent recipe `cmd[0]` is now an absolute path (`/usr/local/bin/<agent>`) with an operator-facing note about adjusting via `which <agent>`.
- Non-blocker: regression test prevents future drift on absolute-path requirement and caps argv element size at 8 KiB.

### Still unresolved / new issues

None.

### Verdict

**Ready to merge.**

```
Drop hook-template setup UI and ship per-agent hook recipes in skill bundles

Strip the hook-template checkbox UI and `aimx-hook` user creation out of
`aimx setup`; teach each agent's skill bundle (Claude Code, Codex, Gemini,
Goose, OpenCode, Hermes, OpenClaw) how to write its own absolute-path
`cmd` recipe so agent-driven `hook_create` calls produce daemon-loadable
argv without templates. Add doctor probe and regression test
(`embedded_skill_recipes_use_absolute_cmd_paths`) that walks every
embedded skill file and asserts each recipe's `cmd[0]` is absolute and
each argv element is under 8 KiB.
```

PR #178, opened 2026-04-27
[Sprint 1] MAILBOX-LIST UDS verb + non-root mailbox_list

uzyn:
## Summary

- Add `AIMX/1 MAILBOX-LIST` UDS verb. The daemon resolves the caller via `SO_PEERCRED` and answers with a JSON array of mailboxes the uid owns; root sees everything.
- New `mailbox_list_handler` reads from the in-memory `Arc<Config>` snapshot and returns rows of `{ name, inbox_path, sent_path, total, unread, sent_count, registered }`. Empty result is `[]`.
- MCP `mailbox_list` becomes a thin UDS client. It no longer reads `/etc/aimx/config.toml` and no longer runs its own authz pre-flight, so non-root agents can enumerate their own mailboxes without `sudo`.
- Daemon-not-running surfaces the canonical `aimx daemon not running. Start with 'sudo systemctl start aimx'` message.
- Codec hardening: `MAILBOX-LIST` rejects any header (including `Owner:` and `Content-Length:`) so a forged owner header cannot defeat the `SO_PEERCRED` filter.
- New `JsonAckResponse` shape for verbs that ship a JSON body alongside the bare-ack response variant.
- Primer + `references/mcp-tools.md` now lead with "read `.md` files directly when you don't need a mutation"; the `mailbox_list` row documents the new JSON shape with a worked next-step example.

## Test plan

- [x] `cargo test` (959 tests) — green
- [x] Verifier service `cargo build` / `clippy` / `fmt` — clean
- [x] New unit tests in `mailbox_list_handler::tests` cover root, owner, stranger, and on-disk count visibility paths.
- [x] New codec tests round-trip `MAILBOX-LIST`, reject `Owner:` and `Content-Length:` headers, and pin the JSON-ack writer wire shape.
- [x] New integration test spawns `aimx serve` + `aimx mcp` and parses the JSON response from `mailbox_list`. A second integration test asserts the canonical missing-socket message when the daemon is absent.

### DIFF
--- a/agents/common/aimx-primer.md
+++ b/agents/common/aimx-primer.md
@@ -19,15 +19,23 @@ startup.
 
 ## Two access surfaces
 
+Mail is stored as `.md` files for a reason — when you do not need a
+mutation, read mailbox files directly with your filesystem tools
+(`Read` / `Glob` / `Grep` / equivalent). MCP is the surface for
+*changes*: send, reply, mark, hook CRUD.
+
 aimx gives you two complementary ways to interact with email:
 
-1. **MCP tools** for all mutations (send, reply, mark read/unread, create
-   hooks). The `aimx` MCP server runs over stdio and is launched
-   on-demand by your MCP client.
-2. **Direct filesystem reads** for reading `.md` email files, scanning
-   directories, or bulk-processing. Mailbox directories are mode `0700`
-   and owned by the mailbox's Linux owner, so you see only the mailboxes
-   owned by the user running your MCP process.
+1. **Direct filesystem reads** for inspecting `.md` email files,
+   scanning directories, or bulk-processing. Mailbox directories are
+   mode `0700` and owned by the mailbox's Linux owner, so you see
+   only the mailboxes owned by the user running your MCP process.
+   `mailbox_list` returns absolute `inbox_path` / `sent_path` values
+   precisely so you can hand them to your filesystem tools without
+   another round-trip.
+2. **MCP tools** for all mutations (send, reply, mark read/unread,
+   create hooks). The `aimx` MCP server runs over stdio and is
+   launched on-demand by your MCP client.
 
 Writes always go through MCP. Never create, modify, or delete `.md` files
 directly. The daemon owns those paths.
--- a/agents/common/references/mcp-tools.md
+++ b/agents/common/references/mcp-tools.md
@@ -36,23 +36,50 @@ are invisible. On a multi-user box they give real isolation.
 
 ### `mailbox_list`
 
-List mailboxes you own with total and unread message counts.
+List mailboxes you own with absolute filesystem paths and message
+counts.
 
 **Parameters:** None.
 
-**Returns:** Formatted list of mailboxes with counts.
+**Returns:** JSON array. One row per visible mailbox with these
+fields:
+
+| Field         | Type   | Description                                                                  |
+|---------------|--------|------------------------------------------------------------------------------|
+| `name`        | string | Mailbox name (the local part).                                               |
+| `inbox_path`  | string | Absolute path to the inbox directory (`/var/lib/aimx/inbox/<name>`).         |
+| `sent_path`   | string | Absolute path to the sent directory (`/var/lib/aimx/sent/<name>`).           |
+| `total`       | number | Total emails in the inbox.                                                   |
+| `unread`      | number | Number of inbox emails with `read = false` in the frontmatter.               |
+| `sent_count`  | number | Total emails in the sent folder.                                             |
+| `registered`  | bool   | `true` for mailboxes in `config.toml`; `false` for stray on-disk dirs only.  |
+
+The empty case returns `[]` (a JSON empty array), never a "no
+mailboxes" string. Mailboxes you do not own are simply absent from
+the array. To create or delete mailboxes, ask the operator to run
+`sudo aimx mailboxes create <name> --owner <user>` or
+`sudo aimx mailboxes delete <name>` on the host.
 
 **Example:**
 ```
 mailbox_list()
-→ "agent (inbox: 12 total, 3 unread; sent: 5 total)
-   reports (inbox: 0 total, 0 unread; sent: 0 total)"
+→ [
+    {
+      "name": "agent",
+      "inbox_path": "/var/lib/aimx/inbox/agent",
+      "sent_path": "/var/lib/aimx/sent/agent",
+      "total": 12,
+      "unread": 3,
+      "sent_count": 5,
+      "registered": true
+    }
+  ]
 ```
 
-Mailboxes you do not own do not appear. To create or delete
-mailboxes, ask the operator to run
-`sudo aimx mailboxes create <name> --owner <user>` or
-`sudo aimx mailboxes delete <name>` on the host.
+**Next step:** read messages directly from disk. Take an `id` from
+`email_list(mailbox: "agent")` (or list the inbox dir) and `Read`
+`<inbox_path>/<id>.md` — no need for `email_read` unless you need
+the daemon to enforce path canonicalisation for you.
 
 ---
 
--- a/book/mcp.md
+++ b/book/mcp.md
@@ -54,7 +54,19 @@ List mailboxes you own.
 
 **Parameters:** none
 
-**Returns:** List of mailboxes with addresses, total count, and unread count. Filtered to caller-owned mailboxes for non-root; root sees everything.
+**Returns:** JSON array. One row per visible mailbox with these fields:
+
+| Field         | Type   | Description                                                                  |
+|---------------|--------|------------------------------------------------------------------------------|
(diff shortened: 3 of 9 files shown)


uzyn (commented):
## Sprint 1 Review — `MAILBOX-LIST` UDS verb + non-root `mailbox_list`

### Sprint Goal Assessment

Sprint 1's goal is met. The MCP `mailbox_list` tool no longer touches `/etc/aimx/config.toml`; it ships an `AIMX/1 MAILBOX-LIST` frame and the daemon answers with a `SO_PEERCRED`-filtered JSON array. Non-root agents enumerate their own mailboxes without `sudo`. Codec discipline (no headers on `MAILBOX-LIST`) and the new `JsonAckResponse` shape are clean and well-tested. Build, clippy, fmt, and the new unit + integration tests all pass locally.

**[S1-1] Add `MAILBOX-LIST` UDS verb to the `AIMX/1` codec**
- [x] `Request::MailboxList` variant with no payload — `src/send_protocol.rs:182`.
- [x] `parse_request_frame` recognizes `MAILBOX-LIST` — `src/send_protocol.rs:402`.
- [x] Round-trip + header-rejection unit tests cover `Owner:` and `Content-Length:` — `mailbox_list_codec_tests`, all 5 tests pass.
- [x] No daemon handler in this story — handler lands in S1-2.
- [x] `cargo test` / `cargo clippy --all-targets -- -D warnings` / `cargo fmt -- --check` clean.

**[S1-2] Daemon-side handler — JSON response with `SO_PEERCRED` owner filter**
- [x] `handle_mailbox_list` in `src/mailbox_list_handler.rs` (the sprint plan explicitly allowed a new file).
- [x] Per-row struct serializes the FR-A3 fields exactly: `name, inbox_path, sent_path, total, unread, sent_count, registered`.
- [x] Root sees all; non-root uses `mailbox::caller_owns` with stray-dir uid fallback via `inbox_owner_uid`.
- [x] Stray dirs visible to root and to the user whose uid owns the inbox dir.
- [x] Empty result is the JSON literal `[]` (verified by `stranger_sees_empty_array`).
- [x] Dispatch in `src/serve.rs:861` writes the JSON ack response.
- [x] Four unit tests cover root, owner, stranger, and on-disk count paths.
- [x] Lints clean.

**[S1-3] MCP `mailbox_list` becomes a thin UDS client**
- [x] `mailbox_list` no longer calls `self.load_config` or `self.authorize_mailbox` — it goes through `submit_mailbox_list_via_daemon`.
- [x] Daemon-not-running message matches `email_send` / `hook_create` exactly.
- [x] `list_mailboxes_with_unread` deleted from `src/mcp.rs`.
- [x] New integration test `mcp_mailbox_list_returns_caller_owned` spawns `aimx serve` + `aimx mcp` in a temp data dir and asserts the JSON shape.
- [x] Old text-format assertion rewritten; new test exists for the missing-socket path (`mcp_mailbox_list_without_daemon_reports_missing_socket`).
- [x] Lints clean.

**[S1-4] Primer FR-F1 lead + `mcp-tools.md` `mailbox_list` row rewrite**
- [x] `agents/common/aimx-primer.md` "Two access surfaces" leads with the exact FR-F1 phrasing verbatim.
- [x] `agents/common/references/mcp-tools.md` `mailbox_list` row rewritten with the JSON field table, an example response, and a "next step" pointing the agent at `<inbox_path>/<id>.md`.
- [x] No reference to `aimx mailboxes list` introduced in the primer.
- [x] `cargo build` rebuilds bundles cleanly.
- [ ] Manual smoke (`aimx agents setup claude --print`) — not explicitly evidenced in the PR description, but the bundle is built from the same `include_str!`/`include_dir!` source so the rebuild covers it.

### Test Coverage

Coverage is solid. The codec layer has 5 round-trip + rejection tests, the handler has 4 unit tests across the visibility matrix, and there are 2 integration tests that exercise the real binary end-to-end (with-daemon and without-daemon paths). Two minor gaps worth flagging — both Non-blocker:

- **Stray on-disk dir visibility for non-root caller is unit-tested only implicitly.** The handler's stray-dir branch (`is_visible_to` falling through to `inbox_owner_uid`) is not exercised by a dedicated test. The `non_root_owner_sees_only_owned_mailboxes` test seeds dirs but the mailboxes are registered, so the fallback path never fires. Adding a test that creates a stray inbox dir (no config entry) chowned to the caller would pin FR-A3's "stray dirs appear with `registered: false`" guarantee for non-root callers.
- **`is_marked_read` cheap-parser divergence from `count_with_unread`.** The handler uses a line-scan that recognizes `read = true` literally; `mailbox::count_with_unread` (used by `aimx mailboxes show`) does a full TOML decode. The two will agree on every well-formed frontmatter aimx writes today, but a hand-edited `read = "true"` (string) would be reported as unread by the new handler and as read by the show command. No test pins this — agree-by-construction with `count_with_unread` would be a useful invariant test.

### Potential Bugs

None of the issues below are blockers — Sprint 1 is internally consistent and the code paths are correct on every input the handler can encounter from aimx-written frontmatter. Calling them out as Non-blocker for the implementer's awareness.

- **Cheap `read = …` parser is permissive about quoted values.** `src/mailbox_list_handler.rs:295-298` parses `read = "true"` (string) as `false` because it compares the trimmed value against the literal `"true"` (without quotes). aimx never writes a quoted bool, so this is unreachable from the daemon's own writers, but it is a divergence from the full TOML decoder used elsewhere. **Non-blocker** — no production frontmatter would hit this; raised for consistency with `count_with_unread`.
- **`splitn(3, "+++")` breaks on a `+++` token inside the frontmatter body.** Same file; same function. The TOML frontmatter contract forbids `+++` inside the block, but the cheap parser is more brittle than the TOML decoder. **Non-blocker**, same reason.
- **`is_visible_to` for catchall.** The PRD wording is "non-root callers never see the catchall — preserves today's behavior" but the sprint plan and the implementation both treat catchall identically to other mailboxes (visible to root or to whoever owns it via `aimx-catchall`). The implementation matches the sprint plan and matches today's `caller_owns` semantics, so this is correct — flagging only because the PRD reads slightly more restrictive at first glance. **Not a bug.**

### Security Issues

None observed. The `MAILBOX-LIST` codec rejects `Owner:` and `Content-Length:` headers on the request frame, so a forged client cannot smuggle an owner hint past the `SO_PEERCRED` filter. `caller.uid` is sourced from the kernel-validated peer credential. No user-supplied input feeds into a path or shell.

### Code Quality

- `submit_mailbox_list_via_daemon` reuses `is_socket_missing` (which catches `NotFound | ConnectionRefused | PermissionDenied`) while older sibling helpers (`submit_via_daemon`, `submit_mark_via_daemon`) still hand-check `ErrorKind::NotFound` only. The new code is the better pattern; consider migrating the older helpers in a follow-up so all daemon-not-running surfaces behave identically when the socket exists but is unreachable. **Non-blocker, follow-up.**
- `decode_mailbox_list_response` reimplements basic `AIMX/1` header parsing (Content-Length read, ERR split). The pattern is broadly the same as `parse_response` but specialised for the JSON-bodied OK frame. Acceptable for the first JSON ack consumer; if Sprint 2's `email_list` ships another JSON ack reader, a shared `parse_json_ack_response` helper would prevent drift.
- `JsonAckResponse::Err` writer is byte-identical to `AckResponse::Err`. The unit test pins this, which is good. As long as the contract holds, clients can use one ERR parser for all ack-style verbs.

### Alignment with PRD

Aligned. FR-A1 through FR-A6 and FR-F1/F2 are all addressed in this sprint. The other FRs (B/C/D/F3/F4) belong to later sprints and are correctly out of scope here.

### Human Review Comments

No human comments on the PR; no linked issue.

- **Blockers:** None.
- **Non-blockers (worth a follow-up but not a re-spin):**
  - Add a unit test that exercises the stray-dir visibility branch for a non-root caller (seed a stray inbox dir with the caller's uid; assert it appears with `registered: false`).
  - Add an invariant test that the cheap `is_marked_read` parser agrees with the full-TOML decoder on every shape aimx itself emits (or replace the cheap parser with the full decoder if perf is not load-bearing here).
  - Migrate `submit_via_daemon` and `submit_mark_via_daemon` to use `is_socket_missing` so all daemon-not-running surfaces behave identically.
- **Nice-to-haves:** Factor the JSON-ack response decoder into a shared helper before Sprint 2 adds a second consumer in `email_list`.

```
[Sprint 1] MAILBOX-LIST UDS verb + non-root mailbox_list

Move mailbox_list off the root-owned config.toml: add an
AIMX/1 MAILBOX-LIST verb the daemon answers with a JSON array
filtered by SO_PEERCRED, and turn the MCP tool into a thin UDS
client. New JsonAckResponse shape; codec rejects Owner: /
Content-Length: headers on MAILBOX-LIST so a forged owner hint
cannot defeat peer-cred filtering. Primer + mcp-tools.md lead
with "read .md files directly when you don't need a mutation"
and document the new JSON shape with a worked next-step example.
```

PR #179, opened 2026-04-27
[Sprint 2] email_list paginated rewrite

uzyn:
## Summary

- Rewrite `email_list` as an index-free, paginated, JSON-shaped MCP tool. Listing N rows reads at most N frontmatter blocks regardless of mailbox size.
- `EmailListParams` drops `unread` / `from` / `since` / `subject`; adds `limit` (default 50, max 200; values above silently clamp) and `offset` (default 0).
- Response is JSON. Inbox rows: `{ id, from, to, subject, date, read }`. Sent rows: `{ id, from, to, subject, date, delivery_status }` — `read` is intentionally absent from sent rows (agents never mark sent mail). Empty mailbox returns the literal `"[]"`.
- Algorithm: `read_dir` enumerates ids without parsing, sort descending lex (filenames are `YYYY-MM-DD-HHMMSS-<slug>` so descending lex == reverse chronological), slice `[offset .. offset + limit]`, then read frontmatter only for that slice. `filter_emails` and the old plaintext `list_emails` helper are deleted.
- Docs rewritten (`agents/common/aimx-primer.md`, `agents/common/references/mcp-tools.md`, `agents/common/references/workflows.md`, `book/mcp.md`, `book/mailboxes.md`) to show the new shape and the client-side filter pattern: list a page, filter `read == false` on the agent side, process. We no longer ask aimx to filter.

## Test plan

- [x] `cargo test` — 1055 tests passing (963 unit + 91 integration + 1 codec, others ignored)
- [x] New unit tests (`src/mcp.rs::email_list_tests`):
  - `page_reads_at_most_limit_frontmatter_blocks` — 10 messages, page-3 reads exactly 3 frontmatter blocks via thread-local counter
  - `ten_thousand_message_page_50_reads_at_most_50_blocks` — pins the perf claim; logs wall time, hard-asserts read count ≤ 50
  - `descending_sort_matches_reverse_insertion_order` — 100 messages with diverse slugs (digits, lowercase, uppercase, punctuation), sort matches reverse insertion order
  - `sent_rows_omit_read_key` — sent rows must not carry `read` (key absence, not `null`)
  - `inbox_rows_carry_read_key` — inbox rows have `read`, no `delivery_status`
  - `empty_mailbox_returns_empty_array_no_reads` — empty mailbox returns `[]` with zero frontmatter reads
  - `missing_mailbox_dir_returns_empty_array` — phantom dir surfaces as `[]`, not an error
  - `offset_beyond_end_returns_empty_array` — offset past end returns `[]` with zero reads
  - `limit_clamps_to_max` — None → 50; 0 → 0; 200 → 200; 201 → 200; u32::MAX → 200
  - `bundle_dir_contributes_one_id_to_listing` — Zola-style bundle dir surfaces as one id; sibling attachments inside the bundle are not listed separately
- [x] Integration tests (`tests/integration.rs`):
  - `mcp_email_list_and_read` rewritten to assert against JSON shape, descending order, and demonstrate the client-side `read == false` filter pattern
  - `mcp_email_list_pagination` exercises `limit=2, offset=1` over 5 messages, asserts the correct slice
  - `mcp_email_list_empty_mailbox_returns_empty_json_array` asserts the literal `"[]"`

### DIFF
--- a/agents/common/aimx-primer.md
+++ b/agents/common/aimx-primer.md
@@ -87,8 +87,12 @@ does not expose `mailbox_create` or `mailbox_delete`.
 
 ### Email tools
 
-- `email_list(mailbox, folder?, unread?, from?, since?, subject?)`: list
-  emails. `folder` is `"inbox"` (default) or `"sent"`. Filters AND together.
+- `email_list(mailbox, folder?, limit?, offset?)`: list a page of email
+  metadata, sorted descending by filename (newest first). `folder` is
+  `"inbox"` (default) or `"sent"`. `limit` defaults to 50, max 200; values
+  above 200 are silently clamped. Returns a JSON array — agents
+  filter client-side (e.g. `read == false` for triage). aimx never
+  scans on your behalf; pass `offset` to page deeper.
 - `email_read(mailbox, id, folder?)`: return the full Markdown file
   (frontmatter + body) for one email.
 - `email_send(from_mailbox, to, subject, body, attachments?)`: compose,
@@ -216,11 +220,15 @@ types, required/optional, and the outbound block.
 ### 1. Check inbox for new mail
 
 ```
-email_list(mailbox: "agent", unread: true)
+email_list(mailbox: "agent")
+# Returns a JSON array — filter to unread on your side.
+# JSON.parse(result).filter(r => r.read === false)
 ```
 
-Loop through results. For each, call `email_read` to get the content, then
-`email_mark_read` when done processing. This is the standard polling pattern.
+We no longer ask aimx to filter; we list a page and the agent decides.
+For each unread row, read the full `.md` directly with your filesystem
+tools (or call `email_read` if you prefer the daemon-mediated path), then
+`email_mark_read` when done. This is the standard polling pattern.
 
 ### 2. Send an email
 
@@ -355,8 +363,9 @@ untrusted mail. When deciding whether to act on an email's content
 
 ## Read / unread
 
-- `email_list` with `unread: true` returns only unread messages. Use this
-  to find new mail.
+- `email_list` returns the `read` flag on each inbox row. To find new
+  mail, page through descending by filename and filter client-side to
+  rows where `read == false`.
 - After processing, call `email_mark_read` to avoid reprocessing.
   `email_mark_unread` reverses the state.
 - Read state lives in the `read` frontmatter field. No separate database or
--- a/agents/common/references/mcp-tools.md
+++ b/agents/common/references/mcp-tools.md
@@ -87,37 +87,52 @@ the daemon to enforce path canonicalisation for you.
 
 ### `email_list`
 
-List emails in a mailbox you own with optional filters. All filters
-AND together.
+List a page of emails in a mailbox you own. Rows are sorted descending
+by filename (newest first). aimx never scans on your behalf — pass
+`offset` to page deeper, and filter client-side on the JSON output.
 
 **Parameters:**
-| Name      | Type    | Required | Description |
-|-----------|---------|----------|-------------|
-| `mailbox` | string  | yes      | Mailbox name (must be owned by you) |
-| `folder`  | string  | no       | `"inbox"` (default) or `"sent"` |
-| `unread`  | bool    | no       | Filter to only unread emails |
-| `from`    | string  | no       | Filter by sender address (substring match) |
-| `since`   | string  | no       | Filter to emails since this datetime (RFC 3339, e.g. `2026-01-01T00:00:00Z`) |
-| `subject` | string  | no       | Filter by subject (case-insensitive substring match) |
+| Name      | Type   | Required | Description |
+|-----------|--------|----------|-------------|
+| `mailbox` | string | yes      | Mailbox name (must be owned by you) |
+| `folder`  | string | no       | `"inbox"` (default) or `"sent"` |
+| `limit`   | u32    | no       | Page size; default 50, hard-capped at 200 (values above silently clamp) |
+| `offset`  | u32    | no       | Number of newest rows to skip; default 0 |
 
-**Returns:** Formatted list of matching emails.
+**Returns:** A JSON array. Inbox rows: `{ id, from, to, subject, date, read }`.
+Sent rows: `{ id, from, to, subject, date, delivery_status }` — `read`
+is intentionally absent from sent rows. Empty mailbox → `"[]"`.
 
-**Example, list unread inbox:**
+**Example, list newest inbox page (default 50):**
 ```
-email_list(mailbox: "agent", unread: true)
-→ "2026-04-15-143022-meeting-notes | From: alice@company.com | Subject: Meeting Notes | 2026-04-15T14:30:22Z
-   2026-04-15-153300-invoice-march | From: billing@vendor.com | Subject: Invoice March | 2026-04-15T15:33:00Z"
+email_list(mailbox: "agent")
+→ '[{"id":"2026-04-15-153300-invoice-march","from":"billing@vendor.com","to":"agent@your.tld","subject":"Invoice March","date":"2026-04-15T15:33:00Z","read":false},
+    {"id":"2026-04-15-143022-meeting-notes","from":"alice@company.com","to":"agent@your.tld","subject":"Meeting Notes","date":"2026-04-15T14:30:22Z","read":true}]'
+```
+
+**Next step:** read messages directly. Take the row's `id` and `Read`
+`<inbox_path>/<id>.md` (the `inbox_path` you got from `mailbox_list`).
+No need to call `email_read` unless you want the daemon to re-canonicalise
+the path for you.
+
+**Example, polling for unread:**
+```
+# List a page, then filter client-side.
+rows = JSON.parse(email_list(mailbox: "agent"))
+unread = rows.filter(r => r.read === false)
+# For each unread row, Read its .md or call email_read, then mark it read.
+```
+
+**Example, paging deeper:**
+```
+email_list(mailbox: "agent", limit: 50, offset: 50)
+# Newest 50 already seen; this returns rows 51..100.
 ```
 
 **Example, list sent mail:**
 ```
 email_list(mailbox: "agent", folder: "sent")
-→ "2026-04-15-160145-re-meeting-notes | To: alice@company.com | Subject: Re: Meeting Notes | 2026-04-15T16:01:45Z"
-```
-
-**Example, filter by sender since a date:**
-```
-email_list(mailbox: "agent", from: "alice", since: "2026-04-01T00:00:00Z")
+→ '[{"id":"2026-04-15-160145-re-meeting-notes","from":"agent@your.tld","to":"alice@company.com","subject":"Re: Meeting Notes","date":"2026-04-15T16:01:45Z","delivery_status":"delivered"}]'
 ```
 
 ---
--- a/agents/common/references/workflows.md
+++ b/agents/common/references/workflows.md
@@ -5,19 +5,35 @@ MCP tool calls in order with example parameters.
 
 ## 1. Triage inbox
 
-Process all unread mail, categorize by sender, and mark as read:
+Process all unread mail, categorize by sender, and mark as read.
+
+We no longer ask aimx to filter; we list a page and the agent decides.
+`email_list` returns a JSON array of metadata (including the `read`
+flag), and the agent client-side filters to `read == false` rows.
 
 ```
-# Get all unread
-email_list(mailbox: "agent", unread: true)
+# List the newest page of inbox metadata as JSON.
+email_list(mailbox: "agent")
+# → [{"id":"...","from":"...","to":"...","subject":"...","date":"...","read":false}, ...]
 
-# For each email in results:
+# Parse and filter to unread rows on your side:
+rows = JSON.parse(result)
+unread = rows.filter(r => r.read === false)
+
+# For each unread row, read the .md directly (cheaper than email_read)
+# using the inbox_path you got from mailbox_list:
+#   Read /var/lib/aimx/inbox/agent/<id>.md
+# Or call email_read if you prefer the daemon-mediated path:
 email_read(mailbox: "agent", id: "<id>")
 # Parse frontmatter: check `from`, `subject`, `trusted`, `auto_submitted`
 # Take appropriate action (reply, forward info, log, ignore)
 
 email_mark_read(mailbox: "agent", id: "<id>")
 ```
+
+If a page (default 50 rows, max 200) does not cover all unread mail,
+pass `offset: 50` and keep paging until the page returns no new
+unread rows or you reach the end of the mailbox.
 
 Triage tip: check `auto_submitted` first. If it is `auto-generated` or
 `auto-replied`, skip replying to avoid infinite loops. Check `trusted` to
@@ -169,13 +185,15 @@ email_mark_read(mailbox: "agent", id: "<bounce-id>")
 Mark every unread email in a mailbox as read:
 
 ```
-email_list(mailbox: "agent", unread: true)
-
-# For each email ID in the results:
+# Page through descending-by-filename, filtering client-side.
+rows = JSON.parse(email_list(mailbox: "agent"))
+unread = rows.filter(r => r.read === false)
+# For each id in unread:
 email_mark_read(mailbox: "agent", id: "<id>")
 ```
 
-There is no bulk mark-read tool. Iterate through each message.
+There is no bulk mark-read tool. Iterate through each message. If 50
+rows is not enough, pass `offset: 50` and keep paging.
 
 ## 9. Check sent mail status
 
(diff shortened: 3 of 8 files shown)


uzyn (commented):
## Sprint 2 Review — `email_list` paginated rewrite

### Sprint Goal Assessment

The sprint goal — "Replace the filter-laden `email_list` with an index-free, paginated, JSON-shaped tool. Listing N rows reads at most N frontmatter blocks regardless of mailbox size" — is achieved. The two-pass algorithm (enumerate ids → sort descending → read frontmatter only for the page slice) is correctly implemented in `enumerate_email_ids` + `list_email_page_json`, and the perf claim is pinned by an instrumented test that hard-asserts `≤ limit` frontmatter reads on a 10k-message mailbox. The JSON shape (separate inbox vs. sent rows, with `read` absent on sent) matches FR-B3 exactly. The schema correctly drops `unread` / `from` / `since` / `subject` and gains `limit` / `offset`. Docs were updated across the primer, references, and `book/`.

**S2-1: Filename-sort + page-bounded frontmatter algorithm**
- [x] `EmailListParams` loses `unread` / `from` / `since` / `subject`; gains `limit: Option<u32>` (default 50, clamped to 200) and `offset: Option<u32>` (default 0). `clamp_limit` honours the silent-clamp contract; `clamp_limit(Some(0))` → 0 returns an empty page without reads.
- [x] Two-pass algorithm: pass 1 `enumerate_email_ids` (no parsing), pass 2 `read_email_frontmatter` only for `[offset .. min(offset+limit, len)]`.
- [x] `filter_emails`, `parse_since`, and the substring matchers are deleted from `src/mcp.rs`.
- [x] No recursion, no `walkdir` — only `read_dir` of the folder root.
- [x] Unit test (`page_reads_at_most_limit_frontmatter_blocks`) seeds 10 messages, asks for `limit=3, offset=2`, asserts exactly 3 frontmatter reads via the thread-local counter and the slice 8/7/6.
- [x] Filename sort regression test (`descending_sort_matches_reverse_insertion_order`) exercises 100 messages with mixed-character slugs (digits, lowercase, uppercase, punctuation `-`/`_`).
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` clean locally.

**S2-2: JSON response shape — inbox vs. sent field divergence**
- [x] Two separate row structs (`InboxListRow`, `SentListRow`) — clean, easy to reason about. `read` is structurally absent from sent rows (struct doesn't define the field), not just `null` — `sent_rows_omit_read_key` parses the JSON and asserts key absence on the parsed object.
- [x] `delivery_status` pass-through reads `OutboundFrontmatter.delivery_status` via the partial `EmailListFm` decoder; fallback to empty string only on malformed frontmatter.
- [x] Empty mailbox / empty page returns `"[]"` (verified by `empty_mailbox_returns_empty_array_no_reads`, `missing_mailbox_dir_returns_empty_array`, `offset_beyond_end_returns_empty_array`, and the integration test `mcp_email_list_empty_mailbox_returns_empty_json_array`).
- [x] Existing `mcp_email_list_and_read` integration test rewritten to assert against the JSON shape and demonstrate the client-side `read == false` filter.
- [x] `inbox_rows_carry_read_key` confirms inbox rows carry `read` and not `delivery_status`.

**S2-3: 10k-message perf test + sort-order regression test**
- [x] `ten_thousand_message_page_50_reads_at_most_50_blocks` seeds 10k inbox files and asserts `reads ≤ 50` on a `limit=50, offset=0` call. Wall-clock is logged, soft-asserted at 5000ms.
- [x] `descending_sort_matches_reverse_insertion_order` covers the diverse-slug sort-order pin.
- [x] Tests run on a non-root cargo-test runner using `tempfile::TempDir`.
- [ ] **PARTIAL:** the sprint plan suggested "New test in `tests/integration.rs` (or `tests/email_list_perf.rs` if a dedicated file is cleaner)". Both new tests live in the `src/mcp.rs::email_list_tests` unit-test module instead. The instrumentation works fine (the thread-local counter is wired correctly for the unit-level `list_email_page_json` call), but the chosen placement diverges from the criterion. Non-blocker.
- [ ] **PARTIAL:** the plan's wall-time soft bound was "e.g. 500ms — 10× the design target"; the PR uses 5000ms (100×). The justification ("CI runners with cold disk + debug builds") is reasonable but loose enough that an order-of-magnitude regression would slip through. Tightening to ~500–1000ms would be more useful.

**S2-4: `workflows.md` polling pattern + `mcp-tools.md` `email_list` row**
- [x] `workflows.md` triage / mark-all-read / "filter by sender since a date" sections rewritten to show `JSON.parse(email_list(...))` + client-side filter, with the verbatim "we no longer ask aimx to filter" note.
- [x] `mcp-tools.md` `email_list` row rewritten with the new parameters, JSON shape, polling-then-direct-read worked example, and a paging-deeper example.
- [x] `book/mcp.md` `email_list` section updated; `book/mailboxes.md` updated to reflect client-side filtering.
- [x] `book/hooks.md` reference is generic (`email_list / email_read`) — no stale filter args.
- [x] No call-site references to the dropped `unread:` / `from:` / `since:` / `subject:` filters anywhere outside future release notes.

### Test Coverage

Coverage is good — every algorithmic branch has a corresponding test:

- `clamp_limit` enumerated for `None`, `0`, `50`, `200`, `201`, `u32::MAX`.
- Bundle-vs-flat enumeration covered by `bundle_dir_contributes_one_id_to_listing` (also confirms sibling attachments inside a bundle are not listed as separate ids).
- Empty / missing / past-end mailboxes covered with explicit `fm_read_count_reset() == 0` assertions to confirm the early-return paths.
- Pagination covered both at unit level and integration level.

Gaps worth noting (none are blockers):

- No test exercises `read_email_frontmatter` returning `Ok(None)` when a bundle's inner `.md` is missing. The code path exists (line 1368) and the consequence is silently skipping the row, which can produce a page shorter than `limit` even when the sorted list has more entries. A unit test that seeds a bundle dir without its inner `.md` and asserts the surrounding rows still surface would pin this contract.
- No test covers `read_email_frontmatter` returning `Ok(None)` when `splitn(3, "+++").parts.len() < 3` (a `.md` file without proper frontmatter delimiters). The behaviour is graceful (skip, don't error), but unverified.
- The empty-`delivery_status` fallback (`unwrap_or_default()` → `""`) on a malformed sent row is unverified.

### Potential Bugs

None found that affect correctness on well-formed aimx-written frontmatter. A couple of sub-blocker observations:

1. **Bundle-row skip can produce a short page** (`src/mcp.rs:1311-1314`, `:1328-1331`). When `read_email_frontmatter` returns `Ok(None)` (missing inner `.md` or non-frontmatter content), the row is silently dropped. The slice `[offset..end]` is fixed before iteration, so a missing inner file shrinks the response without backfilling from the next id. Agents that page on `len(rows) == limit` would not infer "more pages exist" correctly. aimx never writes such a file, but disk corruption or a half-written bundle could trigger it. **Non-blocker** — graceful degradation is the right call here, but worth documenting in the function comment so a future maintainer doesn't "fix" it by backfilling.

2. **`enumerate_email_ids` skips symlinks silently** (`src/mcp.rs:1264-1281`). `entry.file_type()` does not follow symlinks, so a symlinked bundle dir or symlinked `.md` is silently dropped (not `is_dir()` and not `is_file()`). aimx never writes symlinks here, and the conservative behaviour is arguably the right security choice (no path-escape risk via `read_email_frontmatter` since the symlink is never followed). **Non-blocker, with caveat** — if any operator ever copies their inbox via `cp -a` or restores from a backup tool that produces symlinks, those messages will quietly disappear from listings. Worth a one-line comment explaining the policy.

3. **`choose_id` fallback to frontmatter `id`** (`src/mcp.rs:1350-1356`). The fallback runs only when the on-disk filename stem cannot decode to UTF-8 (`path.file_stem().and_then(|f| f.to_str())` returned `None`). That path is already filtered out by `enumerate_email_ids`, which only pushes ids derived from `to_str()`. So the fallback is unreachable from the enumerate path — `filename_stem` is always non-empty when it gets here. Not a bug, but the comment claims a defence that the implementation already prevents upstream. Cosmetic.

### Security Issues

No new attack surface. The MCP-side `authorize_mailbox` pre-flight is preserved on `email_list`, so the existing path-canonicalisation and ownership boundary stays intact. The new code reads files via plain `std::fs::read_to_string` on paths assembled from `read_dir` entries within the mailbox dir — no user-controlled path construction. The `+++` splitter is straightforward and bounded by `splitn(3)`, so a body containing `+++` does not break parsing.

### Code Quality

- `EmailListFm` (the partial frontmatter projection) is a nice optimisation — it avoids decoding the full `InboundFrontmatter` (with attachments arrays, references vectors, etc.) on the listing path. Worth keeping.
- The thread-local `FM_READ_COUNT` instrumentation is `#[cfg(test)]`-gated and only increments inside `read_email_frontmatter`. Tight, won't bleed into production.
- `Folder::Inbox` / `Folder::Sent` branches in `list_email_page_json` duplicate the iteration loop. Could be factored, but it's two short branches and the row-struct types differ — leaving as-is is fine.
- `clamp_limit(Some(0)) → 0` is the right behaviour, and the test pins it. Just be aware: an agent that polls with `limit=0` gets a perpetual empty page; the schema description doesn't call this out.

### Alignment with PRD

- FR-B1 (parameter set) ✓
- FR-B2 (algorithm) ✓ — implementation matches the five-step recipe.
- FR-B3 (JSON shape divergence) ✓
- FR-B4 (single `read_dir`, no recursion) ✓
- FR-F2 (`mcp-tools.md` row rewrite) ✓
- FR-F3 (workflows.md polling pattern) ✓ — the verbatim "we no longer ask aimx to filter; we list a page and the agent decides" appears in `workflows.md:11`.
- §7 perf NFR (page-50 < 50ms cold on 10k) — read-count bound is hard-asserted; wall-time bar is documented in test prose but the assertion is 5000ms (loose).

The PR doesn't touch `email_list`'s `load_config()` / `authorize_mailbox` path, so non-root agents can still hit the same `0640 root:root /etc/aimx/config.toml` issue that motivated Sprint 1's `mailbox_list` fix. PRD §7 only mandates non-root access for `mailbox_list`; `email_list` access is not in scope here. Confirmed against PRD §7 — no divergence.

**Overall verdict:** Needs minor fixes — none of the findings are correctness-affecting on aimx-written frontmatter, but the loose perf bound and test placement deserve a touch.

**Blockers:** None.

**Non-blockers:**
1. Tighten the wall-time soft bound from 5000ms toward ~500–1000ms so a real perf regression is caught (S2-3).
2. Move the perf + sort-order tests from `src/mcp.rs::email_list_tests` to `tests/integration.rs` or `tests/email_list_perf.rs` to match S2-3's stated placement, OR document in the sprint plan that the unit-level placement is intentional. (The thread-local counter would need to become `pub` if moved.)
3. Add a one-line comment in `enumerate_email_ids` explaining the symlink-skip policy, and a one-line comment in `list_email_page_json` explaining that bundle-row skips can shrink the page below `limit` (so a future maintainer doesn't backfill).
4. Optional: a unit test that seeds a bundle dir without its inner `.md` and asserts surrounding rows still surface — pins the "graceful skip" contract.

**Nice-to-haves:**
- The fallback in `choose_id` is unreachable given how `enumerate_email_ids` filters; the comment overstates the defence. Either delete the fallback (let the empty stem propagate, which is itself unreachable) or rephrase the comment.
- Tests for `read_email_frontmatter` returning `Ok(None)` on missing-frontmatter content (`splitn(3) < 3`) and on malformed sent `delivery_status`.

The implementation is in good shape and the algorithmic invariants are correctly pinned. Once the non-blockers above are addressed (or explicitly tabled into the sprint backlog), this is mergeable.

uzyn:
## Cycle 2 — Review Feedback Addressed

Pushed `57a8aa0` on top of `ed6832a`.

### Fixed

1. **Wall-time soft bound tightened from 5000ms to 500ms** — ~10x the <50ms cold-cache design target. An order-of-magnitude regression now trips the assertion. Verified locally (`page-50 on 10k messages` finishes in well under 500ms on a debug build).
2. **Symlink-skip policy comment** added on `enumerate_email_ids` (`src/mcp.rs:1264-1268`) — explains that `file_type()` does not follow symlinks, the silent skip is intentional, and warns a future maintainer not to "fix" it.
3. **"Page may shrink below `limit`" contract** added on the `list_email_page_json` doc comment (`src/mcp.rs:1294-1300`) — explains that a row whose frontmatter is unreadable (`Ok(None)`) is silently dropped without backfilling, and warns this is the right call.
4. **Optional unit test for the bundle-without-inner-md graceful skip** (`bundle_without_inner_md_is_skipped_surrounding_rows_surface`) — seeds a bundle dir with no inner `.md` between two flat siblings; asserts the broken row drops out and the neighbours still surface.

### Also fixed (the two cheap nice-to-haves)

5. **Test for `splitn(3) < 3` malformed-frontmatter skip** (`missing_frontmatter_delimiters_skips_row`) — pins the graceful skip when a `.md` lacks `+++` delimiters.
6. **Test for empty `delivery_status` fallback** (`sent_row_with_missing_delivery_status_falls_back_to_empty_string`) — pins that the `Option<String>` decoder + `unwrap_or_default()` surfaces `""` (not `null`) on a sent row missing the field.

### Tabled into the review backlog

- **Test placement** (perf + sort-order tests in `src/mcp.rs::email_list_tests` rather than `tests/integration.rs` / `tests/email_list_perf.rs`). aimx is a binary-only crate (no `lib.rs`), so external integration tests can only drive the binary via `assert_cmd` — they cannot reach the `#[cfg(test)]` `fm_read_count_reset` instrumentation that pins the read-count invariant. Splitting into a `lib.rs` just to relocate two tests is more disruption than placement aesthetics warrant. The reviewer's "OR document in the sprint plan" alternative is the right call here. Backlog entry added under "Improvements" in `mcp-update-sprint.md` with a note to revisit if the crate ever grows a library target for other reasons.
- **Dead/unreachable `choose_id` fallback** — cosmetic, comment overstates the defence. Tabled to backlog rather than touched in this PR; the implementation is correct, only the comment is misleading.

### Verification

- `cargo test` — 966 unit + 91 integration tests pass, 0 failed.

uzyn (commented):
## Re-review — Cycle 2

Verified `57a8aa0` against the previous review. All four addressed items check out, the two cheap nice-to-haves were also picked up, and the two tabled items have proper rationale + backlog entries.

### Resolved

1. **Wall-time soft bound tightened (5000ms → 500ms).** `src/mcp.rs:1888` now asserts `elapsed.as_millis() < 500`. Comment correctly frames it as ~10x the <50ms cold-cache target. The 10k-message test passes locally on a debug build, so an order-of-magnitude regression will now trip.
2. **Symlink-skip policy comment** at `src/mcp.rs:1265-1269` — concise, accurate, explicitly warns a future maintainer not to "fix" the silent drop, and notes the `read_email_frontmatter` path-escape posture.
3. **"Page may be shorter than `limit`" contract** at `src/mcp.rs:1294-1300` — clearly states the slice is fixed before iteration, that `Ok(None)` rows drop without backfilling, and that this is intentional graceful degradation. Pairs well with the new bundle-skip test below.
4. **Unit test for bundle-without-inner-md skip** (`bundle_without_inner_md_is_skipped_surrounding_rows_surface`) — seeds two flat siblings around an empty bundle dir, asserts the broken row drops and the neighbours surface in correct descending order. Passes.

Plus the two cheap nice-to-haves:

5. **`splitn(3) < 3` malformed-frontmatter skip test** (`missing_frontmatter_delimiters_skips_row`) — a `.md` without `+++` delimiters is silently skipped. Passes.
6. **Empty `delivery_status` fallback test** (`sent_row_with_missing_delivery_status_falls_back_to_empty_string`) — sent row missing the field surfaces as `""` (verbatim, not `null`). Passes.

### Tabled (acceptably)

- **Test placement** for the perf + sort-order tests — reasonable rationale (binary-only crate, no `lib.rs`, `#[cfg(test)]` thread-local counter cannot be reached from `tests/` integration tests via `assert_cmd`). Backlog entry added at `docs/mcp-update-sprint.md:315` with the revisit condition. The previous review explicitly offered this as an acceptable alternative.
- **Dead/unreachable `choose_id` fallback comment** — cosmetic only. Backlog entry at `docs/mcp-update-sprint.md:316`.

### Verification

- `cargo test --bin aimx` — 966 passed, 0 failed.
- `cargo test --bin aimx email_list_tests` — all 13 tests in the module pass, including the tightened 10k-message page-50 perf test.

No regressions introduced by `57a8aa0`. New tests are well-scoped and the new comments are accurate.

### Verdict

**Ready to merge.**

```
Rewrite email_list as paginated, JSON-shaped, index-free MCP tool

Replaces the filter-laden email_list (unread/from/since/subject) with a
two-pass, page-bounded algorithm: read_dir enumerates ids without
parsing, sort descending lexicographically (filenames are
YYYY-MM-DD-HHMMSS-<slug> so descending lex == reverse chronological),
then read frontmatter only for the page slice. Listing N rows reads at
most N frontmatter blocks regardless of mailbox size — pinned by a
10k-message test that asserts the page-50 call hits the disk for
exactly 50 frontmatter blocks under a tightened 500ms wall-time bound
(~10x the <50ms cold-cache design target).

EmailListParams loses unread/from/since/subject and gains
limit (default 50, max 200; values above silently clamp) and
offset (default 0). filter_emails and the old plaintext list_emails
helper are removed.

Response is JSON. Inbox rows: { id, from, to, subject, date, read }.
Sent rows: { id, from, to, subject, date, delivery_status } — read is
intentionally absent from sent rows since agents do not mark sent mail.
Empty mailbox returns the literal "[]". Pages may be shorter than
limit when a row's frontmatter is unreadable (Ok(None)); this is
documented and pinned by tests for missing inner .md, missing +++
delimiters, and missing delivery_status.

Docs (primer, references/mcp-tools.md, references/workflows.md,
book/mcp.md, book/mailboxes.md) rewritten to show the new shape and
the client-side filter pattern: list a page, filter on the agent side,
process. We no longer ask aimx to filter; the agent decides.
```

PR #182, opened 2026-04-27
Clean up post-mcp-update backlog (4 items)

uzyn:
## Summary

Four small backlog items from `docs/mcp-update-sprint.md` Improvements section addressed in one PR. All four are cosmetic / consistency cleanups; no behaviour change in the operator-visible surface.

- **`is_socket_missing` migration (`src/mcp.rs`)** — `submit_via_daemon` and `submit_mark_via_daemon` now call the shared `is_socket_missing` helper instead of hand-checking `ErrorKind::NotFound`, so all MCP UDS clients show the same "aimx daemon not running" message when the socket exists but is unreachable (`ConnectionRefused` / `PermissionDenied`), matching `submit_mailbox_list_via_daemon`'s behaviour. Added a unit test that pins the predicate's accept/reject sets.
- **Delete `SandboxStdin::EmailJson` (`src/platform.rs`)** — the variant had no construction site (`rg EmailJson` confirmed only the declaration). Removed it and its `Email | EmailJson` match arm. Kept a narrow `#[allow(dead_code)]` on `None`, which is constructed only in `run_fallback` tests.
- **`reject_removed_stdin_field` error type (`src/config.rs`)** — function now returns `Result<(), Box<dyn std::error::Error>>` directly so the call site in `Config::load` can drop its `.map_err(|e| -> Box<dyn std::error::Error> { e.into() })?` for symmetry with the surrounding error path. Existing rejection tests pass with the same error wording.
- **`choose_id` comment fix (`src/mcp.rs`)** — chose deletion over rewording. The frontmatter-id fallback was unreachable from `enumerate_email_ids` (only valid UTF-8 stems are pushed). Inlined `id.clone()` at both call sites, dropped the helper, and removed the now-unused `EmailListFm.id` field.

## Test plan

- [x] `cargo test` — 975 unit + 91 integration tests pass.
- [x] No banned planning-doc tokens in the diff.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -563,7 +563,7 @@ fn reject_legacy_schema(toml_text: &str) -> Result<(), Box<dyn std::error::Error
 ///
 /// The error names the hook by its explicit `name` if set, otherwise by
 /// the derived sha256-based effective name (matches `effective_hook_name`).
-fn reject_removed_stdin_field(parsed: &toml::Value) -> Result<(), String> {
+fn reject_removed_stdin_field(parsed: &toml::Value) -> Result<(), Box<dyn std::error::Error>> {
     let Some(mailboxes) = parsed.get("mailboxes").and_then(|v| v.as_table()) else {
         return Ok(());
     };
@@ -591,7 +591,8 @@ fn reject_removed_stdin_field(parsed: &toml::Value) -> Result<(), String> {
             return Err(format!(
                 "hook '{name}' carries removed field 'stdin' — remove this line and \
                  restart aimx serve; the email is always piped to hooks"
-            ));
+            )
+            .into());
         }
     }
     Ok(())
@@ -911,8 +912,7 @@ impl Config {
         // reject the removed `stdin` field with a hook-named error
         // before the structural deserializer trips on its absence.
         let raw_value: toml::Value = toml::from_str(&content)?;
-        reject_removed_stdin_field(&raw_value)
-            .map_err(|e| -> Box<dyn std::error::Error> { e.into() })?;
+        reject_removed_stdin_field(&raw_value)?;
         let config: Config = raw_value.try_into()?;
         for (name, mb) in &config.mailboxes {
             if mb.owner.trim().is_empty() {
--- a/src/mcp.rs
+++ b/src/mcp.rs
@@ -639,7 +639,7 @@ fn submit_via_daemon(args: &SendArgs) -> Result<String, String> {
     };
 
     let outcome = outcome.map_err(|e| {
-        if e.kind() == std::io::ErrorKind::NotFound {
+        if is_socket_missing(&e) {
             "aimx daemon not running. Start with 'sudo systemctl start aimx'".to_string()
         } else {
             format!(
@@ -689,7 +689,7 @@ fn submit_mark_via_daemon(mailbox: &str, id: &str, read: bool) -> Result<(), Str
     };
 
     let outcome = outcome.map_err(|e| {
-        if e.kind() == std::io::ErrorKind::NotFound {
+        if is_socket_missing(&e) {
             "aimx daemon not running. Start with 'sudo systemctl start aimx'".to_string()
         } else {
             format!(
@@ -1211,8 +1211,6 @@ struct SentListRow {
 #[derive(Deserialize)]
 struct EmailListFm {
     #[serde(default)]
-    id: String,
-    #[serde(default)]
     from: String,
     #[serde(default)]
     to: String,
@@ -1303,7 +1301,7 @@ fn list_email_page_json(
                     continue;
                 };
                 rows.push(InboxListRow {
-                    id: choose_id(id, &fm.id),
+                    id: id.clone(),
                     from: fm.from,
                     to: fm.to,
                     subject: fm.subject,
@@ -1320,7 +1318,7 @@ fn list_email_page_json(
                     continue;
                 };
                 rows.push(SentListRow {
-                    id: choose_id(id, &fm.id),
+                    id: id.clone(),
                     from: fm.from,
                     to: fm.to,
                     subject: fm.subject,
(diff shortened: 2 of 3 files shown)


uzyn (commented):
## Standalone PR Review — Post-mcp-update Backlog Cleanup

### Changes Overview

Four small, scoped cleanups to the `mcp-update` backlog:

1. `submit_via_daemon` and `submit_mark_via_daemon` in `src/mcp.rs` migrate from a hand-rolled `ErrorKind::NotFound` check to the shared `is_socket_missing` predicate (which also catches `ConnectionRefused` and `PermissionDenied`), unifying the "daemon not running" surface across all three UDS clients.
2. `SandboxStdin::EmailJson` is removed from `src/platform.rs` along with the `Email | EmailJson` match arm in `run_child_with_timeout`. The `#[allow(dead_code)]` is narrowed from enum-wide to the `None` variant, which is the only one that genuinely has no production constructor.
3. `reject_removed_stdin_field` in `src/config.rs` now returns `Result<(), Box<dyn std::error::Error>>` directly; `Config::load` drops the `.map_err(|e| -> Box<dyn std::error::Error> { e.into() })?` at the call site. Error wording unchanged.
4. The `choose_id` helper in `src/mcp.rs` is deleted along with the `EmailListFm.id` field. Both call sites use `id.clone()` from the `(id, fm_path)` tuple produced by `enumerate_email_ids`.

Net: four small structural improvements, no operator-visible behavior change beyond #1's wider error-kind match.

### Scope Alignment

Matches the PR description and the four ticked items in `docs/mcp-update-sprint.md` Improvements. The four other backlog items (stray-dir visibility test, `is_marked_read` invariant, JSON-ack helper factor-out, perf-test placement) are correctly left unticked. No scope creep.

### Potential Bugs

I verified each change independently:

**`is_socket_missing` migration.** `is_socket_missing` at `src/mcp.rs:942` matches `NotFound | ConnectionRefused | PermissionDenied`. Both migrated callers preserve the same branch shape: predicate hits → "aimx daemon not running. Start with 'sudo systemctl start aimx'". Predicate misses → fall through to the existing detailed-error path. No state change. Behavior delta is purely "more I/O error kinds report as 'daemon not running'", which matches what `submit_mailbox_list_via_daemon` was already doing.

**`SandboxStdin::EmailJson` deletion.** `rg -n 'EmailJson' src/ tests/` returns zero hits after the change. The match in `run_child_with_timeout` correctly collapses to two arms (`Email(b)` and `None`) — no missing branch, no panic-inducing wildcard.

**`reject_removed_stdin_field` error type.** Function still returns the same canonical wording (`"hook 'X' carries removed field 'stdin' — remove this line and restart aimx serve; the email is always piped to hooks"`) — just wrapped via `String::into::<Box<dyn Error>>` instead of returning the raw `String`. Existing rejection tests assert against the wording, not the error type, so they continue to pass.

**`choose_id` deletion.** Verified `enumerate_email_ids` (`src/mcp.rs:1232`) only pushes ids that successfully decoded via `to_str()` — both the `is_dir()` branch and the `is_file()` branch use `path.file_name().and_then(|f| f.to_str())` / `path.file_stem().and_then(|f| f.to_str())` and skip the entry on `None`. Every id in the returned `Vec<(String, PathBuf)>` is therefore a non-empty UTF-8 string, so the `if !filename_stem.is_empty()` guard in the old `choose_id` was dead. Deleting it is the right call.

No blockers, no non-blockers.

### Test Coverage

- New `socket_missing_tests` module pins the predicate's accept set (`NotFound`/`ConnectionRefused`/`PermissionDenied`) and reject set (`TimedOut`/`ConnectionReset`/`Other`). This is the right place to catch a future widening or narrowing of the predicate; it's tighter than testing each caller individually.
- No test for the `EmailJson` deletion or the `choose_id` deletion — appropriate, since both are pure dead-code removal.
- `reject_removed_stdin_field` retains its existing rejection-wording tests; no test changes needed for the error-type swap.

### Code Quality

- Narrowing `#[allow(dead_code)]` from the whole `SandboxStdin` enum to just the `None` variant is the right cleanup — it surfaces any future variant that escapes its construction sites.
- Doc comment on `SandboxStdin::None` ("Constructed only by `run_fallback` tests…") is accurate and aids future readers.

### Security

No new attack surface. The widened `is_socket_missing` could be argued as a marginal info-leak reduction (the underlying error kind is no longer exposed for `ConnectionRefused`/`PermissionDenied`), but the practical impact is nil.

- **Blockers:** none.
- **Non-blockers:** none.
- **Nice-to-have observation (not for this PR):** `is_socket_missing`'s behavior of treating `PermissionDenied` as "daemon not running" is technically a slight overreach — a misconfigured socket policy (e.g., SELinux/AppArmor) would also produce `PermissionDenied`. In practice the daemon's startup chmods the socket world-writable so this is rare, but a future tweak to the doc comment on `is_socket_missing` to acknowledge the policy edge case would help an operator who hits it. Pure documentation; not raising as a backlog item.

```
Clean up post-mcp-update backlog (4 items)

Migrate submit_via_daemon and submit_mark_via_daemon in src/mcp.rs to
the shared is_socket_missing predicate so all UDS clients show the
canonical "aimx daemon not running" message on NotFound /
ConnectionRefused / PermissionDenied. Pin the predicate's accept and
reject sets with a unit-test module.

Delete the unused SandboxStdin::EmailJson variant from src/platform.rs
and collapse the match in run_child_with_timeout to the two live
variants. Narrow #[allow(dead_code)] from the enum to just the None
variant, which is constructed only by run_fallback tests.

Tidy reject_removed_stdin_field's signature in src/config.rs to return
Box<dyn Error> directly so Config::load can drop the explicit
.map_err -> Box<dyn Error> wrapper. Error wording unchanged.

Delete the choose_id helper and the EmailListFm.id field in
src/mcp.rs. enumerate_email_ids only pushes valid UTF-8 stems, so the
fallback was unreachable; both call sites now clone the filename id
directly.
```

PR #183, opened 2026-04-28
feat: surface daemon restart on upgrade + show client/server version drift in doctor

uzyn:
## Summary
Today, upgrading aimx replaces the binary on disk but leaves the operator with no
visibility into whether the running `aimx serve` daemon was actually restarted —
and no way to query the running daemon's version even after the swap. On the
`curl | sh` path the restart can silently no-op (manual `aimx serve`, or
systemctl-present-but-inactive), and on every path `aimx --version` prints only
the *invoked binary's* compile-time version. This PR fills both gaps.

## Requirements
- [x] New `AIMX/1 VERSION` UDS verb (read-only, open to all UDS callers — same posture as `MAILBOX-LIST`).
- [x] Daemon dispatch + thin `version_handler` returning `{tag, git_hash, target, build_date}`.
- [x] Client probe with a short timeout, distinguishing socket-missing from I/O error.
- [x] `aimx doctor` renders `Client version:` + `Server version:` under `SMTP server:`, with an inline warn-coloured drift suffix on mismatch (no Checks entry, no exit-code change).
- [x] `install.sh`: stop/start lines promoted to `say`; systemctl-present-but-inactive path now still calls `start_service` after swap; pgrep-based detection of manually-running `aimx serve` outside systemd/OpenRC (warn only — never signal); single `is-active` check after start.
- [x] `aimx upgrade`: one `term::success` line confirming the restart on the new tag.

## Out of scope
- Killing or signalling a manually-run `aimx serve` from `install.sh`.
- Drift as a hard `aimx doctor` failure (drift is information-only).
- A separate `aimx version --remote` subcommand.
- Auth/authz on the new verb (returns build metadata only).
- Changes to the existing TCP-25 readiness probe.

## Technical approach
- `src/send_protocol.rs`: `Request::Version`, `"VERSION"` parse arm, `VersionResponse`, `write_version_response()`.
- `src/serve.rs::handle_uds_connection_with_timeout`: new `Request::Version` arm.
- `src/version_handler.rs` (new): handler + client probe (or split into `version_handler.rs` + `version_client.rs` depending on what fits).
- `src/doctor.rs`: `StatusInfo` extended; `gather_status_with_ops` invokes the probe; `format_status` renders the two new lines and the drift suffix.
- `install.sh`: surgical edits to `stop_service` / `start_service` / upgrade path.
- `src/upgrade.rs`: success line after `wait_for_service_ready` returns true.

## Test plan
- [x] send_protocol round-trip + reject non-zero Content-Length on VERSION (unit).
- [x] version_handler response framing (unit).
- [x] end-to-end VERSION over UDS against a booted daemon (integration).
- [x] doctor render: matching, drift, unreachable — including exit-code unchanged on drift.
- [x] upgrade: restart confirmation produced on happy path, not produced on rollback.
- [x] shellcheck install.sh.
- [x] CLAUDE.local.md token-sweep regex returns zero NEW hits.

Final gates: \`cargo test --all\` green, \`cargo clippy -- -D warnings\` clean, \`cargo fmt -- --check\` clean.

### DIFF
--- a/book/cli.md
+++ b/book/cli.md
@@ -61,7 +61,7 @@ See [Setup: End-to-end verification](setup.md#end-to-end-verification).
 
 Print server health: configuration path, mailbox counts and unread counts, a per-mailbox table (Mailbox, Address, Total, Unread, Trust, Senders, Hooks), an Ownership section listing each mailbox's `owner` and whether the owner uid resolves on the host, DKIM key presence, SMTP service state, DNS record verification, and a pointer to `aimx logs` for recent service logs. Exits non-zero when any mailbox has an unresolvable owner so monitoring can detect orphans.
 
-The Service section includes `Client version:` (the CLI binary you just invoked) and `Server version:` (probed from the running daemon over the UDS socket). When the two differ, the Server line carries an inline warn-coloured `drift` suffix suggesting `systemctl restart aimx` (or `rc-service aimx restart` on OpenRC) — the operator upgraded the binary on disk but the running daemon is still on the old build. Drift is informational only; it does not change the doctor exit code.
+The Service section includes `Client version:` (the CLI binary you just invoked) and `Server version:` (probed from the running daemon over the UDS socket). Compare them at a glance to see whether the running daemon is still on an older build than the binary on disk — when they differ, restart the service (`systemctl restart aimx` or `rc-service aimx restart` on OpenRC) so the daemon picks up the new binary. The version lines are informational only and do not change the doctor exit code.
 
 No flags.
 
--- a/book/cli.md
+++ b/book/cli.md
@@ -61,6 +61,8 @@ See [Setup: End-to-end verification](setup.md#end-to-end-verification).
 
 Print server health: configuration path, mailbox counts and unread counts, a per-mailbox table (Mailbox, Address, Total, Unread, Trust, Senders, Hooks), an Ownership section listing each mailbox's `owner` and whether the owner uid resolves on the host, DKIM key presence, SMTP service state, DNS record verification, and a pointer to `aimx logs` for recent service logs. Exits non-zero when any mailbox has an unresolvable owner so monitoring can detect orphans.
 
+The Service section includes `Client version:` (the CLI binary you just invoked) and `Server version:` (probed from the running daemon over the UDS socket). When the two differ, the Server line carries an inline warn-coloured `drift` suffix suggesting `systemctl restart aimx` (or `rc-service aimx restart` on OpenRC) — the operator upgraded the binary on disk but the running daemon is still on the old build. Drift is informational only; it does not change the doctor exit code.
+
 No flags.
 
 See [Troubleshooting](troubleshooting.md).
--- a/book/installation.md
+++ b/book/installation.md
@@ -144,9 +144,9 @@ sudo aimx upgrade
 3. Stops `aimx.service` (or the OpenRC equivalent).
 4. Renames the current `/usr/local/bin/aimx` to `/usr/local/bin/aimx.prev` and moves the new binary into place — atomic `rename(2)` so a crash cannot leave a half-written binary.
 5. Restarts the service.
-6. Prints `aimx v<old> → v<new>. Service restarted.`
+6. Waits for the daemon to come back up, then prints a `✓ aimx serve restarted on <tag>` confirmation line followed by `aimx v<old> → v<new>. Service restarted.`
 
-If any step after the stop fails, the rollback path restores `aimx.prev` and restarts the service. A `✗` line names the failed step.
+If any step after the stop fails, the rollback path restores `aimx.prev` and restarts the service. A `✗` line names the failed step. The restart-confirmation line is suppressed on the rollback path.
 
 Flags:
 
--- a/book/release-notes.md
+++ b/book/release-notes.md
@@ -19,7 +19,7 @@ Client version:   v1.2.4 (a1b2c3d4)
 Server version:   v1.2.4 (a1b2c3d4)
 ```
 
-When the tags differ, the Server line carries an inline warn-coloured `drift` suffix suggesting `systemctl restart aimx` (or `rc-service aimx restart` on OpenRC). Drift is informational only — no `DoctorFinding`, no exit-code change. If the daemon is offline the Server line renders `(daemon not running)`; if the probe fails within its 500 ms budget it renders the failure reason in dim text. See [Troubleshooting: Version drift](troubleshooting.md#version-drift-between-client-and-daemon).
+When the tags differ, restart the service (`systemctl restart aimx`, or `rc-service aimx restart` on OpenRC) so the daemon picks up the new binary. The lines are informational only — no `DoctorFinding`, no exit-code change. If the daemon is offline the Server line renders `(daemon not running)`; if the probe fails within its 500 ms budget it renders the failure reason in dim text. See [Troubleshooting: Version drift](troubleshooting.md#version-drift-between-client-and-daemon).
 
 ### `install.sh` upgrade path is louder and self-healing
 
--- a/book/release-notes.md
+++ b/book/release-notes.md
@@ -2,6 +2,42 @@
 
 Version-by-version changelog of operator-visible behavior changes. Use this as the canonical source for "what changed" between aimx releases; individual book chapters describe the current behavior only.
 
+## Unreleased — upgrade visibility
+
+Closes the visibility gap on the upgrade path: operators can now confirm whether the running `aimx serve` daemon was actually restarted on the new binary, and detect drift between the on-disk `aimx` and the still-running daemon.
+
+### `AIMX/1 VERSION` UDS verb
+
+A new read-only verb on `/run/aimx/aimx.sock` returns the daemon's `{tag, git_hash, target, build_date}`. Same authorization posture as `MAILBOX-LIST` — no `SO_PEERCRED` filter, the payload is build metadata only. There is no separate `aimx version --remote` subcommand; consumers go through `aimx doctor`.
+
+### `aimx doctor` renders client + server versions
+
+The Service section now includes two new lines:
+
+```
+Client version:   v1.2.4 (a1b2c3d4)
+Server version:   v1.2.4 (a1b2c3d4)
+```
+
+When the tags differ, the Server line carries an inline warn-coloured `drift` suffix suggesting `systemctl restart aimx` (or `rc-service aimx restart` on OpenRC). Drift is informational only — no `DoctorFinding`, no exit-code change. If the daemon is offline the Server line renders `(daemon not running)`; if the probe fails within its 500 ms budget it renders the failure reason in dim text. See [Troubleshooting: Version drift](troubleshooting.md#version-drift-between-client-and-daemon).
+
+### `install.sh` upgrade path is louder and self-healing
+
+- The stop and start banners are now promoted from `dbg` to `say`, so the operator running `curl | sh` sees the service control happening.
+- The `systemctl is-enabled = true && is-active = false` path now still calls `start_service` after the binary swap. Previously the daemon was never restarted on that path.
+- A new `pgrep`-based detector warns (never signals) when a manually-launched `aimx serve` is running outside systemd / OpenRC.
+- A single post-start `systemctl is-active` check confirms the daemon came up, with a `journalctl -u aimx -n 20` hint on failure.
+
+### `aimx upgrade` confirms the restart
+
+After `wait_for_service_ready` returns true, `aimx upgrade` prints one extra line:
+
+```
+✓ aimx serve restarted on v1.2.4
+```
+
+The line is suppressed on the rollback path so a failed upgrade never claims success.
+
 ## Unreleased — MCP surface cleanup
 
 Three hard breaks tighten the MCP tool surface around aimx's "no index, no scan" design. Canonical tool docs live in [MCP Server](mcp.md); the new hook model lives in [Hooks & Trust](hooks.md).
--- a/book/release-notes.md
+++ b/book/release-notes.md
@@ -8,7 +8,7 @@ Closes the visibility gap on the upgrade path: operators can now confirm whether
 
 ### `AIMX/1 VERSION` UDS verb
 
-A new read-only verb on `/run/aimx/aimx.sock` returns the daemon's `{tag, git_hash, target, build_date}`. Same authorization posture as `MAILBOX-LIST` — no `SO_PEERCRED` filter, the payload is build metadata only. There is no separate `aimx version --remote` subcommand; consumers go through `aimx doctor`.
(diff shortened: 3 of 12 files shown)


uzyn (commented):
## Changes Overview

This PR closes the visibility gap on the upgrade path: an operator running `curl | sh` had no way to confirm whether the running daemon was actually restarted on the new binary, and no way to detect drift between the on-disk `aimx` and the still-running `aimx serve` process. Three surfaces are added:

1. A new `AIMX/1 VERSION` UDS verb (read-only, no authz — same posture as `MAILBOX-LIST`) that returns the daemon's `{tag, git_hash, target, build_date}`.
2. `aimx doctor` now probes the daemon (500 ms hard budget) and renders `Client version:` / `Server version:` under `SMTP server:`, with an inline warn-coloured drift suffix on tag mismatch. Drift is informational only — no `DoctorFinding`, no exit-code change.
3. `install.sh` upgrade path: `say`-promoted stop/start banners, a fix that ensures `start_service` is invoked even when systemd is present-but-inactive (previously the daemon never restarted on that path), a pgrep-based detector for manually-launched `aimx serve` outside systemd/OpenRC (warn-only — never signals), and a final `systemctl is-active` check after start. `aimx upgrade` gets its own `aimx serve restarted on <tag>` confirmation after `wait_for_service_ready` returns true.

## Scope Alignment

Every checked item in the PR description maps to concrete code. Out-of-scope items are honoured: no signalling of manual processes, no drift-as-finding, no `aimx version --remote`, no authz on the new verb.

## Potential Bugs

None found. Specific things I checked:

- The `parse_version_headers` rejects any header line including `Content-Length: 0`, mirroring `parse_mailbox_list_headers`. Good — no future client can ship a header and have it silently dropped.
- `probe_daemon_version_async` checks `socket_path.exists()` first, then maps `NotFound` / `ConnectionRefused` / `PermissionDenied` from the connect path to `SocketMissing`. The race between `exists()` and `connect()` is handled.
- `parse_send_request` (the `#[cfg(test)]`-only SEND-only convenience) gets a new `Request::Version` arm that returns `ParseError::Malformed("expected SEND verb, got VERSION")` — the exhaustive match keeps future verbs honest.
- The post-start `is-active` check in `install.sh` is gated on `_init = "systemd"`, so a host without systemd does not emit a confusing "did not reach active state" warning. OpenRC's `rc-service start` already exits non-zero on failure, so the gate is correct.
- The `detect_manual_aimx_serve` guard returns 0 (skips warning) when either `systemctl status aimx` or `rc-service aimx status` succeeds — it warns only when no init system claims the unit. Reasonable trust boundary.
- `restart_confirmation_line` is unreachable on the rollback path because `attempt_rollback` returns `Err(...)` before the `report.outcome = Some(Outcome::Upgraded)` line — the new `rollback_path_does_not_record_wait_for_ready` test pins this.

## Security Issues

None. The `VERSION` payload is build-time metadata (tag, short hash, target triple, build date) — none of it is sensitive, and the lack of `SO_PEERCRED` filter matches the existing `MAILBOX-LIST` posture documented in the module.

## Test Coverage

Strong. Specifically:

- `send_protocol`: round-trip request, round-trip response, reject `Content-Length` header on `VERSION` (3 unit tests).
- `version_handler`: socket-missing, fake-server round-trip, current-response-through-codec including the 8-char-or-`unknown` git_hash invariant (3 unit tests).
- `doctor::version_render_tests`: matching, drift, unreachable, skipped-when-no-daemon, and `drift_does_not_surface_a_doctor_finding` (5 unit tests — this last one is the right invariant test, asserting drift never reaches `format_checks`).
- `upgrade::tests`: `restart_confirmation_line_carries_tag` and `rollback_path_does_not_record_wait_for_ready` pin both the happy-path message format and the rollback no-print invariant.
- `integration::uds_version_verb_returns_running_daemon_metadata` boots a real `aimx serve`, sends `AIMX/1 VERSION`, and asserts the JSON wire shape plus that the reported tag matches the binary's `--version` output. This is the contract test the doctor probe leans on.

All 13 doctor tests, 10 send_protocol codec tests, and the integration test pass locally. `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check`, shellcheck on `install.sh` (no new findings), and the CLAUDE.local.md token-sweep regex all clean.

## Code Quality

- `current_version_response()`, `probe_daemon_version_async`, and `format_version_lines` are well-scoped and well-commented.
- The async core / sync facade split in `version_handler.rs` (so doctor can call it sync without needing a tokio context) is the right shape and is documented in the doc comment.
- The `Reply::Version` arm is added cleanly to the existing `handle_uds_connection_with_timeout` dispatcher.

## Non-blocker Findings

1. **Drift-hint wording is confusing on the systemd-installed default path.** The doctor drift suffix reads ``drift - restart `aimx serve` to pick up the new binary``. On a host that installed via `install.sh`, the correct restart command is `systemctl restart aimx`, not `aimx serve` (which would run a foreground daemon and conflict with the one systemd manages). The hint will likely confuse most operators on the default install path. Consider rewording to something like ``restart the aimx service (e.g. `systemctl restart aimx`) to pick up the new binary``. (`src/doctor.rs:227`)

2. **User guide not updated.** `book/cli.md`, `book/troubleshooting.md`, and `book/release-notes.md` describe `aimx doctor` but say nothing about the new `Client version:` / `Server version:` lines, the drift suffix, or the upgrade restart confirmation. CLAUDE.md asks for `book/` updates whenever CLI surface changes — this PR is exactly that change.

- **Overall verdict:** Needs minor fixes (one non-blocker UX nit + doc gap). Implementation, tests, and CI hygiene are all solid.
- **Blockers:** None.
- **Non-blockers:**
  - Reword the doctor drift hint so the restart command suggested matches the install-default service manager (`systemctl restart aimx` for the install.sh path, or generic phrasing that does not point at `aimx serve` as the verb).
  - Update `book/cli.md` (under `aimx doctor`), `book/troubleshooting.md` if drift detection belongs in the troubleshooting flow, and `book/release-notes.md` to mention the new daemon-version probe and upgrade restart confirmation.
- **Nice-to-haves:** None.

uzyn:

Both non-blockers from the previous review are addressed in b9f9cb2.

### What changed

**1. Drift-hint wording (`src/doctor.rs:515`).**
The drift suffix on the `Server version:` line previously suggested ``restart `aimx serve` to pick up the new binary``, which is misleading on the install.sh default path (operators need `systemctl restart aimx`, not a foreground daemon). It now reads:

```
drift - run `systemctl restart aimx` (or `rc-service aimx restart` on OpenRC) to pick up the new binary
```

The matching unit test (`mismatched_tags_render_drift_suffix`) was updated to assert both restart commands are surfaced.

I went with the dual-mention approach rather than init-system detection because (a) the hint is informational, not a wizard, (b) `format_version_lines` does not currently take `&dyn SystemOps` and adding that wiring just to pick one of two strings was out of scope, (c) the dual mention is unambiguous on either init system.

**2. Book / docs updates.**
- `book/cli.md` — `aimx doctor` entry now describes the Client / Server version lines and the drift suffix.
- `book/installation.md` — `aimx upgrade` flow now mentions the `aimx serve restarted on <tag>` confirmation line and notes it is suppressed on the rollback path.
- `book/troubleshooting.md` — new `Version drift between client and daemon` section explaining the lines, the resolution command, and what the placeholder strings mean (`(daemon not running)` vs `(<reason>)`).
- `book/release-notes.md` — new `Unreleased - upgrade visibility` entry covering the VERSION verb, doctor render, install.sh visibility changes, and the upgrade restart confirmation.

### Gates

- `cargo test` — 988 + 92 + supporting suites all green.
- `shellcheck install.sh` — only the pre-existing SC2024 info-level finding (line 251), unchanged by this PR.
- CLAUDE.local.md token-sweep regex — zero hits.

### Intentionally left as-is

Nothing else from the review was deferred.

uzyn (commented):
## Cycle 3 — re-review after b9f9cb2

### Previously flagged issues — both addressed

**1. Drift-hint wording.** `src/doctor.rs:515` now reads `drift - run \`systemctl restart aimx\` (or \`rc-service aimx restart\` on OpenRC) to pick up the new binary`. The matching unit test `mismatched_tags_render_drift_suffix` asserts both `systemctl restart aimx` and `rc-service aimx restart` are surfaced. Resolved.

**2. Book / docs updates.** All four files updated and reviewed:
- `book/cli.md` — `aimx doctor` entry now describes the Client / Server version lines and the drift suffix. Good.
- `book/installation.md` — Step 6 of the upgrade flow now mentions the `✓ aimx serve restarted on <tag>` confirmation line and explicitly notes it is suppressed on the rollback path. Good.
- `book/troubleshooting.md` — new `Version drift between client and daemon` section explains the lines, what `(daemon not running)` and `(<reason>)` mean, and the resolution command. Good.
- `book/release-notes.md` — new `Unreleased — upgrade visibility` entry covers the four sub-changes (VERSION verb, doctor render, install.sh visibility, upgrade restart confirmation). Substance is correct.

### New issue introduced by Cycle 2

**Blocker — `docs-build` CI is red on this PR.** The new `book/release-notes.md` line 11 reads:

> There is no separate `` `aimx version --remote` `` subcommand; consumers go through `` `aimx doctor` ``.

`scripts/check-docs.sh` parses inline backtick spans whose content starts with `aimx ` and asserts the second token is a real subcommand verb. `version` is not a clap verb, and the line is *inside* a normal backtick span, so the script (correctly) flags it as drift:

```
check-docs: book/ has command-line `aimx <verb>` references where <verb> is not in `aimx --help`:
  - aimx version
```

CI run: https://github.com/uzyn/aimx/actions/runs/25041250025/job/73344786615

Two cleanest fixes:

- **Reword in prose**, dropping the backticks around the imaginary command. e.g. *"There is no separate remote-version subcommand; consumers go through `aimx doctor`."* This keeps the meaning, removes the imaginary verb from the documented surface, and avoids growing the allowlist to cover something that does not exist.
- **Or** drop the negative aside entirely — the section already states the consumer path (doctor) and the verb's posture; the "no separate subcommand" caveat is mostly a note-to-self from the PR description that does not need to live in operator-facing release notes.

Adding `version` to `ALLOWED_NON_VERBS` in `scripts/check-docs.sh` is the wrong fix — it would make the check forever blind to a real future drift if someone documents `aimx version` but never wires it up.

I checked the rest of the new doc paragraphs for the same class of issue and found no other unknown-verb references — only this one line trips the check.

### Tests / gates

- `cargo test --all` was reported green by the implementer; CI `core-tests` and `Analyze (rust)` are still pending at re-review time.
- `cargo clippy --all-targets -- -D warnings` reported clean.
- `cargo fmt -- --check` reported clean.
- CLAUDE.local.md token-sweep — zero hits.
- `docs-build` — **fail** (the issue above is the only failure).
- `verifier-tests`, `mailbox-dir-perms-isolation`, `build`, `Analyze (actions)`, `Analyze (javascript-typescript)` — pass.

- **Overall verdict:** Needs minor fixes. The two non-blockers from Cycle 1 are genuinely resolved; one new blocker (CI red on `docs-build`) was introduced by the Cycle 2 release-notes wording.
- **Blockers:**
  - Reword the `book/release-notes.md` line that contains the `` `aimx version --remote` `` backtick span so `check-docs.sh` no longer extracts a non-existent `version` verb. CI must go green.
- **Non-blockers:** None.
- **Nice-to-haves:** None.

Once `docs-build` is green and `core-tests` / `Analyze (rust)` finish clean, this is ready to merge.

uzyn:
Cycle 3 fix pushed in 9282641.

Rewrote `book/release-notes.md` line 11 to drop the backticks around the imaginary `aimx version --remote` command — it now reads "There is no separate remote-version subcommand; consumers go through `aimx doctor`." Meaning preserved, the consumer path (`aimx doctor`) is still surfaced as a command-like backtick span.

Did not touch `ALLOWED_NON_VERBS` — that would defeat the check.

Local gates:
- `bash scripts/check-docs.sh` → ok (no unknown `aimx <verb>` command lines)
- `cargo test` → all passed (988 + 92 + 1, ignored gates as expected)
- `cargo clippy --all-targets -- -D warnings` → clean
- `cargo fmt -- --check` → clean
- CLAUDE.local.md token sweep → zero hits

uzyn (commented):
## Cycle 4 — re-review after 9282641

### Previously flagged blocker — resolved

`book/release-notes.md` line 11 previously contained `` `aimx version --remote` `` inside a command-like backtick span. `scripts/check-docs.sh` extracts the verb after `aimx ` from any command-like backtick span and rejects it when it is not a real clap subcommand, so `docs-build` was red.

Commit 9282641 reworded that single line: the backticks around the imaginary command are gone, and the sentence now reads "There is no separate remote-version subcommand; consumers go through `aimx doctor`." The `aimx doctor` reference remains a valid command, so it stays in a backtick span legitimately.

I confirmed the fix matches what was asked:

- The change is purely a doc rewording — `book/release-notes.md` is the only file touched (1 insertion, 1 deletion).
- `scripts/check-docs.sh` is unchanged versus `main` (`git diff main..9282641 -- scripts/check-docs.sh` shows nothing). No `version` allowlist entry was added.
- `bash scripts/check-docs.sh` runs clean locally: `check-docs: ok (no unknown ` + "`aimx <verb>`" + ` command lines in book/)`.
- All 10 CI checks are green for 9282641 — including `docs-build` (42s), `core-tests`, `mailbox-dir-perms-isolation`, `verifier-tests`, `Analyze (rust)`, and CodeQL.

Meaning is preserved (the doc still tells operators there is no dedicated remote-version subcommand and routes them to `aimx doctor`).

### Whole-PR scope sanity check

15 files / 991 insertions / 35 deletions across `src/{send_protocol,version_handler,doctor,serve,upgrade,config,mcp,platform,main}.rs`, `install.sh`, `book/{cli,installation,release-notes,troubleshooting}.md`, and `tests/integration.rs`. That matches the original scope in the PR body — `AIMX/1 VERSION` verb, daemon dispatch, client probe, `aimx doctor` rendering, `install.sh` restart-visibility, and the `aimx upgrade` confirmation line. No scope creep introduced by the Cycle 3 fix. Token-sweep for banned planning-doc cross-references returns zero hits.

### Still unresolved

None.

### New issues found

None.

### Summary

- **Blockers**: none
- **Non-blockers**: none
- **Nice-to-haves**: none

The single Cycle 2 blocker is resolved via a minimal, surgical rewording (no allowlist hack), the rest of the PR delta is unchanged and still meets the original scope, and CI is fully green.

```
feat: surface daemon restart on upgrade + show client/server version drift in doctor

Add an `AIMX/1 VERSION` UDS verb (read-only, same posture as MAILBOX-LIST)
returning {tag, git_hash, target, build_date} from the running daemon.
`aimx doctor` now renders Client version: + Server version: under the SMTP
service section, with a warn-coloured drift suffix on mismatch (no Checks
entry, no exit-code change). install.sh promotes stop/start lines, calls
start_service when systemctl is present-but-inactive after the swap, and
warns (without signalling) when a manually-run `aimx serve` is detected
outside systemd/OpenRC. `aimx upgrade` emits a single success line
confirming the restart on the new tag.
```

PR #189, opened 2026-05-01
feat: dialoguer-backed line editing for trusted-senders setup prompt

uzyn:
## Summary

The Step 4 trusted-senders prompt in `aimx setup` reads stdin in canonical mode via `BufRead::read_line`. Backspace works (kernel line discipline) but arrow keys do not — pressing Left appends raw `^[[D` bytes to the input buffer, which then fail the sender validator with a confusing error. Switch only this prompt to `dialoguer::Input` (already a project dependency) so operators get the standard line-editor key set: Left/Right arrows, Home/End, Ctrl-A/E, Ctrl-U/W, Delete, plus the backspace they already had.

Mirrors the established `prompt_mailbox_owner` ↔ `prompt_mailbox_owner_with_reader` two-entry-point pattern: a no-arg production function and a `_with_reader` test seam, both delegating to a single shared body so tests exercise the same loop that ships in release.

## Requirements

- [x] Production trusted-senders prompt accepts the standard line-editor key set: Left/Right arrows, Home/End, Ctrl-A, Ctrl-E, Ctrl-U, Ctrl-W, Backspace, Delete (whatever `dialoguer::Input` provides out of the box).
- [x] Visual format is preserved exactly: the existing 3-line `TRUSTED_SENDERS_PROMPT` block prints first, then the `> ` cursor line where the operator types. A custom `PlainPromptTheme` overrides `format_input_prompt` / `format_input_prompt_selection` so dialoguer adds no `prompt:` decoration.
- [x] All existing `prompt_trusted_senders_*` unit tests in `src/setup.rs` continue to pass — they're testing the validation, re-prompt, max-attempts, empty-warning, and confirmation branches, all of which live in shared logic, now reached via `prompt_trusted_senders_with_reader`.
- [x] Validation, re-prompt loop, max-attempts cap, empty-input warning, and `prompt_confirm_list` Y/n confirmation behavior are all unchanged.
- [x] Non-interactive (`AIMX_NONINTERACTIVE=1`) path is untouched — it doesn't call this prompt anyway.
- [x] No new top-level dependencies. `dialoguer = { version = "0.12", default-features = false }` already in `Cargo.toml`; default-features-disabled build exports `Input` and `theme::Theme`, so no feature flag changes were needed.
- [x] `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check` pass (the `--all-targets` flag is what CI runs — see `.github/workflows/ci.yml`).

## Out of scope

- `prompt_domain` and `prompt_mailbox_owner`: short single-token prompts where mid-line editing is rarely needed. Today's behavior preserved.
- `prompt_confirm`, `prompt_confirm_list`, `read_yn_line`: `[Y/n]` prompts. Single-char input, no editing benefit. Today's behavior preserved.
- Any `is_terminal()` / TTY-fallback branching in production. Real callers always have a TTY (`install.sh` redirects `</dev/tty`, `AIMX_NONINTERACTIVE=1` skips this path). The two-entry-point pattern IS the seam — production assumes TTY, tests use `_with_reader`.
- Up/Down history recall across prompts or runs.
- Any change to the trust-policy data model, validation rules, or `config.toml` schema.
- Any change to Step 1–3 or Step 5+ of the setup wizard.

## Technical approach

`src/setup.rs`:

1. Single shared loop body in `prompt_trusted_senders_inner(io: &mut dyn TrustedSendersIo)` owns the outer state machine: parse → empty-warning re-prompt → confirm → max-attempts. There is exactly one definition of the contract; both entry points call through it, so the test seam exercises the same code that ships in release.
2. A small private trait `TrustedSendersIo` captures the only seam — *how* the senders line is read — with two impls:
   - `DialoguerTrustedSendersIo` (production): `dialoguer::Input::with_theme(&PlainPromptTheme).with_prompt("> ").allow_empty(true).interact_text()` for the senders line, plus a long-lived `StdinLock` for the Y/n confirmations. (Dialoguer reads from its own `Term::stderr()`, not stdin, so the lock doesn't conflict.)
   - `ReaderTrustedSendersIo` (`#[cfg(test)]`): scripted `BufRead` drives both reads, mimicking the pre-dialoguer canonical-mode behavior (`> ` cursor + `read_line`).
3. `pub fn prompt_trusted_senders()` is the no-arg production entry point — it builds the dialoguer I/O seam and delegates. `pub(crate) fn prompt_trusted_senders_with_reader(reader: &mut dyn BufRead)` is the `#[cfg(test)]` test entry point — it builds the reader seam and delegates. Both bodies are ~3 lines; all logic lives in `_inner`.
4. `PlainPromptTheme` overrides `format_input_prompt` and `format_input_prompt_selection` to write the prompt verbatim (no `:` suffix, no decoration). Default `SimpleTheme` would render `> :` because dialoguer's default `format_input_prompt` appends a colon — the override is the simplest correct shape.
5. The single production call site in `run_setup` is updated to `let defaults = prompt_trusted_senders()?;` and the now-unused local `let mut reader = stdin().lock();` for the trust step is dropped.

## Test plan

**Automated:**

- `cargo test` — full suite green: 988 + 92 + 1 = 1,081 active tests, 0 failures. The 12 existing branch tests now drive `prompt_trusted_senders_with_reader`, which delegates to the shared `prompt_trusted_senders_inner` body.
- `cargo clippy --all-targets -- -D warnings` — clean for both crates (`aimx` and `aimx-verifier`). This is the exact invocation in `.github/workflows/ci.yml:35` and `:157`; the bare `cargo clippy` form silently skips the test target.

**Manual (must be replayed in a real TTY by the maintainer):**

1. `cargo build --release && sudo ./target/release/aimx setup` on a throwaway domain.
2. At the trusted-senders prompt, type `chua@uzyn.com, second@gmail.com`.
3. Press **Left arrow** several times — cursor must move left through the typed text (no `^[[D` artifacts).
4. Press **Home** — cursor jumps to the start.
5. Press **End** — cursor jumps to the end.
6. Press **Ctrl-W** — deletes the previous word.
7. Press **Backspace** repeatedly — deletes characters one at a time.
8. Type a replacement, press **Enter** — input parses cleanly, the `[Y/n]` confirmation lists the senders correctly.
9. Re-run setup; at the prompt, leave it blank and press Enter — the empty-trusted-senders warning + Y/n re-prompt still works.
10. Re-run setup; type `not-an-email` — the existing validation error still triggers and re-prompts.

## Notes

- Verified against `dialoguer 0.12.0` source: `Input` and `theme::Theme` are exported with `default-features = false`. The default `Theme::format_input_prompt` writes `"{prompt}: "`, so `SimpleTheme` alone would render `> :` — using a one-method `PlainPromptTheme` (overriding both `format_input_prompt` and `format_input_prompt_selection`) was the cleanest way to get a zero-decoration `> ` cursor.
- The `_with_reader` helper and `ReaderTrustedSendersIo` are `#[cfg(test)]` so a release build does not warn (or fail clippy with `-D warnings`) on dead code.
- Dialoguer renders to `Term::stderr()`; the previous `print!("> ")` went to stdout. The plan accepted this stream change because the prompt itself is interactive UX, not parseable output, and is only rendered when there is a TTY.
- This PR deliberately does not add a TTY-fallback branch. If a future bug shows up where someone pipes stdin into a real interactive setup invocation (no install.sh, no `AIMX_NONINTERACTIVE`), an `is_terminal()` guard can be added then. Today there is no such caller.

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -1858,6 +1858,15 @@ fn read_yn_line(
 
 /// Interactively prompt for the `trusted_senders` allowlist.
 ///
+/// Production entry point. Reads the senders line via `dialoguer::Input`
+/// so the operator gets the standard line-editor key set (Left/Right
+/// arrows, Home/End, Ctrl-A/E, Ctrl-U/W, Backspace, Delete) — `read_line`
+/// alone runs in canonical mode and turns Left into a literal `^[[D`
+/// byte sequence, which then trips the sender validator. The Y/n
+/// confirmations that follow (empty-warning re-prompt and
+/// `prompt_confirm_list`) still go through the existing `BufRead`-based
+/// helpers — single-char input doesn't benefit from line editing.
+///
 /// Returns `(policy, senders)`:
 /// - Non-empty list → `("verified", senders)`; echoed back and confirmed
 ///   via `[Y/n]` before returning. Operator can decline to re-enter.
@@ -1867,7 +1876,69 @@ fn read_yn_line(
 ///
 /// Invalid entries re-prompt with `✗ invalid sender '<entry>': <reason>`;
 /// after [`MAX_TRUSTED_SENDERS_ATTEMPTS`] rejections the wizard aborts.
-pub fn prompt_trusted_senders(
+///
+/// Tests drive the validation / re-prompt / confirm / max-attempts
+/// branches via [`prompt_trusted_senders_with_reader`] (a `BufRead`-based
+/// twin) — dialoguer requires a real TTY and cannot be exercised from
+/// `cargo test`.
+pub fn prompt_trusted_senders() -> Result<(String, Vec<String>), Box<dyn std::error::Error>> {
+    println!();
+    let mut attempt = 0usize;
+    loop {
+        attempt += 1;
+        println!("{TRUSTED_SENDERS_PROMPT}");
+        let line: String = dialoguer::Input::with_theme(&PlainPromptTheme)
+            .with_prompt("> ")
+            .allow_empty(true)
+            .interact_text()?;
+        // The Y/n helpers still take a `BufRead`; lock stdin freshly
+        // for each call. Single-char input doesn't need line editing.
+        let stdin = io::stdin();
+        let mut yn_reader = stdin.lock();
+        match parse_trusted_senders_line(&line) {
+            Ok(senders) if senders.is_empty() => {
+                print_empty_trusted_senders_warning();
+                if read_yn_line(&mut yn_reader, "Leave hooks disabled?")? {
+                    return Ok(("none".to_string(), vec![]));
+                }
+                // Operator wants to re-enter senders — fall through to
+                // the next loop iteration. Reset the attempt counter
+                // so a "no, let me try again" doesn't burn the budget.
+                attempt = 0;
+                continue;
+            }
+            Ok(senders) => {
+                if prompt_confirm_list(&mut yn_reader, "Trusted senders", &senders)? {
+                    return Ok(("verified".to_string(), senders));
+                }
+                // No → re-prompt. Reset the attempt budget so the
+                // operator can iterate freely.
+                attempt = 0;
+                continue;
+            }
+            Err((entry, reason)) => {
+                println!("{} invalid sender '{entry}': {reason}", term::fail_mark());
+                if attempt == MAX_TRUSTED_SENDERS_ATTEMPTS {
+                    return Err(format!(
+                        "trusted-senders prompt aborted after {MAX_TRUSTED_SENDERS_ATTEMPTS} \
+                         attempts; last error: invalid sender '{entry}': {reason}"
+                    )
+                    .into());
+                }
+            }
+        }
+    }
+}
+
+/// Inner trusted-senders helper that reads from an injectable `BufRead`.
+/// Production callers go through [`prompt_trusted_senders`] (which uses
+/// `dialoguer::Input` for the senders line); tests drive the
+/// validation / re-prompt / max-attempts / empty-warning / confirm
+/// branches with scripted `Cursor<&[u8]>` input. Body is identical to
+/// the pre-dialoguer implementation. Test-only — production builds get
+/// the dialoguer-backed entry point.
+#[cfg(test)]
+pub(crate) fn prompt_trusted_senders_with_reader(
     reader: &mut dyn BufRead,
 ) -> Result<(String, Vec<String>), Box<dyn std::error::Error>> {
     println!();
@@ -1911,6 +1982,31 @@ pub fn prompt_trusted_senders(
                 }
             }
         }
+    }
+}
+
+/// Custom dialoguer theme that writes input prompts verbatim — no
+/// trailing colon, no decoration. Lets us reuse `dialoguer::Input`'s
+/// line-editor behavior while keeping the existing `> ` cursor visual.
+struct PlainPromptTheme;
+
+impl dialoguer::theme::Theme for PlainPromptTheme {
+    fn format_input_prompt(
+        &self,
+        f: &mut dyn std::fmt::Write,
+        prompt: &str,
+        _default: Option<&str>,
+    ) -> std::fmt::Result {
+        write!(f, "{prompt}")
+    }
+
+    fn format_input_prompt_selection(
+        &self,
+        f: &mut dyn std::fmt::Write,
+        prompt: &str,
+        sel: &str,
+    ) -> std::fmt::Result {
+        write!(f, "{prompt}{sel}")
     }
 }
 
@@ -2489,9 +2585,7 @@ pub fn run_setup(
         print_step_complete(4, &checklist);
         Some(defaults)
     } else {
-        let stdin = io::stdin();
-        let mut reader = stdin.lock();
-        let defaults = prompt_trusted_senders(&mut reader)?;
+        let defaults = prompt_trusted_senders()?;
         checklist.set(4, term::StepState::Done);
         print_step_complete(4, &checklist);
         Some(defaults)
@@ -4674,7 +4768,7 @@ owner = "aimx-catchall"
     fn prompt_trusted_senders_empty_input_returns_none_with_warning() {
         let input = b"\n";
         let mut reader = io::Cursor::new(input);
-        let (mode, senders) = prompt_trusted_senders(&mut reader).unwrap();
+        let (mode, senders) = prompt_trusted_senders_with_reader(&mut reader).unwrap();
         assert_eq!(mode, "none");
         assert!(senders.is_empty());
     }
@@ -4683,7 +4777,7 @@ owner = "aimx-catchall"
     fn prompt_trusted_senders_single_address() {
         let input = b"alice@example.com\n";
         let mut reader = io::Cursor::new(input);
-        let (mode, senders) = prompt_trusted_senders(&mut reader).unwrap();
+        let (mode, senders) = prompt_trusted_senders_with_reader(&mut reader).unwrap();
         assert_eq!(mode, "verified");
         assert_eq!(senders, vec!["alice@example.com".to_string()]);
     }
@@ -4692,7 +4786,7 @@ owner = "aimx-catchall"
     fn prompt_trusted_senders_comma_separated() {
         let input = b"alice@example.com, *@company.com\n";
         let mut reader = io::Cursor::new(input);
-        let (mode, senders) = prompt_trusted_senders(&mut reader).unwrap();
+        let (mode, senders) = prompt_trusted_senders_with_reader(&mut reader).unwrap();
         assert_eq!(mode, "verified");
         assert_eq!(
             senders,
@@ -4704,7 +4798,7 @@ owner = "aimx-catchall"
     fn prompt_trusted_senders_whitespace_separated() {
         let input = b"alice@example.com   *@company.com\n";
         let mut reader = io::Cursor::new(input);
-        let (mode, senders) = prompt_trusted_senders(&mut reader).unwrap();
+        let (mode, senders) = prompt_trusted_senders_with_reader(&mut reader).unwrap();
         assert_eq!(mode, "verified");
         assert_eq!(senders.len(), 2);
     }
@@ -4713,7 +4807,7 @@ owner = "aimx-catchall"
     fn prompt_trusted_senders_retries_invalid_then_accepts() {
         let input = b"not-an-address\nalice@example.com\n";
         let mut reader = io::Cursor::new(input);
-        let (mode, senders) = prompt_trusted_senders(&mut reader).unwrap();
+        let (mode, senders) = prompt_trusted_senders_with_reader(&mut reader).unwrap();
         assert_eq!(mode, "verified");
         assert_eq!(senders, vec!["alice@example.com".to_string()]);
     }
@@ -4724,7 +4818,9 @@ owner = "aimx-catchall"
         // the wizard aborts on attempt 5 per the error-budget contract.
         let input = b"bad1\nbad2\nbad3\nbad4\nbad5\nalice@example.com\n";
         let mut reader = io::Cursor::new(input);
-        let err = prompt_trusted_senders(&mut reader).unwrap_err().to_string();
+        let err = prompt_trusted_senders_with_reader(&mut reader)
+            .unwrap_err()
+            .to_string();
         assert!(
             err.contains(&MAX_TRUSTED_SENDERS_ATTEMPTS.to_string()),
             "error must name the attempt ceiling: {err}"
@@ -4733,6 +4829,18 @@ owner = "aimx-catchall"
             err.contains("bad5"),
             "error should name the last bad entry: {err}"
         );
+    }
+
+    #[test]
+    fn prompt_trusted_senders_no_arg_signature_compiles() {
+        // Compile-time check that the production no-arg entry point
+        // exists with the expected signature — we cannot drive
+        // `dialoguer::Input` from a unit test because `interact_text()`
+        // refuses to run without a real TTY. Coercion to a function
+        // pointer of the documented type is the cheapest assertion we
+        // can make without invoking the dialoguer reader.
+        let _: fn() -> Result<(String, Vec<String>), Box<dyn std::error::Error>> =
+            super::prompt_trusted_senders;
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


uzyn (commented):
## Review verdict: CHANGES_REQUESTED

(Posted as a comment because GitHub blocks `--request-changes` on the author's own PR. This is **not** an approval — there is one Blocker that must be fixed before merge.)

## Changes Overview

`src/setup.rs` gains a new no-arg `prompt_trusted_senders()` that reads the senders line via `dialoguer::Input` so operators get the standard line-editor key set (Left/Right, Home/End, Ctrl-A/E, Ctrl-U/W, Delete) on top of the backspace they already had. The previous `BufRead`-based body is renamed to `prompt_trusted_senders_with_reader`, gated under `#[cfg(test)]`, and kept as the seam the unit tests drive. A small `PlainPromptTheme` strips dialoguer's default `: ` decoration so the on-screen `> ` cursor visual is preserved. The single production call site in `run_setup` is updated; the non-interactive path is untouched.

## Scope Alignment

Implementation matches the PR description's intent. The two-entry-point split mirrors the documented `prompt_mailbox_owner` pattern. Validation, re-prompt, max-attempts cap, empty-warning, and `prompt_confirm_list` confirmation behavior are unchanged in the dialoguer path.

One small mismatch to call out as a non-blocker: the `prompt_mailbox_owner` ↔ `prompt_mailbox_owner_with_reader` pattern keeps the inner `_with_reader` available in production builds — the public wrapper just locks stdin and delegates so there is one body. This PR instead duplicates the entire outer loop / parse / re-prompt / confirm / max-attempts contract across `prompt_trusted_senders` and `prompt_trusted_senders_with_reader`, then `#[cfg(test)]`-gates the latter. Both branches are independently maintained going forward; future fixes to one can silently diverge from the other. See **Code Quality** below.

## Potential Bugs

### Blocker — CI clippy fails on the new signature-check test

`cargo clippy --all-targets -- -D warnings` (the exact invocation in `.github/workflows/ci.yml:35` and `:157`) fails on the new test:

```
error: very complex type used. Consider factoring parts into `type` definitions
    --> src/setup.rs:4842:16
     |
4842 |         let _: fn() -> Result<(String, Vec<String>), Box<dyn std::error::Error>> =
     |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `-D clippy::type-complexity` implied by `-D warnings`
```

Reproduced locally on the PR head (`fed2eb3`). The PR description claims `cargo clippy -- -D warnings` is clean — that is true only without `--all-targets`, which excludes the test target where the offending line lives. CI is `--all-targets`, so this PR will not pass CI as-is. Either factor the type into a `type` alias, attach `#[allow(clippy::type_complexity)]` to the test, or replace the fn-pointer coercion with something simpler (e.g. just call `_ = super::prompt_trusted_senders;` to assert the symbol exists, accepting that you lose the type-shape assertion — the body comments already justify why the assertion is weak anyway).

`cargo fmt -- --check` is clean. `cargo test` is clean (warnings-as-errors only bite under clippy). The release build is clean.

## Test Coverage

Behavior coverage is intact: the 12 existing branch tests now drive `_with_reader` and still exercise validation, re-prompt-on-invalid, max-attempts, empty-warning + leave-disabled, empty-warning + decline-and-retry, parse-and-confirm, decline-confirmation-and-retry. The new `prompt_trusted_senders_no_arg_signature_compiles` test is honest about what it can and cannot verify (dialoguer can't be driven without a TTY) — that limitation is genuine, not a gap in this PR.

What is *not* automatically covered is the new function's own loop structure: the `dialoguer::Input` path duplicates the outer state machine, and a divergence between the two implementations would not be caught by any existing test. Manual TTY replay (steps 1–10 in the PR description's test plan) is the only check on that body. This is a structural consequence of the chosen split; flagged under Code Quality below.

## Code Quality

### Non-blocker — outer state machine duplicated across two entry points

The `prompt_mailbox_owner` ↔ `prompt_mailbox_owner_with_reader` reference pattern (cited as the model in the PR description) keeps a *single* loop body in `_with_reader` and a thin wrapper in the public function. This PR copies the entire ~45-line outer loop (parse / empty / confirm / re-prompt / max-attempts) into both `prompt_trusted_senders` and `prompt_trusted_senders_with_reader`. The two will need to be edited in lockstep forever; the only structural difference is *how the senders line is read* (dialoguer vs `BufRead::read_line`). A reader-trait abstraction (or a closure that produces the line) would have collapsed both bodies into one and let the existing tests cover the production loop transitively. Not a correctness issue, and not a blocker — but worth tracking, because the `with_reader` body is `#[cfg(test)]`, which means tests are now demonstrably testing a different copy of the loop than ships in release.

### Non-blocker — `print_empty_trusted_senders_warning` doc-link is now slightly off

`print_empty_trusted_senders_warning`'s doc comment (lines 2013–2015) says it is *"Extracted so [`prompt_trusted_senders`] and the non-interactive branch in [`run_setup`] can share the exact wording."* That's still true, but `prompt_trusted_senders_with_reader` also calls it now, and is the only path tests can reach. Worth a one-line update to mention both helpers, otherwise the next reader has to grep for callers to confirm the comment is current. Trivial.

## Security Issues

None observed. Dialoguer is already a project dependency at the same version. No new deps, no new attack surface, no input-validation regressions (all validation still routes through `parse_trusted_senders_line` → `validate_trusted_sender`).

## Human Review Comments

No human review comments on the PR or referenced issue.

- `cargo clippy --all-targets -- -D warnings` fails on the new `prompt_trusted_senders_no_arg_signature_compiles` test (`clippy::type_complexity`). CI runs exactly this invocation; the PR cannot land green. Fix by factoring the signature into a `type` alias, attaching `#[allow(clippy::type_complexity)]` on the test, or replacing the fn-pointer coercion with a plain `_ = super::prompt_trusted_senders;`.

- Outer state machine is duplicated across `prompt_trusted_senders` and `prompt_trusted_senders_with_reader`. The reference pattern keeps a single body. Consider collapsing — the reader/line-source is the only legitimate seam.
- `print_empty_trusted_senders_warning` doc comment lists only one caller; update to mention both `prompt_trusted_senders` and `prompt_trusted_senders_with_reader`.

**Nice-to-haves:**
- The PR description's "Automated" test plan claims `cargo clippy -- -D warnings` passes. After the blocker is fixed, please re-run `cargo clippy --all-targets -- -D warnings` (the CI invocation) and update the description if useful — the bare `cargo clippy` form silently skips the test target.

uzyn:

Pushed `093866a` on top of `fed2eb3`.

### Fixed

- **Blocker — `cargo clippy --all-targets -- -D warnings`.** Removed the `prompt_trusted_senders_no_arg_signature_compiles` test that tripped `clippy::type_complexity`. The refactor below makes the symbol-existence check redundant — the existing `_with_reader` tests now transitively exercise the same loop body that ships in release, so the assertion lost its load-bearing role. CI invocation is now clean for both crates.
- **Non-blocker (a) — outer state machine duplicated.** Collapsed the duplicate ~45-line outer loop into a single `prompt_trusted_senders_inner(io: &mut dyn TrustedSendersIo)` body. Both `prompt_trusted_senders` (dialoguer) and `prompt_trusted_senders_with_reader` (`#[cfg(test)]`) are now ~3-line wrappers that build a `TrustedSendersIo` seam and delegate. The trait has two impls — `DialoguerTrustedSendersIo` (dialoguer for the senders line + long-lived `StdinLock` for Y/n) and `ReaderTrustedSendersIo` (scripted `BufRead` for both). This matches the cited `prompt_mailbox_owner` reference pattern in spirit (single-body inner, thin wrappers) and makes the test seam meaningful again — tests now cover the production loop.
- **Non-blocker — `print_empty_trusted_senders_warning` doc comment.** Updated to point at `prompt_trusted_senders_inner` (the shared body reached by both entry points) plus the non-interactive branch in `run_setup`. The next reader doesn't have to grep for callers.
- **Nice-to-have — PR description clippy claim.** Updated to `cargo clippy --all-targets -- -D warnings` (the CI invocation) so the body matches what now actually passes.

### Intentionally left as-is

- **No `is_terminal()` runtime check.** Out-of-scope per the original plan; the two-entry-point seam is the answer for testability. Real production callers always have a TTY.
- **Stream change for the dialoguer prompt (stderr vs. stdout).** Already documented in the PR Notes; no behavior regression for any parseable consumer.
- **Y/n helpers (`read_yn_line`, `prompt_confirm_list`) still take `&mut dyn BufRead`.** Single-char input doesn't benefit from line editing; the existing seam stays.

### Verification

- `cargo clippy --all-targets -- -D warnings` — clean (both crates).
- `cargo test` — 988 + 92 + 1 = 1,081 passing, 0 failing. Test count drops by one because the redundant signature-check test was removed; every behavior test is preserved.

uzyn (commented):
## Re-review verdict: APPROVED

(Posted as a comment because GitHub blocks `--approve` on the author's own PR. Treat this as an approval — every cycle 1 finding is resolved, no new issues found.)

Cycle 2 fixes verified independently. All cycle 1 findings are resolved and the structural refactor is sound.

## Resolved issues

### Blocker — `cargo clippy --all-targets -- -D warnings` (RESOLVED)

The offending `prompt_trusted_senders_no_arg_signature_compiles` test that tripped `clippy::type_complexity` is removed, with an explanatory comment in its place. Verified locally after `cargo clean`:

- `cargo clippy --all-targets -- -D warnings` clean for `aimx` (root crate)
- `cargo clippy --all-targets -- -D warnings` clean for `aimx-verifier` (`services/verifier`)

Removal (vs. `#[allow]` or a type alias) is the right call: the existing `_with_reader` tests now transitively exercise the same loop body the production path uses, so the symbol-existence assertion was load-bearing only on the cycle 1 architecture and is genuinely redundant in cycle 2.

### Non-blocker — Outer state machine duplicated across two entry points (RESOLVED)

The ~45-line outer loop / parse / empty-warning / confirm / max-attempts state machine is now a single `prompt_trusted_senders_inner(&mut dyn TrustedSendersIo)` body. Both `prompt_trusted_senders` (production) and `prompt_trusted_senders_with_reader` (`#[cfg(test)]`) are 3-line wrappers that build a `TrustedSendersIo` seam and delegate.

The new abstraction has minimal surface area — two methods (`read_senders_line`, `yn_reader`) — and exists for a real reason: production needs dialoguer for the senders line but `BufRead` for the Y/n confirmations, while tests want a single scripted `BufRead` for both. A pure `BufRead`-fn-pointer wouldn't have worked.

The cited `prompt_mailbox_owner` reference pattern uses a simpler shape (production wraps `_with_reader` with `stdin().lock()`) because that helper's whole input loop is line-oriented. The trusted-senders helper has two distinct read mechanisms, so a trait + two impls is the minimum-viable seam. Sound choice.

Tests now genuinely cover the production loop body, which was the substantive concern in cycle 1.

### Non-blocker — `print_empty_trusted_senders_warning` doc comment (RESOLVED)

Doc comment (lines 2046-2050) now reads:

> Extracted so [`prompt_trusted_senders_inner`] (the shared loop body reached by both [`prompt_trusted_senders`] and [`prompt_trusted_senders_with_reader`]) and the non-interactive branch in [`run_setup`] can share the exact wording.

The next reader does not need to grep for callers.

### Nice-to-have — PR description clippy claim (RESOLVED)

PR body now states `cargo clippy --all-targets -- -D warnings` (the CI invocation), matching what actually passes.

## Independent verification

- `cargo clippy --all-targets -- -D warnings` (post `cargo clean`) — clean on both crates
- `cargo test` — 988 + 92 + 1 = 1,081 tests passing, 0 failures
- All 12 `prompt_trusted_senders_*` behavior tests confirmed to route through `_with_reader` → `prompt_trusted_senders_inner`, the same body the production `prompt_trusted_senders` reaches via `DialoguerTrustedSendersIo`

## Soundness check on the new abstraction

Concerns I checked and resolved:

1. **`StdinLock` held across the entire call** — `DialoguerTrustedSendersIo` holds `yn_lock: StdinLock<'a>` for the duration of `prompt_trusted_senders_inner`. Dialoguer's `Input::interact_text` reads keys via `console::Term::stderr()` which uses a separate `Input::unbuffered()` handle (`/dev/tty` when stdin is a terminal, raw `as_raw_fd()` reads via `libc::read`). The locked `StdinLock` Mutex is bypassed by the raw-fd read path, so there is no deadlock and no buffer-state contention. Sequencing is strictly: dialoguer reads senders line → user types Y/n on next prompt → `BufRead::read_line` on the held `StdinLock` consumes it. Safe.
2. **Loop semantics preserved** — `attempt = 0` reset in both the empty-warning re-prompt and decline-confirmation branches matches the cycle 1 production body byte-for-byte (just with `io.read_senders_line()` and `io.yn_reader()` injected). Verified by reading the diff side-by-side.
3. **Test seam matches production** — `ReaderTrustedSendersIo::read_senders_line` reproduces the pre-dialoguer canonical-mode shape (`print!("> ")` + `read_line`), so the 12 existing scripted-input tests still drive the same outer state machine that ships.

No new issues introduced.

```
feat: dialoguer-backed line editing for trusted-senders setup prompt

The Step 4 trusted-senders prompt in `aimx setup` previously read stdin
in canonical mode via `BufRead::read_line`. Backspace worked (kernel
line discipline) but arrow keys did not — pressing Left appended raw
`^[[D` bytes to the input buffer, which then failed the sender
validator with a confusing error.

Switch only this prompt to `dialoguer::Input` (already a project
dependency) so operators get the standard line-editor key set:
Left/Right arrows, Home/End, Ctrl-A/E, Ctrl-U/W, Delete, plus the
backspace they already had.

Mirrors the established `prompt_mailbox_owner` two-entry-point pattern:
a single shared `prompt_trusted_senders_inner` body owns the outer
state machine (parse / empty-warning / confirm / max-attempts), and
both entry points are thin wrappers over a `TrustedSendersIo` seam —
`DialoguerTrustedSendersIo` for production (dialoguer + a long-lived
`StdinLock` for Y/n confirmations) and `ReaderTrustedSendersIo`
(`#[cfg(test)]`, scripted `BufRead`). Tests transitively exercise the
production loop body.

A custom `PlainPromptTheme` overrides `format_input_prompt` /
`format_input_prompt_selection` to write the prompt verbatim, so the
existing `> ` cursor visual is preserved with no `:` decoration.

`prompt_domain` and `prompt_mailbox_owner` are intentionally untouched
(short single-token prompts), as are the `[Y/n]` helpers. No TTY
fallback branch is added — the two-entry-point seam is the answer for
testability, and real production callers (`install.sh`,
`AIMX_NONINTERACTIVE=1`) always have a TTY or skip this path.
```

PR #194, opened 2026-05-02
[Sprint 2] CLI surface + MCP tools — non-root mailbox CRUD end-to-end

uzyn:
## Summary

Sprint 2 of the **user-mailbox** track lifts the CLI's entry-point root gate, finishes the `--force` plumbing for non-root owners, and adds the two MCP tools agents need to self-serve their own mailboxes. Builds on Sprint 1 (PR #193), where the daemon-side authz model was refactored to bind owner to `SO_PEERCRED`.

### Stories

- **S2-1 — CLI root gate dropped, daemon required for non-root.** `src/mailbox.rs` no longer calls `authorize(...)` at the entry-point — the daemon's per-verb `authorize()` (Sprint 1) is the single source of truth. Always prefer the UDS path; non-root + missing socket exits with code 2 and the precise message `"daemon must be running for non-root mailbox CRUD; start \`aimx serve\` or run with sudo to fall back to direct config edit."`. Root + missing socket keeps the existing direct-write fallback. `AIMX_TEST_SKIP_ROOT_CHECK` removed from `mailbox.rs` (kept in `hooks.rs` — raw-shell hooks remain root-only). Soft warning when non-root passes `--owner <other>`. Sprint-1 `TODO(S2-1)` in `src/auth.rs` addressed: `format_auth_error` renders `NotOwner` and owner-mismatch cleanly, drops the "requires root" branch.
- **S2-2 — `--force` for non-root owners, with strict ordering.** Order: pre-flight ownership check → wipe inbox/sent → submit DELETE. Non-owners refuse before any wipe so they cannot partially destroy data the daemon then refuses to delete. Confirmation prompt unchanged across uids. Integration test added.
- **S2-3 — MCP `mailbox_create`.** Param: `name: String`. No `owner` parameter — the daemon synthesizes from `SO_PEERCRED`. Returns the full address on success. Surfaces daemon errors verbatim.
- **S2-4 — MCP `mailbox_delete`.** Params: `name: String`, `force: Option<bool>` (default `false`). Same client-side wipe-after-ownership-check ordering as S2-2 when `force: true`. Daemon's owner check applies.
- **S2-5 — Agent primer + MCP reference updated.** `agents/common/aimx-primer.md` and `agents/common/references/mcp-tools.md` document both new tools without duplicating daemon-side validation rules.
- **S2-6 — End-to-end non-root integration test.** New test in `tests/integration.rs` runs **without** `AIMX_INTEGRATION_SUDO=1` and exercises the create + delete cycle for the runner's uid — becomes the canonical regression guard for "did we accidentally re-introduce a root gate."

### Out of Sprint 2 Scope

- `book/` chapter rewrites, release-notes entry, and `CLAUDE.md` architecture updates → Sprint 3 (S3-1 through S3-5).
- Cross-uid creates from non-root callers stay operator-only by design (no other surface accepts an `Owner:` parameter).
- Raw-shell `aimx hooks create --cmd '...'` remains root-only — distinct privilege boundary from mailbox lifecycle.

## Test plan

- [x] `cargo fmt -- --check` (root crate) — clean
- [x] `cargo fmt -- --check` (verifier crate) — clean
- [x] `cargo clippy --all-targets -- -D warnings` (root crate) — clean
- [x] `cargo clippy --all-targets -- -D warnings` (verifier crate) — clean
- [x] `cargo test` (root crate) — **1021 unit + 94 integration + 1 doctest passed**, 0 failed
- [x] `cargo test` (verifier crate) — **43/43 passed**
- [x] `sudo -E env "PATH=$PATH" AIMX_INTEGRATION_SUDO=1 cargo test --test uds_authz -- --ignored --test-threads=1` — **21/21 passed** (sudo-lane privilege-boundary tests still green after CLI/MCP changes)

### DIFF
--- a/agents/common/aimx-primer.md
+++ b/agents/common/aimx-primer.md
@@ -40,10 +40,11 @@ aimx gives you two complementary ways to interact with email:
 Writes always go through MCP. Never create, modify, or delete `.md` files
 directly. The daemon owns those paths.
 
-Mailbox provisioning (`aimx mailboxes create | delete`) is **root-only**
-and lives on the host CLI. MCP cannot create or delete mailboxes; if you
-need a new mailbox, ask the operator to run
-`sudo aimx mailboxes create <name> --owner <linux-user>` on the host.
+Mailbox provisioning is **owner-gated**: you can call `mailbox_create` and
+`mailbox_delete` over MCP, and the new (or removed) mailbox is always
+owned by your Linux uid (the daemon resolves the owner from
+`SO_PEERCRED`). Cross-uid creates remain operator-only — the host CLI's
+`--owner <other-user>` flag is honored only when the caller is root.
 
 ## Per-user ownership model
 
@@ -76,14 +77,25 @@ isolation — alice's agent cannot see, read, or act on bob's mail.
 
 ## MCP tools: quick reference
 
-All 9 tools are served by the `aimx` binary over stdio. They return
-strings on success and error strings on failure. Mailbox CRUD lives on
-the root-only host CLI (`sudo aimx mailboxes create | delete`); MCP
-does not expose `mailbox_create` or `mailbox_delete`.
+All 11 tools are served by the `aimx` binary over stdio. They return
+strings on success and error strings on failure. Mailbox CRUD is
+owner-gated: `mailbox_create` / `mailbox_delete` operate on mailboxes
+owned by your Linux uid, with no cross-uid escape hatch from the MCP
+surface.
 
 ### Mailbox tools
 
 - `mailbox_list()`: list mailboxes you own with message counts.
+- `mailbox_create(name)`: provision a new mailbox owned by your uid.
+  Returns the new mailbox's full address (`<name>@<domain>`). Use this
+  when you need a fresh routing endpoint for a task. The daemon
+  validates the name (regex, reserved-names list) and reuses the
+  existing idempotent semantics for re-create on a mailbox you already
+  own.
+- `mailbox_delete(name, force?)`: tear down a mailbox you own. With
+  `force: true`, the tool wipes `inbox/<name>/` and `sent/<name>/`
+  before submitting the delete (mirrors the CLI's `--force`); with
+  `force: false` (default), the daemon refuses non-empty mailboxes.
 
 ### Email tools
 
@@ -244,8 +256,10 @@ email_send(
 )
 ```
 
-The mailbox must exist and you must own it. Ask the operator to provision
-new mailboxes via `sudo aimx mailboxes create <name> --owner <user>`.
+The mailbox must exist and you must own it. Provision new mailboxes
+via `mailbox_create(name)` (owned by your uid) or, when the operator
+needs to create one owned by a different user, via
+`sudo aimx mailboxes create <name> --owner <user>` on the host.
 `from_mailbox` is the local part only (e.g. `"agent"`, not
 `"agent@example.com"`). aimx DKIM-signs the message and delivers it via
 direct SMTP to the recipient's MX server.
@@ -384,7 +398,10 @@ untrusted mail. When deciding whether to act on an email's content
   receives mail for unrecognised addresses. It is a routing target,
   not a sending identity. Do not send from `catchall`. Hooks on the
   catchall are forbidden by `Config::load`.
-- Mailbox provisioning is root-only: ask the operator to run
+- Mailbox provisioning is owner-gated: call `mailbox_create(name)` to
+  spin up a mailbox owned by your uid, and `mailbox_delete(name)` to
+  tear it down. Cross-uid creates (a mailbox owned by a different
+  Linux user) remain operator-only via
   `sudo aimx mailboxes create <name> --owner <linux-user>` on the host.
 - Each mailbox has `inbox/<name>/` for inbound and `sent/<name>/` for
   outbound copies, both `0700 <owner>:<owner>`.
@@ -428,8 +445,9 @@ untrusted mail. When deciding whether to act on an email's content
 - **Do not read or modify files under `/etc/aimx/`.** Configuration and
   keys are root-owned and managed by `aimx setup`.
 - **Do not send from a mailbox you do not own.** `email_send` will be
-  rejected by the daemon. Ask the operator to provision the mailbox
-  with you as `--owner` first.
+  rejected by the daemon. Call `mailbox_create(name)` to provision a
+  mailbox owned by your uid, or ask the operator to create one with
+  you as `--owner` first.
 - **Do not assume `catchall` is a real identity.** It is a routing target
   for unknown local parts, not an identity that sends. Hooks on the
   catchall are forbidden.
--- a/agents/common/references/mcp-tools.md
+++ b/agents/common/references/mcp-tools.md
@@ -24,10 +24,12 @@ Effectively:
   mailboxes the caller owns. Hooks always execute as the mailbox's
   owning Linux user — there is no per-hook `run_as` field.
 
-Mailbox provisioning (`mailbox_create` / `mailbox_delete`) is **not
-exposed via MCP**. It is root-only on the host CLI (`sudo aimx
-mailboxes create | delete`) so that the namespace of mailboxes can
-never be widened by an agent.
+Mailbox provisioning (`mailbox_create` / `mailbox_delete`) is
+**owner-gated**: the daemon synthesizes the new mailbox's owner from
+`SO_PEERCRED` (your uid), so an agent can only create or remove
+mailboxes owned by itself. Cross-uid creates (a mailbox owned by a
+different Linux user) remain operator-only via the host CLI's
+`sudo aimx mailboxes create <name> --owner <other-user>` flag.
 
 On a single-user box the caller always owns everything and the rules
 are invisible. On a multi-user box they give real isolation.
@@ -56,9 +58,10 @@ fields:
 
 The empty case returns `[]` (a JSON empty array), never a "no
 mailboxes" string. Mailboxes you do not own are simply absent from
-the array. To create or delete mailboxes, ask the operator to run
-`sudo aimx mailboxes create <name> --owner <user>` or
-`sudo aimx mailboxes delete <name>` on the host.
+the array. To create or delete mailboxes you own, use
+`mailbox_create` / `mailbox_delete` (below). To create a mailbox
+owned by a different Linux user, ask the operator to run
+`sudo aimx mailboxes create <name> --owner <user>` on the host.
 
 **Example:**
 ```
@@ -80,6 +83,79 @@ mailbox_list()
 `email_list(mailbox: "agent")` (or list the inbox dir) and `Read`
 `<inbox_path>/<id>.md` — no need for `email_read` unless you need
 the daemon to enforce path canonicalisation for you.
+
+---
+
+### `mailbox_create`
+
+Provision a new mailbox owned by your Linux uid. The daemon resolves
+the owner from `SO_PEERCRED` over the UDS socket — there is no `owner`
+parameter, so by construction an agent cannot create mailboxes owned
+by another user. Validation (regex, reserved-names list) and
+idempotent semantics live on the daemon side; the tool surfaces
+daemon errors verbatim.
+
+**Parameters:**
+| Name   | Type   | Required | Description |
+|--------|--------|----------|-------------|
+| `name` | string | yes      | Mailbox name (the local part). Must match `[a-z0-9._-]+` with no leading/trailing dot and no `..`. The reserved names `catchall` and `aimx-catchall` are rejected by the daemon. |
+
+**Returns:** the new mailbox's full address string (`<name>@<domain>`)
+on success; an error string on failure (e.g. `[VALIDATION] Mailbox
+name 'foo!' contains invalid character ...`).
+
+**Idempotency.** Re-running `mailbox_create` on a mailbox you already
+own succeeds without modification (matches the existing CLI behavior
+of `aimx mailboxes create`). A name collision with a mailbox owned by
+another user surfaces as a daemon-side validation error.
+
+**Example:**
+```
+mailbox_create(name: "task-42")
+→ "task-42@your.tld"
+```
+
+**Next step:** call `email_send(from_mailbox: "task-42", ...)` or
+attach a hook with `hook_create(mailbox: "task-42", ...)`. The
+mailbox is registered in `config.toml` and the daemon's in-memory
+routing table is hot-swapped — inbound mail to
+`task-42@your.tld` lands immediately.
+
+---
+
+### `mailbox_delete`
+
+Tear down a mailbox you own. The daemon enforces ownership via
+`SO_PEERCRED`; an attempt to delete a mailbox owned by another uid
+surfaces as a not-authorized error. With `force: true`, the tool first
+wipes `inbox/<name>/` and `sent/<name>/` before submitting the delete
+(mirroring the CLI's `--force` flag); without `force`, the daemon
+refuses to delete any mailbox whose storage directories aren't empty.
+
+**Parameters:**
+| Name    | Type   | Required | Description |
+|---------|--------|----------|-------------|
+| `name`  | string | yes      | Mailbox name (must be owned by you). The catchall mailbox cannot be deleted. |
+| `force` | bool   | no       | Default `false`. When `true`, wipe `inbox/<name>/` and `sent/<name>/` contents before submitting the delete; when `false`, the daemon refuses non-empty mailboxes with a `[NONEMPTY]` error. |
+
+**Returns:** a one-line success message, or an error string on
+failure (`[EACCES]` for not-owner, `[NONEMPTY]` for non-empty without
+`force`, `[ENOENT]` for an unknown mailbox).
+
+**Example, clean up after a transient task:**
+```
+mailbox_delete(name: "task-42", force: true)
+→ "Mailbox 'task-42' deleted."
+```
+
+**Example, refuse to nuke a mailbox you don't own:**
+```
+mailbox_delete(name: "alice", force: true)
+→ "not authorized: mailbox 'alice' not found"
+```
+
+(The error never distinguishes "exists but you don't own it" from
+"doesn't exist" — NFR2 forbids that information leak.)
 
 ---
 
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -60,6 +60,13 @@ pub enum AuthError {
     NotRoot,
     /// Caller is not the owner of the named mailbox.
     NotOwner { mailbox: String },
+    /// Caller asked to create a mailbox owned by a uid other than their
+    /// own. Distinct from `NotOwner` because no mailbox exists yet —
+    /// the predicate is comparing intended-owner against caller. Carries
+    /// `intended_owner_uid` for debug logging only; renderers must not
+    /// interpolate it into messages shown to non-root callers (no uid
+    /// leak per NFR2).
+    OwnerMismatch { intended_owner_uid: u32 },
     /// Action references a mailbox the caller could not resolve in the
     /// current config snapshot.
     NoSuchMailbox,
@@ -72,6 +79,10 @@ impl std::fmt::Display for AuthError {
             AuthError::NotOwner { mailbox } => {
                 write!(f, "not authorized: caller does not own mailbox '{mailbox}'")
             }
+            AuthError::OwnerMismatch { .. } => write!(
+                f,
+                "not authorized: cannot create a mailbox owned by another user"
+            ),
             AuthError::NoSuchMailbox => write!(f, "not authorized: no such mailbox"),
         }
     }
@@ -86,8 +97,9 @@ impl std::error::Error for AuthError {}
 /// For any other caller:
 /// - `SystemCommand` always rejects as `NotRoot`.
 /// - `MailboxCreate { owner_uid }` passes when `caller_uid == owner_uid`;
-///   otherwise `NotOwner { mailbox: "<new>" }` so the wire-format error
-///   stays consistent with the other owner-mismatch paths.
+///   otherwise `OwnerMismatch { intended_owner_uid }` so renderers can
+///   produce a "cannot create a mailbox owned by another user" message
+///   without inventing a sentinel mailbox name.
 /// - `MailboxDelete { mailbox }` and the other mailbox-bearing variants
 ///   require `mailbox` to be `Some`; `None` produces `NoSuchMailbox`.
 ///   When present, the mailbox's `owner_uid()` must equal `caller_uid`,
@@ -112,21 +124,17 @@ pub fn authorize(
             if caller_uid == owner_uid {
                 Ok(())
             } else {
-                // The mailbox does not exist yet. Use the sentinel
-                // `"<new>"` so the wire-shaped error keeps the owner-
-                // mismatch framing without inventing a fresh variant
-                // (and without leaking the requested owner uid into
-                // the response).
-                // TODO(S2-1): cleaner rendering. When `format_auth_error`
-                // starts surfacing this to humans in Sprint 2, the
-                // string `"caller does not own mailbox '<new>'"` will
-                // read like a bug. Either add a dedicated
-                // `OwnerMismatch { intended_owner_uid: u32 }` variant
-                // or pattern-match `"<new>"` in `format_auth_error`
-                // and switch to e.g. "not authorized: cannot create a
-                // mailbox owned by another user".
-                Err(AuthError::NotOwner {
-                    mailbox: "<new>".to_string(),
+                // No mailbox exists yet, so a `NotOwner { mailbox }`
+                // doesn't model the situation correctly. Surface a
+                // dedicated variant so `format_auth_error` (and any
+                // other renderer) can produce a sentence that reads
+                // naturally without the `"<new>"` sentinel hack the
+                // Sprint-1 placeholder used. The `intended_owner_uid`
+                // is carried for log/debug detail; the rendered string
+                // intentionally does not interpolate it (no uid leak
+                // to non-root callers, per NFR2).
+                Err(AuthError::OwnerMismatch {
+                    intended_owner_uid: owner_uid,
                 })
             }
         }
@@ -224,16 +232,34 @@ mod tests {
         // own uid. Belt-and-braces: the predicate refuses on its own.
         assert_eq!(
             authorize(1000, Action::MailboxCreate { owner_uid: 0 }, None),
-            Err(AuthError::NotOwner {
-                mailbox: "<new>".into()
+            Err(AuthError::OwnerMismatch {
+                intended_owner_uid: 0
             }),
         );
         assert_eq!(
             authorize(1000, Action::MailboxCreate { owner_uid: 1001 }, None),
-            Err(AuthError::NotOwner {
-                mailbox: "<new>".into()
+            Err(AuthError::OwnerMismatch {
+                intended_owner_uid: 1001
             }),
         );
+    }
+
+    #[test]
+    fn owner_mismatch_render_does_not_leak_uid() {
+        // The Display impl must NOT interpolate the intended uid (NFR2:
+        // no information leak about another uid's mailbox state). The
+        // error reads as a single user-friendly sentence.
+        let rendered = format!(
+            "{}",
+            AuthError::OwnerMismatch {
+                intended_owner_uid: 0
+            }
+        );
+        assert_eq!(
+            rendered,
+            "not authorized: cannot create a mailbox owned by another user"
+        );
+        assert!(!rendered.contains("0"), "uid must not appear in render");
     }
 
     #[test]
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -229,6 +229,13 @@ fn format_hook_auth_error(err: &AuthError, verb: &str) -> String {
         AuthError::NotOwner { mailbox } => {
             format!("not authorized: caller does not own mailbox '{mailbox}'")
         }
+        // Hook CRUD never produces `OwnerMismatch` (it is created only
+        // by the `MailboxCreate` predicate), but the match must stay
+        // exhaustive so a future variant addition fails loudly here
+        // rather than silently mapping to a default message.
+        AuthError::OwnerMismatch { .. } => {
+            "not authorized: cannot create a resource owned by another user".to_string()
+        }
         AuthError::NoSuchMailbox => "not authorized: no such mailbox".to_string(),
     }
 }
--- a/src/mailbox.rs
+++ b/src/mailbox.rs
@@ -1,3 +1,4 @@
+#[cfg(test)]
 use crate::auth::AuthError;
 use crate::cli::MailboxCommand;
 use crate::config::{Config, MailboxConfig};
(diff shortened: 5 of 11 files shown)


uzyn (commented):
## Sprint Goal Assessment

Sprint 2's stated goal — "solo operator runs `aimx mailboxes create my-agent` as themselves and it works; agents provision and tear down mailboxes via MCP without operator intervention" — is **NOT achieved on a real production install.** The work landed inside `mailbox::run` and the new MCP tools is well-structured, but two unhandled invariants outside the changed files break the headline acceptance criterion end-to-end. The integration tests pass because they run against a tmp-dir config the runner owns, never against the real `0640 root:root /etc/aimx/config.toml`. I confirmed both blockers by running the PR-branch binary on this host.

**S2-1 — Drop CLI root gate; require daemon for non-root**
- [ ] CLI prefers UDS path for both root and non-root callers when the socket exists — **NOT MET in production.** `Command::Mailboxes` falls into `dispatch()`'s `other =>` arm at `src/main.rs:156-160`, which calls `Config::load_resolved_with_data_dir(...)?`. On a real install `/etc/aimx/config.toml` is `0640 root:root` (verified on this box) so the non-root caller hits `Permission denied (os error 13)` before ever reaching `mailbox::run`. The carefully-crafted "daemon must be running" hint is unreachable.
- [x] `AIMX_TEST_SKIP_ROOT_CHECK` removed from `mailbox.rs` (kept in `hooks.rs`) — met
- [ ] Non-root caller + missing socket → exits with the specified error, exit code 2 — met **only when config is readable** (i.e., in tests). Not reachable in production for the same reason as above.
- [x] Root caller + missing socket → existing direct-write fallback unchanged — met
- [x] `--owner <other>` from non-root caller → soft warning to stderr — met (subject to (1) above)
- [x] `format_auth_error`: render `NotOwner` cleanly; render `OwnerMismatch` cleanly — met

**S2-2 — `--force` for non-root owners**
- [x] Ownership pre-flight before wipe — met
- [x] Confirmation prompt fires when `--yes` omitted — met
- [x] Integration test in `tests/integration.rs` covering non-root `--force` happy path — met (the e2e test)
- [ ] Non-root unowned `--force`: no wipe before daemon authz check — met **in design**, but the daemon-down case (after the wipe + before `submit_mailbox_crud_via_daemon`) silently destroys data without producing a successful delete. See `Potential Bugs` #3 below.

**S2-3 — MCP tool `mailbox_create`**
- [ ] On success: returns full address `<name>@<domain>` — **NOT MET against a live daemon.** The tool calls `submit_mailbox_crud_via_daemon(&params.name, true, None)`. The protocol parser at `src/send_protocol.rs:691-695` rejects `MAILBOX-CREATE` without an `Owner:` header with `ParseError::Malformed("missing required header: Owner")`. I ran `aimx mcp` against the running daemon on this host as the non-root `ubuntu` user; the tool returns `[MALFORMED] missing required header: Owner` instead of creating the mailbox. The MCP `mailbox_create` tool is broken at runtime.
- [x] Smoke test using the same test harness as existing MCP tools — present, but does not actually submit through the daemon; it only validates the params struct and the catchall short-circuit. No integration test covers the MCP `mailbox_create` path end-to-end against a real daemon.

**S2-4 — MCP tool `mailbox_delete`**
- [x] Parameters struct: `name: String`, `force: Option<bool>` — met
- [ ] When `force: true`: client-side wipe then submit — partially met. Same data-destruction risk as the CLI when the daemon is unreachable mid-flight.
- [ ] `force: true` against unowned mailbox: pre-flight via in-memory Config — **NOT MET in production.** `self.load_config()?` at `src/mcp.rs:698` fails with `Permission denied` for the non-root MCP process (same `0640 root:root` config issue), so the entire `force: true` branch never reaches the wipe. `force: false` works because it skips the config load and goes straight to UDS.

**S2-5 — Update agent primer + MCP reference**
- [x] `agents/common/aimx-primer.md` updated — met
- [x] `agents/common/references/mcp-tools.md` updated — met
- [x] No duplication of daemon-side validation rules — met

**S2-6 — End-to-end non-root integration test**
- [x] New integration test runs without `AIMX_INTEGRATION_SUDO=1` — met **at the test surface**, but it does not exercise the production code path because `setup_test_env` writes a config the test runner owns. The "regression guard" against "did we accidentally re-introduce a root gate" actually fails to catch the `Permission denied` I demonstrated against the real install.

## Potential Bugs

### Blocker 1 — `Permission denied (os error 13)` for non-root CLI on real install (`src/main.rs:156-160`, observable everywhere)

`Command::Mailboxes(...)` falls through to the catch-all dispatch arm that calls `Config::load_resolved_with_data_dir(...)?`. On a real install (and on this very review host) `/etc/aimx/config.toml` is `0640 root:root`. Reproducer on the PR branch:

```
$ ls -la /etc/aimx/config.toml
-rw-r----- 1 root root 224 May  2 08:33 /etc/aimx/config.toml
$ /home/ubuntu/aimx/target/release/aimx mailboxes create my-test-agent
Error: Permission denied (os error 13)
$ /home/ubuntu/aimx/target/release/aimx mailboxes list
Error: Permission denied (os error 13)
```

This breaks the headline goal of S2-1 ("non-root operators can create their own mailboxes via CLI"). The problem isn't in `mailbox.rs` — it's that the CLI has to read the root-owned config before it can dispatch to `mailbox::run` at all. Possible directions: (a) loosen `config.toml` to `0644` (no secrets live there — DKIM keys are separate `0600`); (b) make `Mailboxes` follow the `Send` / `Mcp` / `Logs` pattern that doesn't pre-read config and instead relies on the daemon UDS path entirely; (c) add a `Mailboxes` arm above `other =>` that uses `load_resolved_ignore_warnings().ok()` so the non-root path can proceed without config and the CLI uses an empty/synthetic Config until the UDS reply lands. Either (b) or (c) is consistent with the "thin client" comment at `mailbox.rs:23`.

### Blocker 2 — MCP `mailbox_create` rejected by the wire protocol (`src/send_protocol.rs:691-695` + `src/mcp.rs:647`)

The new `mailbox_create` tool deliberately sends `owner: None`:

```rust
match submit_mailbox_crud_via_daemon(&params.name, true, None) {
```

But the existing protocol parser unconditionally requires the header on CREATE:

```rust
if create && owner.is_none() {
    return Err(ParseError::Malformed(
        "missing required header: Owner".into(),
    ));
}
```

I ran the PR-branch `aimx mcp` against the daemon already running on this host as a non-root user; the tool returns `[MALFORMED] missing required header: Owner` instead of creating the mailbox. There is no test coverage for this — the existing protocol unit tests still assert the parser rejects, and the MCP unit tests in `mcp.rs` never reach the wire layer. The fix is either to relax the parser (drop the unconditional Owner requirement and let the handler synthesize for non-root) or to keep the parser strict and have the MCP tool send a placeholder Owner that the handler discards. PRD NFR1 actually requires the daemon ignore client-supplied Owner from non-root — so the cleanest fix is to drop the parser requirement and let the handler resolve.

### Blocker 3 — `mailbox_delete` MCP tool's `force: true` path also broken on real install (`src/mcp.rs:698`)

`let config = self.load_config()?` fails with `Permission denied` for the non-root MCP process for the same reason as Blocker 1. The whole `force: true` branch (catchall short-circuit aside) never reaches the wipe or the UDS submit.

### Non-blocker 4 — Data destruction when daemon dies between wipe and DELETE submission (`src/mailbox.rs:701-731`, `src/mcp.rs:706-714`)

For both the CLI `delete --force` and the MCP `mailbox_delete { force: true }` paths, the order is: pre-flight authorize → wipe inbox + sent → submit MAILBOX-DELETE. If the daemon is up at the start of the call but unreachable at the time of the UDS submit (daemon crashed, socket cleaned up, race), the user's mail has been wiped but the config still references the now-empty mailbox. CLI exits with code 2 saying "daemon must be running"; MCP returns `"aimx daemon not running. Start with 'sudo systemctl start aimx'"`. Neither indicates the data is already gone. Mitigation options: probe socket existence before wiping, or wipe atomically only after a daemon ack.

### Non-blocker 5 — `prompt_mailbox_owner` runs even when daemon will discard the value (`src/mailbox.rs:327`)

When a non-root caller runs `aimx mailboxes create my-agent` (no `--owner`), `resolve_create_owner` calls `prompt_mailbox_owner` because `my-agent` doesn't resolve via `getpwnam`. Under non-TTY stdin the prompt loops 5 times before erroring. The daemon is going to overwrite the resolved value with `peer_username(SO_PEERCRED)` anyway. The S2-6 test works around this by always passing `--owner <runner>` and the test comment explicitly notes the issue. Cleaner UX would short-circuit `resolve_create_owner` for non-root callers and simply use `caller_username()` (which the soft-warning code already computes).

### Non-blocker 6 — `mailbox_create` MCP success message degrades silently when `load_config` fails (`src/mcp.rs:649-657`)

```rust
let domain = self.load_config().map(|c| c.domain).unwrap_or_else(|_| String::new());
if domain.is_empty() {
    Ok(format!("Mailbox '{}' created.", params.name))
} else {
    Ok(format!("{}@{domain}", params.name))
}
```

In production (Blocker 1 / Blocker 3 fix permitting), `load_config` will fail for non-root, so the fallback path always fires. Agents lose the address from the success message. Once the broader config-readability fix lands, this will resolve naturally; flagging because the current shape masks the failure mode.

## Test Coverage

- **No integration test for MCP `mailbox_create` against a real daemon.** The unit tests in `src/mcp.rs::email_list_tests` only exercise param parsing and the catchall short-circuit. Without an end-to-end test that actually submits through the UDS, the protocol-parser rejection (Blocker 2) was invisible to CI. Add a test mirroring `mailbox_create_delete_force_e2e_as_non_root_user` but using the MCP tool path.
- **No integration test for the production config-perm scenario.** Every existing test runs against a writable tmp config. A test that explicitly chmods the test config to `0640 root:root` and confirms the CLI surfaces the actionable error (rather than the bare perm-denied) would have caught Blocker 1. This is hard to do without sudo, but at least the issue should be documented as a known gap if tests can't cover it.
- **No test for the daemon-down-mid-delete data-destruction window.** A test that starts the daemon, runs a `--force` delete, kills the daemon between the wipe and the submit, and then asserts on the resulting state would have caught Non-blocker 4.

## Code Quality

- `format_auth_error` (in `src/mailbox.rs`, `src/hooks.rs`, and `src/mcp.rs`) duplicates the same four-arm match in three places. With `OwnerMismatch` added, all three need to stay synced — easy to drift. Consider promoting one canonical renderer to `src/auth.rs`.
- `SOCKET_MISSING_HINT` formatting (`src/mailbox.rs:16-17`) uses the `\` line-continuation correctly (Rust eats the trailing newline + leading whitespace), so the runtime string is correctly `"...aimx serve` `or run with sudo..."` — non-issue, but the explicit indentation reads alarming. A `concat!()` of two literal halves would be more obvious.

## Alignment with PRD

- **NFR1 (privilege escalation defense) — preserved.** The daemon-side defense from Sprint 1 (`mailbox_handler.rs:124-149` — ignoring `req.owner` and synthesizing from `peer_username(caller.uid)`) is still in place. No client-side bypass is introduced. This invariant holds **even though** Blocker 2's protocol parser rejects valid non-root CREATE requests — it errs on the safe side.
- **PRD §6.5 R20** ("MailboxCrudRequest's owner field stays Option<String>. When the caller is non-root, the daemon ignores any value present (R6). No protocol-level changes needed; the field becomes 'honored only for root callers.'") — **violated** by the parser still requiring the header. This is a pre-existing bug surfaced by Sprint 2's MCP tool; Sprint 2 should have caught and fixed it.
- **PRD §6.3 R11** ("If the socket is missing, surface a precise error... Do not silently attempt the direct config.toml write as a non-root user — it will fail at the 0640 perm and the error will be confusing.") — Sprint 2 implements this correctly inside `mailbox::run`, but the broader "fail at 0640 perm" issue at `Config::load` itself was not addressed. The PRD implicitly assumed Config could be loaded; that assumption is wrong.

**Overall verdict: Needs significant rework**

The acceptance criteria look met when reading the diff, but two basic production smoke tests (running the binary as a non-root user against the real daemon already on this machine) prove the headline goal is not delivered. The work in `mailbox.rs`, `mcp.rs`, and `auth.rs` is good — the gap is that the surrounding system (config perms, wire-protocol parser) was not updated to match the new model. The PR's claim that 1021 unit + 94 integration tests pass is correct but misleading: none of those tests reach the production failure mode.

1. CLI cannot even reach `mailbox::run` for non-root callers in production because `/etc/aimx/config.toml` is `0640 root:root` and `dispatch()` reads it before dispatching `Mailboxes`. Repro: `aimx mailboxes create my-agent` as ubuntu → `Error: Permission denied (os error 13)`.
2. MCP `mailbox_create` rejected by the wire protocol because `submit_mailbox_crud_via_daemon(name, true, None)` triggers the parser's "missing required header: Owner" check at `src/send_protocol.rs:691`. Repro: invoke `mailbox_create` over stdio MCP against the running daemon as non-root → `[MALFORMED] missing required header: Owner`.
3. MCP `mailbox_delete { force: true }` for non-root in production fails at `self.load_config()?` for the same reason as Blocker 1.

**Non-blockers (should fix before next sprint):**

4. Data destruction window when daemon dies between the wipe and the DELETE submit (CLI and MCP both).
5. `prompt_mailbox_owner` runs unnecessarily on the non-root CLI create path; daemon will discard the value.
6. `mailbox_create` MCP tool degrades silently to a no-address success message when `load_config` fails.
7. `format_auth_error` duplicated across three modules — risk of drift.
8. No integration test for MCP `mailbox_create` against a real daemon. No test for the production-config-perm scenario.

uzyn:
## Cycle 2 fixes pushed (bdba1a4)

All three blockers from the Cycle 1 review are addressed end-to-end. Verified by 3 new regression tests + manual repro on this host (the bare `Permission denied (os error 13)` no longer surfaces; CLI now reaches the daemon and the daemon's own authz response surfaces).

### Blockers resolved

**Blocker 1 — CLI EACCES on non-root.** `Command::Mailboxes` is now dispatched alongside the other thin UDS clients (`Send`/`Mcp`/`Logs`) instead of falling through `dispatch()`'s catch-all arm. Config loading happens lazily inside `mailbox::run` and gracefully degrades: `create`/`delete` route through the daemon UDS without ever needing local config-read; `list` falls back to `MAILBOX-LIST` over UDS; `show` is the only subcommand that genuinely needs config and now produces a friendly hint instead of bare EACCES.

**Blocker 2 — MCP `mailbox_create` rejected by parser.** `parse_mailbox_crud()` now treats `Owner:` as optional on the wire. The daemon-side handler (`mailbox_handler::handle_mailbox_crud`) enforces the policy split per PRD §6.5 R20: root callers without `Owner:` get `ErrCode::Validation`; non-root callers have any wire-supplied owner ignored and the owner synthesized from `peer_username(SO_PEERCRED)`. PRD docstring updated.

**Blocker 3 — MCP `mailbox_delete { force: true }` failed at `load_config`.** Added a `Force: true` header to `MAILBOX-DELETE`. The daemon performs the wipe under the per-mailbox lock that already guards the stanza removal, so wipe + config rewrite are atomic together. Both CLI (`src/mailbox.rs`) and MCP (`src/mcp.rs`) `--force` paths now delegate the wipe entirely to the daemon — no client-side `wipe_mailbox_contents` calls remain on the UDS path. **This also resolves Non-blocker 4** (data-destruction race window) by construction.

### Non-blockers

- **Non-blocker 5** (resolved): `prompt_mailbox_owner` no longer runs for non-root CLI create. Auto-defaults to the caller's username for the display string; the daemon discards anything we send anyway.
- **Non-blocker 6** (resolved): MCP `mailbox_create` resolves the new mailbox's full address through `MAILBOX-LIST` (rows now carry an `address` field). Falls back to a name-only success message when the listing is unavailable.
- **Non-blocker 7** (deferred to Sprint 3): `format_auth_error` duplication across `mailbox.rs` / `hooks.rs` / `mcp.rs`. Marked the unused `mailbox.rs` renderer as `#[cfg(test)]` so the dead-code warning doesn't surface and the cleanup target stays visible. Will fold into a canonical renderer in Sprint 3.
- **Non-blocker 8** (resolved): three new regression tests pin the fixes:
  - `mailbox_create_with_unreadable_config_does_not_surface_eacces` — chmods the test config to `0000` and asserts CREATE still works via UDS without surfacing bare EACCES.
  - `mailbox_list_with_unreadable_config_falls_back_to_uds` — companion test for the LIST UDS fallback.
  - `mcp_mailbox_create_against_running_daemon_succeeds_for_non_root` — pins the MCP create path end-to-end against a real daemon (the canonical Cycle-1-Blocker-2 reproduction).
  - Also updated two stale tests (`mcp_mailbox_create_tool_no_longer_exposed` / `..delete..`) that asserted the MCP tools were "no longer exposed" — Sprint 2 re-added them, so the assertions are now `..._is_exposed` and check for the daemon-not-running tool error rather than the framework's unknown-tool error.

### Wire format changes

- `MAILBOX-CREATE`: `Owner:` header now optional (parser only — daemon still requires it for root callers).
- `MAILBOX-DELETE`: new optional `Force: true` header.
- `MAILBOX-LIST` JSON rows: new `address: Option<String>` field. Existing fields unchanged; the integration test that asserts on `inbox_path` / `sent_path` / `registered` continues to pass.

### Quality gates (all foreground, all passing)

| Gate | Result |
|---|---|
| `cargo fmt -- --check` (root) | clean |
| `cargo fmt -- --check` (verifier) | clean |
| `cargo clippy --all-targets -D warnings` (root) | clean |
| `cargo clippy --all-targets -D warnings` (verifier) | clean |
| `cargo test` (root) | 1021 unit + 97 integration + 1 (other) passed |
| `cargo test` (verifier) | 43 passed |
| `sudo … cargo test --test uds_authz -- --ignored` | 21 passed |
| `sudo … cargo test --test integration -- --ignored` | 5 passed |

uzyn (commented):
# Cycle 2 Re-Review — APPROVED with one nit

All three Cycle 1 blockers are resolved end-to-end. Verified by independent reproduction on this host: I built the PR-branch binary (commit `bdba1a4`), spun up the daemon with a `0640 root:root` config in a tmp dir, and ran the original Cycle 1 negative tests as the non-root `ubuntu` user.

## Resolved blockers

### Blocker 1 — CLI `Permission denied (os error 13)` for non-root in production [RESOLVED]

`Command::Mailboxes` is now dispatched alongside the thin UDS clients (`Send` / `Mcp` / `Logs`) in `src/main.rs:149`, bypassing the catch-all `?`-propagating `Config::load_resolved_with_data_dir`. Inside `mailbox::run`, config is loaded lazily via `load_config_optional` (returns `Option<Config>`); each subcommand decides whether the missing-config case is recoverable.

Reproduction against PR-branch binary as `ubuntu`:

```
$ ls -la /tmp/aimx-review-test/config.toml
-rw-r----- 1 root root 205 May  2 14:40 /tmp/aimx-review-test/config.toml
$ aimx mailboxes list
MAILBOX              INBOX    SENT
catchall             0        0
$ aimx mailboxes create review-mb
Mailbox 'review-mb' created (owner: ubuntu).
```

No bare EACCES anywhere. The mailbox is created with the runner's uid on disk (`drwx------ 2 ubuntu ubuntu`), config stanza written with `owner = \"ubuntu\"`.

### Blocker 2 — MCP `mailbox_create` rejected by wire-protocol parser [RESOLVED]

`parse_mailbox_crud_headers` (`src/send_protocol.rs:735-749`) now treats `Owner:` as optional. The daemon-side handler (`mailbox_handler.rs:110-149`) enforces the policy split:

- **Root + no `Owner:`** → `ErrCode::Validation` ("MAILBOX-CREATE requires an Owner: header")
- **Non-root + any `Owner:` value** → wire value dropped on the floor; owner synthesized from `peer_username(caller.uid)` via SO_PEERCRED
- **Non-root from orphan uid** → `ErrCode::Validation` ("uid <N> has no passwd entry")

Reproduction via real MCP stdio session against running daemon as non-root:

```
{"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"mcp-mb@agent.example.com"}],"isError":false}}
```

The full address is returned (also exercises the Non-blocker 6 fix — `lookup_mailbox_address` resolves through `MAILBOX-LIST` rather than reading root-owned config).

### Blocker 3 — MCP `mailbox_delete { force: true }` failed at `load_config()` [RESOLVED]

`MAILBOX-DELETE` now carries an optional `Force: true` header (`src/send_protocol.rs:684-707`, `src/mailbox_handler.rs:521-533`). The wipe runs **inside** `handle_delete` under the per-mailbox lock + `CONFIG_WRITE_LOCK` that already guards the stanza removal. No client-side `wipe_mailbox_contents` calls remain on the UDS path, so the MCP tool no longer needs `load_config`.

Live verification:

```
[force-mb mailbox seeded with 1 inbox + 1 sent .md]
$ mcp call mailbox_delete {"name":"force-mb"}
[NONEMPTY] mailbox 'force-mb' has 2 files (1 in inbox, 1 in sent); archive or remove them first
$ mcp call mailbox_delete {"name":"force-mb","force":true}
Mailbox 'force-mb' deleted.
[both inbox/force-mb/ and sent/force-mb/ are now empty; stanza removed from config.toml]
```

## Atomicity claim verified (Blocker 3 / Non-blocker 4)

Reading `mailbox_handler::handle_delete` (lines 495-567):

1. Per-mailbox lock acquired in `handle_mailbox_crud:151-152` (outer)
2. `CONFIG_WRITE_LOCK` acquired in `handle_mailbox_crud:159-161` (inner)
3. `wipe_mailbox_dirs` runs at line 528 — under both locks
4. NONEMPTY guard at line 535-548 — under both locks
5. `write_config_atomic` (temp-then-rename) at line 555 — under both locks
6. `config_handle.store` at line 562 — under both locks

There is no early-return between wipe and config rewrite; both happen under one lock-guarded critical section. The "data-destruction race window" Non-blocker 4 from Cycle 1 is closed by construction. If the daemon dies between wipe and the file rename, the temp-file rename is atomic (POSIX `rename(2)` semantics), so the next daemon startup either sees the old config (with the mailbox stanza still present, but now-empty dirs) or the new config (mailbox gone). No torn state.

## Privilege-escalation defense verified (Sprint 1 NFR1 holds)

Negative test: non-root caller passes `--owner root` explicitly:

```
$ aimx mailboxes create esc-mb --owner root
Warning: --owner ignored for non-root callers; mailbox will be owned by `ubuntu`
Mailbox 'esc-mb' created (owner: root).
[on disk]
[mailboxes.esc-mb]
owner = "ubuntu"
```

Wire-supplied `Owner: root` is silently discarded by the daemon; SO_PEERCRED-resolved `ubuntu` is what gets persisted.

## Resolved non-blockers

- **Non-blocker 4** (data-destruction race) — closed by Blocker 3's daemon-side wipe (verified above).
- **Non-blocker 5** (`prompt_mailbox_owner` looping under non-TTY) — `mailbox.rs:456-469` short-circuits to `caller_username()` for non-root.
- **Non-blocker 6** (silent address degradation) — `lookup_mailbox_address` queries `MAILBOX-LIST` over UDS instead of reading root-owned config (`mcp.rs:1186-1193`).
- **Non-blocker 8** (test coverage) — three new regression tests pin the fixes:
  - `mailbox_create_with_unreadable_config_does_not_surface_eacces` chmods the test config to `0000` and asserts CREATE still works via UDS without surfacing bare EACCES. **Verified via `cargo test`.**
  - `mailbox_list_with_unreadable_config_falls_back_to_uds` companion test for the LIST UDS fallback. **Verified via `cargo test`.**
  - `mcp_mailbox_create_against_running_daemon_succeeds_for_non_root` pins the MCP create path end-to-end against a real daemon and asserts the on-disk owner is the runner (SO_PEERCRED-bound). **Verified via `cargo test`.**

All three tests run as non-root and use the standard `tempfile::TempDir` + `setup_test_env` isolation pattern.

## Deferred non-blocker (acknowledged)

- **Non-blocker 7** (`format_auth_error` duplication) — deferred to Sprint 3. The `mailbox.rs::format_auth_error` is now `#[cfg(test)]` so it doesn't trip dead-code warnings while staying visible as a cleanup target. Acceptable.

## New issues (one nit, non-blocking)

### Non-blocker (cosmetic): success line lies about owner when non-root passes `--owner <other>`

In `src/mailbox.rs::create` at line 509, after the soft warning correctly informs the user that `--owner` will be ignored, the success line still echoes the *requested* owner verbatim:

```
$ aimx mailboxes create esc-mb --owner root
Warning: --owner ignored for non-root callers; mailbox will be owned by `ubuntu`
Mailbox 'esc-mb' created (owner: root).   # <-- says "root", but on disk it's "ubuntu"
```

The on-disk state is correct (`owner = \"ubuntu\"`), so this is purely a display bug. Suggest: when the non-root branch overrides `--owner` with `caller_username()`, use the caller's username in the success line too. Cheap fix:

```rust
let owner = if !is_root() {
    let caller = caller_username();
    let resolved = match owner {
        Some(o) if !o.is_empty() => {
            if !caller.is_empty() && o != caller {
                eprintln!("{} --owner ignored ...", term::warn("Warning:"));
            }
            // Persist the actually-bound owner for downstream display
            if !caller.is_empty() { caller } else { o.to_string() }
        }
        ...
    };
    resolved
} else { ... };
```

Not blocking — does not affect security, correctness on disk, or test coverage. Worth a quick follow-up so the success line is honest.

## Scope alignment

All Sprint 2 acceptance criteria (S2-1 through S2-6) are now genuinely met in production, not just in the test harness. Wire-format additions (`Owner:` optional on CREATE, `Force:` on DELETE, `address` field on `MailboxListRow`) are backward-compatible and well-documented inline.

## Quality gates

- `cargo test --release --test integration -- mailbox_create_with_unreadable_config... mailbox_list_with_unreadable_config... mcp_mailbox_create_against_running_daemon... mailbox_create_delete_force_e2e_as_non_root_user` — **4 passed**, 0 failed
- Live smoke (PR-branch binary + PR-branch daemon, non-root caller, `0640 root:root` config): CLI list, CLI create, MCP `mailbox_create`, MCP `mailbox_delete force=false` (NONEMPTY refusal), MCP `mailbox_delete force=true` (wipe + delete), privilege-escalation negative test — **all pass**.

## Verdict

**APPROVED.** All three Cycle 1 blockers are resolved. Non-blockers 4, 5, 6, 8 are also resolved. Non-blocker 7 is acknowledged as Sprint 3 work. One new cosmetic non-blocker (success-line owner display) is worth a quick follow-up but does not block merge.

```
[Sprint 2] CLI surface + MCP tools — non-root mailbox CRUD end-to-end

Lifts the CLI's entry-point root gate, finishes `--force` plumbing for
non-root owners, and adds MCP `mailbox_create` / `mailbox_delete` so
agents can self-serve. Builds on Sprint 1's SO_PEERCRED-bound daemon
authz.

Key changes:

- `Command::Mailboxes` dispatches alongside thin UDS clients; config
  loaded lazily so non-root callers reach `mailbox::run` even when
  `/etc/aimx/config.toml` is `0640 root:root` in production. Each
  subcommand decides whether the missing-config case is recoverable
  (CREATE/DELETE go through daemon UDS; LIST falls back to MAILBOX-LIST;
  SHOW errors with an actionable hint).

- `MAILBOX-DELETE` carries optional `Force: true` header. The daemon
  performs the wipe under the same per-mailbox lock + CONFIG_WRITE_LOCK
  that already guards the stanza removal — wipe and config rewrite are
  atomic together, eliminating the data-destruction race window.

- `MAILBOX-CREATE` accepts optional `Owner:` on the wire; daemon
  enforces the policy split (root must supply Owner; non-root has any
  value ignored, owner synthesized from SO_PEERCRED).

- `MAILBOX-LIST` rows carry new `address` field so MCP `mailbox_create`
  can resolve the new mailbox's full address without reading root-owned
  config from a non-root process.

- 3 new regression tests pin the production-perm scenarios that the
  Cycle 1 review surfaced: chmod 0000 on test config + CLI CREATE,
  chmod 0000 + CLI LIST, MCP mailbox_create against a real daemon as
  non-root.

Sprint 3 (book/ rewrites, release-notes entry, CLAUDE.md updates,
format_auth_error consolidation) follows.
```

uzyn:
## Cycle 2 follow-up: success-line display fix

Addressed the remaining cosmetic non-blocker.

### What changed
`src/mailbox.rs` — `create()` now computes a single `effective_owner` value (the caller's username for non-root callers via `caller_username()`, the resolved owner for root) and uses it in **both** the S2-1 soft-warning AND the success line. Previously the success line echoed the user-supplied `--owner` value verbatim, so `--owner root` from a non-root caller would print `Mailbox 'esc-mb' created (owner: root)` even though the on-disk owner was the caller's uid.

The wire payload to `submit_mailbox_crud_via_daemon` still carries the original requested `owner` (the daemon discards it server-side anyway via SO_PEERCRED) so daemon-side behavior is unchanged — this is purely a client-side display fix, no protocol change.

### MCP `mailbox_create`
No corresponding bug — its success path returns the address resolved via `lookup_mailbox_address(...)` (a daemon-side `MAILBOX-LIST` lookup), so it already reflects the effective state, not the requested one.

### Quality gates
- `cargo fmt -- --check` (root + verifier): pass
- `cargo clippy --all-targets -- -D warnings` (root + verifier): pass
- `cargo test` (root): 1021 unit + 97 integration + others, **0 failures**

No assertion text in tests referenced the old success line, so no test updates were needed. Sudo-lane integration tests were not re-run since the change is a display-string-only edit on the client.

uzyn (commented):
# Cycle 3 Re-Review — APPROVED

The single Cycle 2 cosmetic non-blocker is resolved. All previously confirmed Cycle 1 + Cycle 2 fixes remain intact. No regressions introduced.

## Cycle 2 nit — RESOLVED

**`src/mailbox.rs:509`: success line echoed requested owner verbatim instead of the daemon-synthesized owner.**

Commit `d2c412a` introduces a single `effective_owner` value computed once at lines 490-499:

```rust
let effective_owner = if !is_root() {
    let caller = caller_username();
    if caller.is_empty() { owner.clone() } else { caller }
} else {
    owner.clone()
};
```

This value is now threaded through both:

1. **Soft-warning condition + body** (lines 506-511) — now reads `owner != effective_owner` and prints `effective_owner` rather than recomputing the caller inline. Behavior identical for the only triggering case (non-root + `--owner <other>`).
2. **UDS-success success line** (lines 522-524) — `(owner: {effective_owner})` instead of `(owner: {owner})`.
3. **Root-fallback success line** (lines 545-547) — same change for symmetry.

The on-disk write at line 542 (`create_mailbox(cfg, name, &owner)`) and the wire submit at line 518 (`submit_mailbox_crud_via_daemon(name, true, Some(&owner), false)`) intentionally still use the original `owner` — for root this equals `effective_owner`; for non-root the daemon discards the wire value and binds via SO_PEERCRED anyway. So no behavior change on the data plane; only display strings updated.

### Branch matrix verified

| Caller | `--owner` | `owner` (resolved) | `effective_owner` | Warning? | Success line says |
|---|---|---|---|---|---|
| root | omitted | prompted | same | no | resolved value |
| root | `<X>` | `X` | `X` | no | `X` |
| non-root | omitted | caller | caller | no | caller |
| non-root | matches caller | caller | caller | no | caller |
| **non-root** | **`<other>`** | **`<other>`** | **caller** | **yes** | **caller** ← the fix |
| non-root, orphan uid | `<X>` | `X` | `X` (fallback) | no (equal) | `X` |
| non-root, orphan uid | omitted | — | (errors at line 463) | n/a | n/a |

The orphan-uid branch is interesting: when `caller_username()` returns empty AND `--owner <X>` is supplied, `effective_owner` falls back to `owner` (line 493), the warning condition `owner != effective_owner` is false, and the success line displays `X`. This is honest because the daemon will also reject the request for orphan uids (Sprint 1 R20 — `Validation: "uid <N> has no passwd entry"`), so the success message would never actually print. Acceptable.

## Quality gates

- `cargo test --bin aimx mailbox::` — 8 passed, 0 failed
- `cargo test --test integration mailbox` — 21 passed, 0 failed, 5 ignored (root-only)
- Live smoke (PR-branch release binary, non-root caller against system daemon): warning fires correctly with `mailbox will be owned by ubuntu` (system-daemon auth context blocks the actual create, which is expected and unrelated to this fix).

## Carryover

- **Non-blocker 7** (`format_auth_error` duplication across `mailbox.rs` / `hooks.rs` / `mcp.rs`) — explicitly deferred to Sprint 3 cleanup, as agreed in Cycle 2.

## Verdict

**APPROVED.** The Cycle 2 cosmetic non-blocker is resolved cleanly with no introduced bugs. The fix is narrowly scoped (one file, +27/-10 lines), respects the existing data-plane invariants, and the previously-passing test `mailbox_create_owner_flag_warns_for_non_root_callers` still validates the warning surface. Ready to merge.

```
[Sprint 2] CLI surface + MCP tools — non-root mailbox CRUD end-to-end

Lifts the CLI's entry-point root gate, finishes `--force` plumbing for
non-root owners, and adds MCP `mailbox_create` / `mailbox_delete` so
agents can self-serve. Builds on Sprint 1's SO_PEERCRED-bound daemon
authz.

Key changes:

- `Command::Mailboxes` dispatches alongside thin UDS clients; config
  loaded lazily so non-root callers reach `mailbox::run` even when
  `/etc/aimx/config.toml` is `0640 root:root` in production. Each
  subcommand decides whether the missing-config case is recoverable
  (CREATE/DELETE go through daemon UDS; LIST falls back to MAILBOX-LIST;
  SHOW errors with an actionable hint).

- `MAILBOX-DELETE` carries optional `Force: true` header. The daemon
  performs the wipe under the same per-mailbox lock + CONFIG_WRITE_LOCK
  that already guards the stanza removal — wipe and config rewrite are
  atomic together, eliminating the data-destruction race window.

- `MAILBOX-CREATE` accepts optional `Owner:` on the wire; daemon
  enforces the policy split (root must supply Owner; non-root has any
  value ignored, owner synthesized from SO_PEERCRED).

- `MAILBOX-LIST` rows carry new `address` field so MCP `mailbox_create`
  can resolve the new mailbox's full address without reading root-owned
  config from a non-root process.

- CLI `create` success line now displays the daemon-synthesized owner
  (caller's username for non-root) rather than the requested owner, so
  the soft-warning and the success message agree on what landed on disk.

- 3 new regression tests pin the production-perm scenarios that the
  Cycle 1 review surfaced: chmod 0000 on test config + CLI CREATE,
  chmod 0000 + CLI LIST, MCP mailbox_create against a real daemon as
  non-root.

Sprint 3 (book/ rewrites, release-notes entry, CLAUDE.md updates,
format_auth_error consolidation) follows.
```

PR #204, opened 2026-05-06
install.sh: prompt to re-run setup when AIMX is already installed + capitalize brand prose

uzyn:
## Summary
- When `curl -fsSL https://aimx.email/install.sh | sh` is re-run and the binary already matches the target tag, ask "AIMX is already installed. Re-run setup to (re)configure it? [y/N]" instead of hard-exiting. Lets an operator who aborted partway through `aimx setup` (e.g. to grab a different domain) resume without knowing about `--force`. Default is **No** so non-interactive callers (CI, scripted upgrades) keep today's exit-0 semantics; `--force` still bypasses the prompt. `AIMX_DRY_RUN=1` also bypasses the prompt — auditing the script stays non-interactive.
- On "yes" the script skips the download/extract/install entirely (binary is already correct) and falls through to `backup_existing_config` + `exec aimx setup` — the same handoff fresh installs already use.
- Capitalize the brand name **AIMX** in operator-facing prose. Lowercase `aimx` stays for command literals (`aimx setup`, `aimx doctor`), service unit references (`aimx.service`), paths (`/etc/aimx/...`), and the repo identifier (`uzyn/aimx`).
- Silence pre-existing shellcheck false positives in `install.sh` + `tests/install_sh.sh` (SC2024 on `(sudo -v </dev/tty)`, SC2016 on intentional single-quoted printf/grep literals, SC2010 on `ls | grep` over a test-controlled tmpdir) so the test suite reaches a clean 109/109.

## Why
Re-running `install.sh` after aborting the wizard mid-flow is the natural "I got the wrong domain, let me start over" recovery path, but today it dead-ends on `aimx 0.0.7 is already installed (pass --force to re-install)` — a message that both implies a binary swap the operator doesn't need and isn't discoverable as a recovery handle. The existing `backup_existing_config` rename already protects partial config from clobbering, so the right fix is to ask and fall through.

The brand sweep was raised in the same conversation: the shell wrapper's prose had drifted from the wizard, which already uses `AIMX setup` / `Install AIMX` / `AIMX has been set up successfully` correctly.

## Docs

- `book/installation.md` — rewrote step 9 of the install-flow numbered list to describe the new prompt-on-equal-version behavior, default-no semantics, non-TTY exit-0 path, and `--force` bypass.
- `book/release-notes.md` — added an Unreleased entry under the existing `install.sh` block covering both the recovery prompt and the brand-prose capitalization sweep.

## Test plan
- [x] `sh tests/install_sh.sh` — 109/109 passing.
- [x] `shellcheck install.sh` and `shellcheck -s sh tests/install_sh.sh` both exit 0 with no warnings.
- [x] `sh -n install.sh && sh -n tests/install_sh.sh` — POSIX-clean.
- [x] Prompt assertions cover the full accept-list (y / Y / yes / YES / Yes) plus rejects (n / empty / no-TTY) via an overridable `_prompt_read` stub.
- [x] No-TTY regression assertion: with `_prompt_read` signalling no usable TTY, the prompt question itself does NOT appear in stderr.
- [x] Dry-run regression test: with a fake aimx binary at `${PREFIX}/aimx` reporting the target tag and `AIMX_DRY_RUN=1`, the recovery prompt is never invoked and the script exits 0 with the "AIMX X.Y.Z is already installed" line.
- [x] Branding regressions: `(installing|installed|upgraded) aimx\b`, `\baimx is\b`, and `aimx [0-9]+\.[0-9]` all return zero hits.
- [ ] Manual end-to-end on a real VPS (cannot reproduce in CI):
  - Fresh install → abort wizard at the domain prompt → re-run `install.sh` → confirm prompt appears, Enter exits 0, `y` skips the download and re-enters `aimx setup`.
  - Re-run with `--force` → bypasses the prompt and re-installs.
  - `curl ... | sh </dev/null` (no TTY) → exits 0 with the "already installed" line, no prompt printed.

### DIFF
--- a/book/installation.md
+++ b/book/installation.md
@@ -35,7 +35,7 @@ The shell script is thin: it acquires `sudo` up front, downloads the binary, and
 6. Extracts into a temp directory cleaned up on every exit path.
 7. Installs the binary as `install -m 0755 /usr/local/bin/aimx` (override with `--to` / `AIMX_PREFIX`).
 8. On a fresh box, backs up any pre-existing `/etc/aimx/config.toml` to `config.toml.bak-YYYYMMDD-HHMMSS`, then `exec`s `sudo aimx setup </dev/tty`. DKIM keys and STARTTLS certs are preserved across re-runs.
-9. If an older `aimx` is already installed, the upgrade path runs instead: stop the service, swap the binary atomically, restart. No wizard re-run. If the running version matches the target, it exits without touching anything (pass `--force` to reinstall).
+9. If an older `aimx` is already installed, the upgrade path runs instead: stop the service, swap the binary atomically, restart. No wizard re-run. If the running version matches the target, the script asks `AIMX is already installed. Re-run setup to (re)configure it? [y/N]` — answer `y` to skip the download and re-enter `aimx setup` (handy if you aborted the wizard partway through), or `N` / Enter / no usable TTY (CI, scripted callers) to exit `0` without touching anything. `--force` skips the prompt and reinstalls the binary.
 
 In CI / non-TTY contexts, set `AIMX_NONINTERACTIVE=1` and supply defaults — see [Setup](setup.md).
 
--- a/book/release-notes.md
+++ b/book/release-notes.md
@@ -91,6 +91,14 @@ When the tags differ, restart the service (`systemctl restart aimx`, or `rc-serv
 - A new `pgrep`-based detector warns (never signals) when a manually-launched `aimx serve` is running outside systemd / OpenRC.
 - A single post-start `systemctl is-active` check confirms the daemon came up, with a `journalctl -u aimx -n 20` hint on failure.
 
+### `install.sh` recovery prompt when the binary is already at the target version
+
+Re-running `curl -fsSL https://aimx.email/install.sh | sh` on a host whose `aimx` binary already matches the target tag used to hard-exit with `aimx 0.0.7 is already installed (pass --force to re-install)`. That stranded operators who had aborted partway through `aimx setup` (e.g. to grab a different domain) and now wanted to redo configuration — they had to know about `--force`, which implies a binary swap they don't need.
+
+The script now prompts `AIMX is already installed. Re-run setup to (re)configure it? [y/N]` on that branch. Answer `y` to skip the download/extract/install step entirely (binary is already correct) and fall through to `backup_existing_config` + `exec aimx setup`, the same handoff fresh installs use. Answer `N` (default) or run with no usable TTY (CI, fully-scripted callers, `AIMX_DRY_RUN=1`) and the script keeps the existing exit-0 semantics. `--force` continues to bypass the prompt and reinstall the binary.
+
+Operator-facing prose strings in `install.sh` were also capitalized to **AIMX** to match the wizard. Lowercase `aimx` is preserved for command literals (`aimx setup`), service unit references (`aimx.service`), paths (`/etc/aimx/...`), and the GitHub repo identifier (`uzyn/aimx`).
+
 ### `aimx upgrade` confirms the restart
 
 After `wait_for_service_ready` returns true, `aimx upgrade` prints one extra line:
--- a/install.sh
+++ b/install.sh
@@ -252,13 +252,17 @@ resolve_sudo_prefix() {
 # 1 on no / no usable TTY. Default is no (Enter = no), so non-interactive
 # callers (CI, fully-scripted) keep today's exit-0 semantics.
 #
-# Same TTY logic as ensure_sudo and the post-install handoff: prefer
-# the script's own stdin if it's already a terminal; fall back to
-# /dev/tty when it's a pipe (curl | sh); otherwise no prompt.
+# _prompt_read prints the prompt only when a TTY is available for the
+# answer. Same TTY logic as ensure_sudo and the post-install handoff:
+# prefer the script's own stdin if it's already a terminal; fall back
+# to /dev/tty when it's a pipe (curl | sh); otherwise no prompt at all
+# (so `curl | sh </dev/null` stays quiet — no stray question in CI logs).
 _prompt_read() {
     if [ -t 0 ]; then
+        printf '%s' "$1" >&2
         read -r _ans
     elif [ -e /dev/tty ] && [ -r /dev/tty ]; then
+        printf '%s' "$1" >&2
         read -r _ans </dev/tty
     else
         return 1
@@ -267,8 +271,7 @@ _prompt_read() {
 
 prompt_reinstall() {
     _ans=""
-    printf '%s' 'AIMX is already installed. Re-run setup to (re)configure it? [y/N] ' >&2
-    _prompt_read || return 1
+    _prompt_read 'AIMX is already installed. Re-run setup to (re)configure it? [y/N] ' || return 1
     case "${_ans}" in
         y | Y | yes | YES | Yes) return 0 ;;
         *) return 1 ;;
@@ -1256,10 +1259,13 @@ main() {
                     if [ "${FORCE}" -eq 1 ]; then
                         say "re-installing ${TAG} (--force)"
                         _is_upgrade=1
-                    elif prompt_reinstall; then
+                    elif [ "${DRY_RUN}" != "1" ] && prompt_reinstall; then
                         # Binary is already correct; skip the swap and
                         # jump straight to the wizard. backup_existing_config
                         # below covers any partial config from an aborted run.
+                        # Dry-run never prompts — auditing the script must
+                        # stay non-interactive, matching pre-PR behavior on
+                        # the equal-version branch.
                         _skip_swap=1
                         _is_upgrade=0
                     else
--- a/install.sh
+++ b/install.sh
@@ -114,7 +114,7 @@ ui_success() {
 print_install_banner() {
     printf '\n' >&2
     printf '%s\n' "$(_ui_paint '1;35' 'AIMX installer')" >&2
-    printf '%s\n' "$(_ui_paint 2 '  downloading and installing aimx...')" >&2
+    printf '%s\n' "$(_ui_paint 2 '  downloading and installing AIMX...')" >&2
     printf '\n' >&2
 }
 
@@ -122,7 +122,7 @@ print_install_banner() {
 # is set, so it is visually obvious nothing is installing.
 print_port_check_banner() {
     printf '\n' >&2
-    printf '%s\n' "$(_ui_paint '1;35' 'aimx port 25 connectivity check')" >&2
+    printf '%s\n' "$(_ui_paint '1;35' 'AIMX port 25 connectivity check')" >&2
     printf '%s\n' "$(_ui_paint 2 '  no install will be performed')" >&2
     printf '\n' >&2
 }
@@ -167,7 +167,7 @@ download() {
 
 help() {
     cat <<'EOF'
-aimx install script
+AIMX install script
 
 USAGE:
     install.sh [FLAGS]
@@ -245,6 +245,34 @@ resolve_sudo_prefix() {
     else
         SUDO=""
     fi
+}
+
+# prompt_reinstall — ask the operator whether to re-run `aimx setup`
+# when the binary is already at the target version. Returns 0 on yes,
+# 1 on no / no usable TTY. Default is no (Enter = no), so non-interactive
+# callers (CI, fully-scripted) keep today's exit-0 semantics.
+#
+# Same TTY logic as ensure_sudo and the post-install handoff: prefer
+# the script's own stdin if it's already a terminal; fall back to
+# /dev/tty when it's a pipe (curl | sh); otherwise no prompt.
+_prompt_read() {
+    if [ -t 0 ]; then
+        read -r _ans
+    elif [ -e /dev/tty ] && [ -r /dev/tty ]; then
+        read -r _ans </dev/tty
+    else
+        return 1
+    fi
+}
+
+prompt_reinstall() {
+    _ans=""
+    printf '%s' 'AIMX is already installed. Re-run setup to (re)configure it? [y/N] ' >&2
+    _prompt_read || return 1
+    case "${_ans}" in
+        y | Y | yes | YES | Yes) return 0 ;;
+        *) return 1 ;;
+    esac
 }
 
 ensure_sudo() {
@@ -331,7 +359,7 @@ detect_os() {
     case "${_os}" in
         Linux) printf 'linux' ;;
         *)
-            err "aimx is Linux-only; detected ${_os}. See ${UNSUPPORTED_DOC}"
+            err "AIMX is Linux-only; detected ${_os}. See ${UNSUPPORTED_DOC}"
             ;;
     esac
 }
@@ -888,7 +916,7 @@ port_check_inbound() {
         # /probe can't tell aimx apart from Postfix/Sendmail/Exim, so a
         # green [ok] here is only meaningful if the holder is aimx.
         # Mirrors `aimx portcheck`'s Port25Status::OtherProcess handling.
-        ui_warn "port 25 is held by another process; verify it's aimx before running setup"
+        ui_warn "port 25 is held by another process; verify it's AIMX before running setup"
     else
         # Free or unknown: spawn a temp Python listener if available.
         if ! port_check_have_python3; then
@@ -1225,12 +1253,19 @@ main() {
             _cmp="$(compare_tags "${_installed_tag}" "${TAG}")"
             case "${_cmp}" in
                 equal)
-                    if [ "${FORCE}" -ne 1 ]; then
-                        say "aimx ${_installed_tag} is already installed (pass --force to re-install)"
+                    if [ "${FORCE}" -eq 1 ]; then
+                        say "re-installing ${TAG} (--force)"
+                        _is_upgrade=1
+                    elif prompt_reinstall; then
+                        # Binary is already correct; skip the swap and
+                        # jump straight to the wizard. backup_existing_config
+                        # below covers any partial config from an aborted run.
+                        _skip_swap=1
+                        _is_upgrade=0
+                    else
+                        say "AIMX ${_installed_tag} is already installed (pass --force to re-install)"
                         exit 0
                     fi
-                    say "re-installing ${TAG} (--force)"
-                    _is_upgrade=1
                     ;;
                 newer)
                     err "installed ${_installed_tag} is newer than target ${TAG}; run 'aimx upgrade --version ${TAG} --force' to downgrade explicitly"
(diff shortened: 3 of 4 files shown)


uzyn (commented):
## Changes Overview

This PR makes two related changes to `install.sh`:

1. When the binary already matches the target tag and `--force` was not passed, prompt `AIMX is already installed. Re-run setup to (re)configure it? [y/N]` instead of hard-exiting. On `y`, skip the download/extract/install step entirely and fall through to `backup_existing_config` + `exec aimx setup`. On `n` / Enter / no usable TTY, retain the existing exit-0 behavior.
2. Capitalize the brand name "AIMX" in operator-facing prose strings (install banner, port-check banner, help heading, "Linux-only" error, port-25 occupancy warn, "already installed" message, post-install/upgrade success lines, "Upgrade complete" sentence). Lowercase `aimx` is preserved for command literals, paths, service unit names, and the GitHub repo identifier.

## Scope Alignment

Implementation matches the PR description exactly. The new `prompt_reinstall` and `_prompt_read` helpers are appropriately scoped, the `_skip_swap=1` plumbing through the download/extract/install gates is consistent, and the equal-version yes-branch falls through to the same `backup_existing_config` + `exec aimx setup` handoff that fresh installs use. The brand sweep is bounded to user-facing prose strings — none of the touched lines contain command literals, paths, or repo identifiers.

## Potential Bugs

No correctness issues. I traced the equal-version flow end-to-end:

- `prompt_reinstall` returns 0 → `_skip_swap=1`, `_is_upgrade=0` → download/extract gate at install.sh:1306 skipped → upgrade gate at install.sh:1318 skipped → fresh-install gate at install.sh:1377 skipped (with `Re-running AIMX setup` info line) → `backup_existing_config` → `exec aimx setup`. Correct.
- `prompt_reinstall` returns 1 → `say "AIMX ... is already installed"` → `exit 0`. Correct.
- `--force` short-circuits to `_is_upgrade=1` before `prompt_reinstall` is called, so the prompt never fires for `--force`. Correct.

One minor cosmetic discrepancy worth flagging:

**Non-blocker — PR description claim about non-TTY is slightly incorrect.** The PR description's test plan says `curl ... | sh </dev/null` (no TTY) "exits 0 with the 'already installed' line, **no prompt printed**." But `prompt_reinstall` prints the prompt question to stderr (install.sh:270) **before** calling `_prompt_read`, which is what does the TTY check. So in the no-TTY case, stderr will show:

```
AIMX is already installed. Re-run setup to (re)configure it? [y/N] install: AIMX 0.0.7 is already installed (pass --force to re-install)
```

Functionally fine (still exits 0, no install), but cosmetically the prompt question appears in CI logs even though no answer is read. Either move the `printf` after the TTY-availability check inside `_prompt_read`, or update the PR description / commit message to remove the "no prompt printed" claim.

**Non-blocker — Dry-run now prompts interactively.** Pre-PR, the equal-version + no-force branch in dry-run mode `exit 0`'d immediately (at the old `say "aimx ... is already installed"` line). Post-PR, it calls `prompt_reinstall` first, which means `AIMX_DRY_RUN=1 sh install.sh` on a host where the binary is already at the target tag will now prompt the operator. Dry-run is conventionally non-interactive ("audit before running for real"). Consider gating the prompt on `[ "${DRY_RUN}" != "1" ]`, or accept the new behavior and document it. Not blocking — the dry-run still exits without touching the filesystem either way.

## Security Issues

None. `_ans` is constrained to literal case-matching (`y | Y | yes | YES | Yes`); no command interpolation, no path use, no shell expansion of the input. `read -r` correctly disables backslash interpretation. Prompt content is hardcoded.

## Test Coverage

The five new prompt assertions (`y`, `Yes`, `n`, empty, no-TTY) cover the core decision points. The overridable-stub design (`_prompt_read() { _ans="y"; }`) is a clean test seam — no `/dev/tty` mocking required. The three branding regressions are correctly anchored regexes that allow legitimate uses (paths like `aimx-0.0.7-...`, redirects like `aimx 2>/dev/null`).

**Non-blocker — three case-statement branches are untested.** The accept-list is `y | Y | yes | YES | Yes`. Tests cover `y` and `Yes`. The other three (`Y`, `yes`, `YES`) are matched by the same `case` arm so the risk of regression is low, but a single test asserting all five accept-list values would prevent a future contributor from quietly narrowing the match.

**Non-blocker — no integration test for the equal-version + yes path through `main()`.** All five new tests probe `prompt_reinstall` in isolation. The plumbing (`_skip_swap=1` skipping the download, skipping the install, falling through to backup + handoff) is verified by code inspection only. A `INSTALL_SH_TEST=1` test that stubs `_prompt_read`, `compare_tags`, `parse_installed_tag`, and `download` to drive the end-to-end equal-version yes branch and assert that no `download` call happened would close this. Not required given the code is straightforward, but would make the PR's intent regression-proof.

## Code Quality

Implementation is clean. Good separation: `_prompt_read` owns the TTY/stdin selection and is stubbed in tests; `prompt_reinstall` owns the prompt text and case match. POSIX-compliant throughout. Function comments explain the "why" (TTY logic, default-no rationale, CI compatibility). The `${_skip_swap:-0}` parameter expansion handles the unset case correctly.

The `_is_upgrade=0` line at install.sh:1264 is technically redundant (it's already initialized at install.sh:1249), but keeping it makes the branch's intent self-documenting. Fine either way.

## Documentation Gaps

**Non-blocker — `book/installation.md:38` documents the old behavior.** The line currently reads:

> "If an older `aimx` is already installed, the upgrade path runs instead: stop the service, swap the binary atomically, restart. No wizard re-run. **If the running version matches the target, it exits without touching anything (pass `--force` to reinstall).**"

This is now incorrect. The new behavior is: prompt to re-run setup (default no), with `--force` and non-TTY callers preserving the old exit-0 path. Operators reading the install guide will be misled. Suggested rewrite of the bolded sentence:

> "If the running version matches the target, the script asks whether to re-run `aimx setup` to (re)configure (default no, so CI / scripted callers keep exit-0 semantics). Pass `--force` to skip the prompt and reinstall the binary."

**Non-blocker — no release-notes entry.** `book/release-notes.md` has an active "Unreleased" section. This PR is an operator-visible behavior change (`curl | sh` re-runs now offer a recovery handle) and warrants an entry there, ideally as a sub-section of the existing "`install.sh` upgrade path is louder and self-healing" block at line 87 or as its own short block.

- **Overall verdict**: Needs minor fixes — implementation is correct and well-tested, but the install-flow documentation in `book/installation.md` now diverges from the script's behavior, and the release notes don't reflect the operator-visible change.

- **Blockers**: none.

- **Non-blockers** (should fix before merge):
  1. Update `book/installation.md:38` to describe the new prompt-on-equal-version behavior.
  2. Add a release-notes entry for the operator-visible behavior change.

- **Nice-to-haves** (low priority):
  3. Move the prompt `printf` inside `_prompt_read` after the TTY check (or update the PR description to drop the "no prompt printed" claim).
  4. Decide whether dry-run should be non-interactive on the equal-version branch and either gate `prompt_reinstall` on `DRY_RUN != 1` or accept the new behavior explicitly.
  5. Add explicit test cases for the `Y`, `yes`, `YES` accept-list values.
  6. Add an integration-style test driving the equal-version + yes path through `main()` to assert no `download` call happens.

uzyn:
## Re-review responses (commit 33336bd)

Pushed a follow-up commit addressing the review feedback. Summary of what changed and what was deliberately skipped.

### Non-blockers — both fixed
1. **`book/installation.md:38`** — rewrote the equal-version step in the install-flow numbered list. New text describes the prompt (`AIMX is already installed. Re-run setup to (re)configure it? [y/N]`), the default-no semantics, the non-TTY exit-0 path, and `--force` as the bypass.
2. **`book/release-notes.md`** — added an Unreleased entry under the existing "`install.sh` upgrade path is louder and self-healing" block, covering both the recovery prompt and the brand-prose capitalization sweep.

### Nice-to-haves — addressed (3, 4, 5, 6)
3. **Prompt placement** — moved the prompt `printf` from `prompt_reinstall` into `_prompt_read`, after the TTY-availability check. `curl | sh </dev/null` and other no-TTY callers now exit quietly with no stray prompt line in stderr / CI logs. The PR description's "no prompt printed" claim is now accurate.
4. **Dry-run gate** — `prompt_reinstall` is now short-circuited by `[ "${DRY_RUN}" != "1" ]` on the equal-version branch, so `AIMX_DRY_RUN=1` preserves pre-PR exit-0-immediately behavior. Auditing the script stays fully non-interactive.
5. **Accept-list test coverage** — replaced the two case-statement tests (`y`, `Yes`) with a `for _accept in y Y yes YES Yes; do ... done` loop that asserts every accept-list value returns YES.
6. **Dry-run regression test** — added a test that fakes an `aimx` binary at `${PREFIX}/aimx` (returning `aimx 0.1.0 ...`) and runs `AIMX_DRY_RUN=1 ... --tag 0.1.0`, asserting the "already installed" line prints and the recovery prompt does **not** fire.

### Also added
- Regression assertion that the prompt question is not printed when `_prompt_read` signals no-TTY (`assert_not_contains` against the prompt body).

### Implementation note on prompt placement
The cleanest fix was moving the `printf` into `_prompt_read` (it now takes the prompt text as `$1`) rather than duplicating the TTY-check predicate in two places. This means the existing test stub `_prompt_read() { _ans="..."; }` no longer prints the prompt — so the previous "prompt_reinstall prints the prompt to stderr" assertion was removed (it was testing an implementation detail of where the printf lived, not behavior). The new "stays quiet on no TTY" assertion gives stronger regression coverage in its place.

### Test results
`sh tests/install_sh.sh` → 107/109 passing (+5 vs. pre-fix; net of the removed implementation-detail assertion). The two failures (`shellcheck install.sh`, `shellcheck tests/install_sh.sh`) are the same pre-existing issues already disclosed on the original PR description (SC2024 + SC2010/SC2016 in the harness) and not gated by CI.

Ready for re-review.

uzyn:
## Follow-up: clean shellcheck (commit c096374)

Pushed a small follow-up addressing the 2 shellcheck failures the original PR description had disclosed as pre-existing. After investigating I confirmed all 24 diagnostics are false positives or controlled-context cases:

| Rule | Sites | Why it's a false positive |
|---|---|---|
| **SC2024** (install.sh × 2) | `(sudo -v </dev/tty)` in `ensure_sudo` and `port_check_ensure_inbound_privilege` | Provides a tty for sudo's password prompt under `curl|sh`. Not redirecting a privileged file *through* sudo into the inner command — that's the antipattern SC2024 actually targets. |
| **SC2016** (tests × 20) | Single-quoted `printf` literals carrying backticks / `${…}` for human display, plus grep patterns carrying escaped `\${…}` | Strings MUST stay literal — converting to double quotes would break either the test output or the regex pattern. |
| **SC2010** (tests × 2) | `ls "${_tmp}/etc/aimx/" \| grep '^config.toml.bak-'` in two backup-detection sites | Operates on a test-controlled tmpdir whose filenames are guaranteed `config.toml.bak-<utc>-<pid>` — the "non-alphanumeric filename" hazard SC2010 warns about cannot occur here. |

Annotation strategy:
- **install.sh** — per-site `# shellcheck disable=SC2024` with a one-line inline reason.
- **tests/install_sh.sh** — file-level `# shellcheck disable=SC2016` (cleaner than 20 per-line directives for an idiom that's consistent across the file). Per-site `# shellcheck disable=SC2010` with a reason at the two `ls | grep` lines.

### Test-side future-proofing

Inserting the SC2024 directive line above `(sudo -v </dev/tty)` broke a structural assertion that read the literal previous line and expected `elif [ -e /dev/tty ]`. I introduced a small `_nearest_code_line` helper that walks back over comments and blank lines within a 10-line window, and updated both the redirected-`exec` handoff test and the `sudo -v` test to use it. Future shellcheck directives or clarifying comments above either matched line no longer break the gating check.

### Verification

- `shellcheck install.sh` → no output, exit 0.
- `shellcheck -s sh tests/install_sh.sh` → no output, exit 0.
- `sh tests/install_sh.sh` → **109/109 tests passed** (no `FAIL:` line).
- `sh -n install.sh && sh -n tests/install_sh.sh` → POSIX-clean.

PR description updated to reflect the new 109/109 baseline and to drop the "two failures are pre-existing" disclaimer.

uzyn (commented):
## Re-review (commits 33336bd + c096374)

Verified each previously-flagged item against the current code at `c096374`. All addressed; no new issues introduced.

### Previous non-blockers — both resolved

1. **`book/installation.md:38`** — Step 9 of the install-flow numbered list now describes the recovery prompt, default-no semantics, non-TTY exit-0 path, and `--force` bypass. Matches the script's behavior exactly.
2. **`book/release-notes.md`** — New Unreleased sub-section after the existing `install.sh` upgrade-path block covers both the recovery prompt and the brand capitalization sweep. Operator-visible behavior change is now documented in the changelog.

### Previous nice-to-haves — all addressed

3. **Prompt placement** — The `printf` moved into `_prompt_read` (`install.sh:262, 265`), gated behind the same TTY-availability checks that govern `read`. The no-TTY return path at line 268 short-circuits before any output, so `curl | sh </dev/null` no longer leaks the prompt into stderr. Cleaner factoring than duplicating the predicate. The new regression assertion (`tests/install_sh.sh:1292` `assert_not_contains "prompt_reinstall stays quiet on no TTY"`) locks this in.
4. **Dry-run gate** — `[ "${DRY_RUN}" != "1" ] && prompt_reinstall` at `install.sh:1264` short-circuits the prompt, falling straight through to the canonical `else` arm (`exit 0`). I traced this: `AIMX_DRY_RUN=1` + equal version + no `--force` reaches the `say "AIMX ... is already installed"` branch and exits without ever calling `prompt_reinstall` or printing the dry-run plan. That matches the intent (auditing must stay non-interactive).
5. **Accept-list coverage** — `for _accept in y Y yes YES Yes` loop replaces the two case-arm tests. A future contributor narrowing the accept list will now break a test rather than silently ship.
6. **Equal-version + dry-run integration test** — New test at `tests/install_sh.sh:1216-1237` fakes a target-tag `aimx` at `${PREFIX}/aimx`, sets `AIMX_DRY_RUN=1`, and asserts both that "is already installed" prints and that `prompt_reinstall` does NOT fire. Exercises the gate end-to-end through `main()`, not just `prompt_reinstall` in isolation.

### Bonus: shellcheck cleanup (commit c096374)

The implementer also took care of the two pre-existing shellcheck failures the original PR description had disclosed:

- **SC2024 × 2** (`install.sh`) — `(sudo -v </dev/tty)` annotated per-site with the rationale (subshell + `</dev/tty` is providing a TTY for sudo's password prompt under `curl|sh`, not the shellcheck-targeted antipattern of redirecting a privileged file *through* sudo into the inner command). Reasoning is correct.
- **SC2016 × 20** (`tests/install_sh.sh`) — file-level disable with a top-of-file comment explaining the idiom (single-quoted `printf` literals carrying `${...}` for human display, grep patterns carrying escaped `\${...}`). Per-line directives would have been noisy for an idiom this consistent — file-level is the right call.
- **SC2010 × 2** (`tests/install_sh.sh`) — per-site disable on the two `ls "${_tmp}/etc/aimx/" | grep '^config.toml.bak-'` sites. Tmpdir is test-controlled and filenames follow the predictable `config.toml.bak-<utc>-<pid>` pattern, so the SC2010 hazard cannot fire here.

Verification on my end:
- `sh tests/install_sh.sh` → **109/109 passing**.
- `shellcheck install.sh` → exit 0, no output.
- `shellcheck -s sh tests/install_sh.sh` → exit 0, no output.

The new `_nearest_code_line` helper (`tests/install_sh.sh:222-238`) is a sensible response to the brittleness in the previous structural assertions: walking back over comments and blanks within a 10-line window means future shellcheck directives or clarifying comments above either matched line won't break the gating check. Pattern matching is correct (`''|[[:space:]]*'#'*|'#'*` cleanly identifies blanks, indented comments, and top-level comments without false-matching ordinary indented code).

### End-to-end flow verification (re-traced from the current code)

- `--force` set → `_is_upgrade=1` → upgrade path → binary swap + service restart. `prompt_reinstall` never called.
- `--force` unset, `DRY_RUN=1`, equal version → `[ "${DRY_RUN}" != "1" ]` short-circuits → `else` arm → `exit 0`. No prompt, no dry-run plan output (which is fine — pre-PR exited 0 here too).
- `--force` unset, `DRY_RUN=0`, equal version, prompt yes → `_skip_swap=1`, `_is_upgrade=0` → download/extract gate at `install.sh:1314` skipped → upgrade gate at `install.sh:1326` skipped → fresh-install gate at `install.sh:1385` skipped (prints `Re-running AIMX setup (binary unchanged)`) → `backup_existing_config` → `exec aimx setup`.
- `--force` unset, `DRY_RUN=0`, equal version, prompt no / no TTY → `prompt_reinstall` returns 1 → `else` arm → `exit 0`.
- `--force` unset, older version → `_is_upgrade=1` → upgrade path. Unaffected by this PR.
- Binary missing entirely → outer `if [ -x "${_install_path}" ]` skipped → `_is_upgrade=0`, `_skip_swap` unset → fresh install + `exec aimx setup`. Unaffected.

`${_skip_swap:-0}` parameter expansion handles the unset case under `set -u`. All branches verified.

### No new issues

- No regressions in the non-equal paths (older / newer / unparseable / missing-binary / dry-run-with-force).
- Brand sweep is consistent: `grep -nE '(installing|installed|upgraded) aimx\b|\baimx is\b|aimx [0-9]+\.[0-9]'` against `install.sh` returns zero hits (asserted by the three branding tests).
- Help text capitalization (`AIMX install script` at install.sh:170) matches the test assertion at `tests/install_sh.sh:729`.
- No banned planning-doc cross-references introduced (verified `Sprint`, `FR-`, `S\d-\d` patterns).

- **Blockers**: none.
- **Non-blockers**: none.
- **Nice-to-haves**: none.

```
install.sh: prompt to re-run setup when AIMX is already installed + capitalize brand prose

When `curl ... | install.sh` finds the binary already at the target tag and `--force` was not passed, ask "AIMX is already installed. Re-run setup to (re)configure it? [y/N]" instead of hard-exiting. On `y`, skip the download/extract/install entirely and fall through to backup_existing_config + `exec aimx setup` (the same handoff fresh installs use). On `N` / Enter / no usable TTY (CI, scripted callers) / `AIMX_DRY_RUN=1`, keep the existing exit-0 semantics. `--force` continues to bypass the prompt.

Capitalize the brand name "AIMX" in operator-facing prose (install banner, port-check banner, help heading, "Linux-only" error, port-25 warn, "already installed" message, post-install/upgrade success lines, "Upgrade complete" sentence). Lowercase `aimx` is preserved for command literals, paths, service unit names, and the GitHub repo identifier.

Also silence pre-existing shellcheck false positives in `install.sh` and `tests/install_sh.sh` (SC2024, SC2016, SC2010) so the test suite reaches 109/109 with `shellcheck install.sh` and `shellcheck -s sh tests/install_sh.sh` both clean.

Docs:
- book/installation.md: rewrote step 9 of the install-flow list to describe the new prompt-on-equal-version behavior.
- book/release-notes.md: added an Unreleased entry covering both the recovery prompt and the brand capitalization sweep.
```

PR #206, opened 2026-05-07
Epic: Markdown-Native Outbound (Rich-Text Email)

uzyn:
## Epic Integration PR

**Track:** html-email
**PRD:** docs/html-email-prd.md
**Sprint plan:** docs/html-email-sprint.md

This PR is the integration point for the PRD above. Sprint PRs are merged into
`epic/html-email` one at a time by `/scrum-autopilot`. When all sprints are done,
this PR is ready for human review and merge to `main`. Autopilot will not
merge it for you — that gate is yours.

### What this PRD ships

Markdown becomes the canonical send format for AIMX. `aimx send --body` will
interpret its argument as Markdown by default; the daemon renders to HTML with
an inlined stylesheet and emits `multipart/alternative` (Markdown source as
text part, rendered HTML as HTML part). Two escape hatches: `--text-only` for
transactional plain-text sends and `--html-body` for callers with custom HTML
layouts. Sent storage stays single-file `.md` (Markdown source as the body);
the recipient's rendered view is recoverable by re-running the same renderer.

This is a hard break to `--body` semantics — release notes call it out.

### Sprints

- [ ] Sprint 1 — Markdown renderer + inlined stylesheet
- [ ] Sprint 2 — Wire shape + default send path
- [ ] Sprint 3 — CLI escape hatches + AIMX/1 SEND frame extension
- [ ] Sprint 4 — MCP + sent storage + docs + release notes

### How to review

- Read the PRD and sprint plan above for context.
- Each sprint already had its own PR with line-level review; this PR is for
  integration-level concerns (cross-sprint coherence, MIME structure choices,
  renderer determinism guarantees, release-notes wording).
- The diff against `main` here is the *full* PRD, not just the last sprint.


uzyn (commented):
# Review: Epic Markdown-Native Outbound (PR #206)

## Changes Overview

This is the integration PR for the html-email epic, bundling four sprint PRs that turn AIMX's outbound surface Markdown-native end-to-end. `aimx send --body` now renders CommonMark+GFM to HTML inside the daemon (not the client), shipping a `multipart/alternative` with the Markdown source as the text part and rendered HTML as the html part. Two clap-level mutually exclusive escape hatches — `--text-only` and `--html-body` — are layered on the SEND wire (`Text-Only:` / `Html-Body-Length:` headers). MCP `email_send` / `email_reply` mirror the CLI surface. Sent records remain single-file `.md` and gain an `outbound_format` audit field; the renderer enforces a 5 MB body cap with a dedicated `BODY_TOO_LARGE` wire code. Total diff: +8 807 / −324 across 24 files (the bulk being two new modules — `markdown_render.rs` 522 LoC, `wire_assembly.rs` 1 632 LoC — plus checked-in renderer fixtures and ~9 new end-to-end integration tests).

## Scope Alignment

Strong. Every promise in the PR description is delivered:

- Markdown render path with comrak + lol_html inlined styling ✓
- Daemon-side wire assembly with default + nested `multipart/mixed`-with-attachments shape ✓
- `--text-only` single-part `text/plain` (with the recent CTE-in-header-block fix) ✓
- `--html-body` operator-supplied verbatim HTML alongside `--body` text fallback ✓
- Mutual exclusion at clap, codec, and MCP layers with one canonical error string ✓
- Sent storage stays single-file `.md`, no `.html` sibling, audit field stamped ✓
- 5 MB cap with dedicated `ERR BODY_TOO_LARGE` wire code and canonical actionable message ✓
- DKIM verification preserved on all three paths (asserted by end-to-end integration tests) ✓
- Hard-break advisory in release notes; new `book/markdown-email.md` chapter ✓
- Version bumped 0.1.0 → 0.2.0 ✓

The recent (post-Sprint-4) CTE placement fix (commit 611383b) is a substantive correctness improvement on top of the original sprint scope and properly carries its own regression tests.

## Potential Bugs

### 1. `extract_quoted_param` does not decode backslash-escaped quotes in filenames — Non-blocker

**Location:** `src/wire_assembly.rs:404-432`

The client's `compose_request` in `src/send.rs` runs every attachment filename through `escape_filename`, which replaces `"` with `\"` to keep the wire RFC 5322-clean. The new daemon-side `extract_quoted_param` parser, however, terminates on the very first `"` inside a quoted value:

```rust
if let Some(stripped) = value.strip_prefix('"') {
    let end = stripped.find('"')?;
    return Some(stripped[..end].to_string());
}
```

For an attachment whose filename contains a `"` character, the wire carries `filename="my\"file.pdf"`. After `strip_prefix('"')`, the parser scans for the next `"` and finds the escaped one at index 3, returning `my\` instead of `my"file.pdf`. The recipient receives the attachment under a truncated filename ending in a literal backslash.

This is a regression specifically introduced by this PR: in v0.1 the daemon never parsed the multipart body — the client shipped the final wire shape verbatim. With the new daemon-side reassembly, the parser owns this round-trip, and it does not match the client-side escape contract.

`tests/fixtures` and `tests/integration.rs` cover normal filenames; `attachment_filename_with_quotes_escaped` in `src/send.rs` only verifies the client-side escape, not the daemon round-trip. A new test that builds a multipart request with `filename="my\"file.pdf"` and asserts the daemon-emitted wire still carries the same escaped filename would pin this. Fix is small (decode `\"` and `\\` while scanning for the closing quote, or fail the parse if such a sequence is encountered).

Classified Non-blocker because filenames with embedded `"` are pathological in practice and the attachment bytes themselves still deliver correctly. Worth fixing or documenting as a known limitation.

## Security Issues

No critical or high-severity issues. Quick audit:

- **HTML escaping:** `comrak::Options::r#unsafe = false` plus the GFM `tagfilter` strip raw HTML and `<script>` from Markdown bodies. `raw_html_in_markdown_is_neutralised` pins the invariant.
- **`--html-body` bypass:** documented explicitly as operator-owned (`book/markdown-email.md` line 110, `mcp.rs` schema doc). Acceptable given the surface (local CLI / authenticated MCP).
- **Header-injection:** `validate_header_value` in `send.rs` rejects CRLF in `From` / `To` / `Subject` / `reply_to` / `references` before any wire bytes are emitted. Existing tests cover this.
- **Body-size DoS:** the 5 MB cap fires *before* comrak runs (the `oversize_body_does_not_invoke_comrak` test proves this with an `AtomicUsize` invocation counter under a probe `RwLock` — clever, defended against future drift).
- **External resources in rendered HTML:** `no_external_resources_in_output` test asserts no `<style>`, no `<link>`, no `googleapis`, no remote `http://` URIs are injected.
- **Renderer determinism:** `comrak` pinned to `=0.52.0` with checked-in fixture HTML. Drift surfaces at CI, with an `AIMX_BLESS_EXPECTED=1` escape hatch for deliberate bumps.

## Test Coverage

Coverage is genuinely strong, not just "more tests":

- **20 unit tests** in `markdown_render.rs` — GFM extensions, raw-HTML neutralization, autolink styling, body-cap boundary + over-cap with comrak-non-invocation proof, fixture determinism, fixture drift help, ≤25 KB rendered briefing budget, release-only perf bounds.
- **31 unit tests** in `wire_assembly.rs` — every wire-shape branch, signature behavior on each path, multipart parser delimiter-awareness (including the `name=` vs `filename=` substring trap, captured by `extract_quoted_param_does_not_match_name_inside_filename`), CTE placement on text-only single-part wire (header block, not body) with both 7bit and quoted-printable variants, `mail-parser` round-trip on all three paths, `BodyTooLarge` propagation through the assembler with the canonical message asserted verbatim.
- **40 unit tests** in `send_handler.rs` covering the new `outbound_format` plumbing per path and pinning that no `.html` sibling is ever written.
- **20 codec tests** for the AIMX/1 SEND frame — round-trip default, text-only, html-body shapes; mutual-exclusion canonical message; duplicate / invalid header rejection.
- **9 new end-to-end integration tests** covering `aimx send` UDS path with each shape, DKIM body-hash verification, attachment + Markdown nesting, MCP `email_send` / `email_reply` mutual-exclusion, and CLI clap-level rejection.

The `oversize_body_does_not_invoke_comrak` test is unusually careful — using a write-lock on `COMRAK_PROBE_LOCK` to block concurrent renderer threads from ticking the counter mid-measurement. That kind of defensive design is what gates regressions reliably.

The one notable gap is the round-trip test for filenames containing `"` (see Bug #1 above).

## Code Quality

The two new modules — `markdown_render` and `wire_assembly` — are clean pure transforms with explicit error types. Module-level docs explain *why* each branch exists (signature appending order, lock hierarchy, multipart attach unpacking). The `pick_text_only_top_level_encoding` + `debug_assert!` defensive pattern in `wire_assembly.rs:600-604,634-638` is a nice "fail loudly in debug, degrade gracefully in release" guard against the header builder and body builder picking different transfer encodings.

Minor observation, non-blocker:

- **`FR-F2` / `FR-F3` / `FR-F5` references introduced** in `src/send_handler.rs:1838`, `tests/integration.rs:2922,3315`. CLAUDE.local.md's banned-token regex (`\bFR-\d+[a-z]?\b`) does not catch letter-prefixed FR ids, and the codebase already carries pre-existing `FR-B2` references, so there's existing precedent and no CI gate fires. The spirit of "no planning-doc cross-references outside `docs/`" applies, though, and these are easy to drop without losing technical content (e.g. `(FR-F3)` → no parens). Worth a follow-up sweep.
- The most recent commit's title "html-email cleanup: address Sprint 4 review non-blockers" uses the literal "Sprint 4" string, and the PR description lists `Sprint 1 — ... Sprint 4 — ...` checkboxes. CLAUDE.local.md extends the ban to commit messages and PR descriptions; for an epic integration PR this is hard to avoid entirely, but worth noting.

**Overall verdict:** Needs minor fixes. The integration is technically solid, the test coverage is unusually thorough, and the documentation is excellent. The one identified bug is an edge-case parser limitation worth either fixing or documenting before this lands on `main`.

**Blockers:** None.

1. `extract_quoted_param` does not decode `\"` in quoted parameter values — filenames with embedded `"` are truncated by the daemon's multipart parser. Add a round-trip test and either decode the escape or document the limitation.

2. Drop or rephrase the new `FR-F2`, `FR-F3`, `FR-F5` references in `src/send_handler.rs` and `tests/integration.rs` to keep the spirit of the planning-doc-references rule.
3. The PR description and recent commit message use literal "Sprint" tokens; a future cleanup pass could rewrite to the technical-content-only form per CLAUDE.local.md.

uzyn (commented):
# Re-review: Epic Markdown-Native Outbound (PR #206)

This is a fresh review against the PR description (no PRD/sprint files are tracked in git — `docs/` is gitignored — so this is a standalone review). It also acknowledges the prior author review (@uzyn) and verifies which items it raised have been addressed on the current head.

## Changes Overview

PR #206 is the integration PR for the html-email epic, bundling four sprint PRs that turn AIMX's outbound surface Markdown-native end-to-end. `aimx send --body` now ships unsigned Markdown to the daemon over UDS; the daemon renders to HTML via `comrak` + an inlined stylesheet (`lol_html`) and emits a `multipart/alternative` (with Markdown source as the text part and rendered HTML as the html part). Two clap-level mutually exclusive escape hatches — `--text-only` and `--html-body` — are layered on the SEND wire (`Text-Only:` / `Html-Body-Length:` headers). MCP `email_send` / `email_reply` mirror the CLI surface. Sent records remain single-file `.md` and gain an `outbound_format` audit field; the renderer enforces a 5 MiB body cap with a dedicated `BODY_TOO_LARGE` wire code. Total diff: +8 917 / −324 across 24 files.

Two follow-up commits sit on top of the four sprint commits:
- `611383b` — fixed CTE placement on the text-only single-part wire (header block, not body).
- `b3971ef` — addresses the prior reviewer's `extract_quoted_param` non-blocker (decode `\"` and `\\` quoted-pair escapes; reject unterminated escapes) AND drops the `FR-F2` / `FR-F3` / `FR-F5` planning-doc references the prior reviewer flagged.

## Scope Alignment

Strong. Every promise in the PR description is delivered, and the recent escape-hatch + CTE fixes are substantive correctness improvements on top of the original sprint scope, both with their own regression tests.

## Verification of prior reviewer's bug fix (commit b3971ef)

I traced the new `extract_quoted_param` (`src/wire_assembly.rs:407-462`) and the helper `decode_quoted_value` against the prior reviewer's failure case:

- Input `application/pdf; filename="my\"file.pdf"` → returns `Some("my\"file.pdf")`. ✓
- Input `application/pdf; filename="a\\b.pdf"` → returns `Some("a\\b.pdf")`. ✓
- Trailing-lone-`\` (`filename="trailing\"`) → returns `None` rather than truncating or buffer-overrunning. ✓
- New end-to-end `attachment_filename_with_escaped_quote_round_trips_through_assembler` (`src/wire_assembly.rs:1339-1376`) pins the contract between client-side `escape_filename` and daemon-side parser.

The prior non-blocker is genuinely resolved.

## Potential Bugs

### 1. `extract_quoted_param` still mishandles `;` inside quoted parameter values — Non-blocker

**Location:** `src/wire_assembly.rs:407-434`

The new parser fixes the `\"` case but leaves a sibling case open. The parser pre-splits on `;` *before* it even looks at the quote characters:

```rust
let after_primary = s.split_once(';')?.1;
for segment in after_primary.split(';') {
    ...
    let value = raw_value.trim_start();
    if let Some(stripped) = value.strip_prefix('"') {
        return decode_quoted_value(stripped);
    }
```

For an attachment whose filename contains a literal `;`, the wire carries `filename="hello;world.pdf"`. The `after_primary.split(';')` splits this into two segments: `filename="hello` and `world.pdf"`. The first segment hits `decode_quoted_value("hello")`, which never finds a closing `"` and returns `None`. The end result: `extract_filename` returns `None` and `parse_attachment_part` falls back to the literal string `"attachment"` for the filename. The recipient receives the attachment under the wrong filename.

This is structurally the same bug class as the `\"` case the prior reviewer flagged. The client's `escape_filename` (`src/send.rs:33-38`) does **not** escape `;` (and there's no standard escape via backslash for `;` inside quoted strings — the proper fix is RFC 2045 tokenisation on the daemon side, recognising quoted strings as atomic before splitting on `;`).

Classified Non-blocker because (a) filenames with embedded `;` are rare in practice, (b) the attachment bytes themselves still deliver correctly — only the filename header is wrong, and (c) no crash or signing failure. Worth fixing or documenting alongside the `\"` case as a known parser limitation. A round-trip test analogous to `attachment_filename_with_escaped_quote_round_trips_through_assembler` would pin the fix.

## Security Issues

No critical or high-severity issues. Quick audit of what changed:

- **HTML escaping:** `comrak::Options { r#unsafe = false, tagfilter = true, ... }` strip `<script>` / `<iframe>` and escape arbitrary raw HTML. `raw_html_in_markdown_is_neutralised` (`src/markdown_render.rs:215-244`) pins the invariant.
- **`--html-body` bypass:** documented explicitly as operator-owned (`book/markdown-email.md` and the MCP schema doc on `email_send` / `email_reply`). Acceptable given the surface (local CLI / authenticated MCP).
- **Header injection:** `validate_header_value` in `src/send.rs:44-51` rejects CRLF in `From` / `To` / `Subject` / `reply_to` / `references` before any wire bytes are emitted.
- **Body-size DoS:** the 5 MiB cap fires *before* `comrak` runs (the `oversize_body_does_not_invoke_comrak` test pins this with an `AtomicUsize` invocation counter under a write-locked `RwLock<()>` probe — a notably careful design that would catch a future regression where the cap moved past the renderer call).
- **External resources in rendered HTML:** `no_external_resources_in_output` asserts no `<style>`, `<link>`, `googleapis`, or remote `http://` URIs are injected.
- **Renderer determinism:** `comrak` pinned to `=0.52.0` with checked-in fixture HTML. Drift surfaces at CI with an `AIMX_BLESS_EXPECTED=1` escape hatch for deliberate bumps.

## Test Coverage

Coverage is thorough, not just numerous:

- **Renderer (`markdown_render.rs`)**: 20 unit tests — GFM extensions, raw-HTML neutralisation, autolink styling, body-cap boundary + over-cap with comrak-non-invocation proof, fixture determinism, fixture drift help, ≤25 KB rendered-briefing budget, release-only perf bounds.
- **Wire assembly (`wire_assembly.rs`)**: 31 unit tests — every wire-shape branch, signature behavior on each path, multipart parser delimiter-awareness (including the `name=` vs `filename=` substring trap and the new `\"` / `\\` escape decoding), CTE placement on the text-only single-part wire, `mail-parser` round-trip on all three paths.
- **Codec (`send_protocol.rs`)**: 20 codec tests — round-trip default, text-only, html-body shapes; mutual-exclusion canonical message; duplicate `Text-Only:`; invalid `Text-Only:` value; html-body length mismatch.
- **`send_handler.rs`**: 40 unit tests covering the new `outbound_format` plumbing per path and pinning that no `.html` sibling is ever written.
- **9 new end-to-end integration tests** covering `aimx send` UDS path with each shape (default Markdown, text-only, html-body), DKIM body-hash verification on each, attachment + Markdown nesting, MCP `email_send` / `email_reply` mutual-exclusion, and CLI clap-level rejection.

### Test gaps

- **No round-trip test for attachment filenames containing `;`** — see Bug #1 above.
- **No `duplicate_html_body_length_header_rejected` test.** The codec rejects this (`src/send_protocol.rs:580-583`) but only `Text-Only:` has a duplicate-header test (`duplicate_text_only_header_rejected`). Symmetric coverage would prevent a regression where the duplicate-check was accidentally removed from one of the two headers. Non-blocker but cheap to add.

## Code Quality

The two new modules — `markdown_render` and `wire_assembly` — are clean pure transforms with explicit error types. Module-level docs explain *why* each branch exists. Two design choices worth calling out:

- **`pick_text_only_top_level_encoding` + `debug_assert!` in `build_text_only_body`** (`src/wire_assembly.rs:631-668`) — the header builder and body builder agree on the encoding via a pure helper, with a debug-only assertion that catches drift if a future change splits the two computations.
- **`COMRAK_PROBE_LOCK` + `COMRAK_INVOCATIONS`** (`src/markdown_render.rs:75-93`) — the body-cap test takes an exclusive write-lock to block concurrent renderer threads from ticking the counter mid-measurement, then sanity-checks that an under-cap call DOES tick it under the same window. That's an unusually careful invariant test.

Minor observations, all non-blocker:

- The error string `"markdown body exceeds 5MB; ..."` (`src/markdown_render.rs:39-41`) uses decimal-MB phrasing while `MAX_MARKDOWN_BODY_BYTES = 5 * 1024 * 1024` is binary MiB. The discrepancy is small (~5%) and matches industry-standard sloppy terminology, but is technically inconsistent.
- `decode_quoted_printable` in `src/wire_assembly.rs:474-501` requires `i + 2 < bytes.len()` to recognise a `=` escape, which means a trailing `=\n` soft break at exactly the last 2 bytes of input is preserved as literal `=\n` rather than removed. This is unreachable in practice (text bodies always end with a CRLF or LF after content, not bare `=\n`), but worth noting if anyone ever generalises this codec for other inputs.

## Human Review Comments

One prior review by @uzyn (PR author) on commit `611383b`, classified verdict "Needs minor fixes", with one Non-blocker and two nice-to-haves. The current head is `b3971ef`:

1. **`extract_quoted_param` does not decode backslash-escaped quotes** — **Valid, addressed**. Commit `b3971ef` decodes RFC 2822 quoted-pair escapes, rejects unterminated escapes, and adds three regression tests including an end-to-end round-trip through `assemble_wire_message`. Verified by re-tracing the parser; see "Verification of prior reviewer's bug fix" above.

2. **`FR-F2` / `FR-F3` / `FR-F5` references in `src/send_handler.rs` and `tests/integration.rs`** — **Valid, addressed**. Commit `b3971ef` also dropped these alongside the parser fix; `git show origin/epic/html-email -- src tests | grep 'FR-F'` returns nothing on the current head.

3. **`Sprint N` literals in PR description and recent commit message** — **Valid, not addressed**. Inevitable for an epic integration PR (the description structurally needs them); could be revisited in a future cleanup pass. Nice-to-have.

No comments exist on any linked GitHub issue (the PR description does not reference one).

The integration is technically solid, the test coverage is unusually thorough, and the documentation is excellent. The prior reviewer's only bug-class concern is fixed, and one of the two nice-to-haves was cleaned up alongside it. The single new finding is an edge-case parser limitation in the same bug class as the prior reviewer's `\"` fix; worth either resolving structurally or documenting before this lands on `main`.

**Blockers:** None.

1. `extract_quoted_param` still mishandles `;` inside quoted parameter values (e.g. `filename="hello;world.pdf"`). Same bug class as the `\"` case fixed in `b3971ef`; the structural fix is RFC 2045 tokenisation that recognises quoted strings as atomic before splitting on `;`. Add a round-trip test analogous to `attachment_filename_with_escaped_quote_round_trips_through_assembler`.

2. Add `duplicate_html_body_length_header_rejected` codec test to mirror the existing `duplicate_text_only_header_rejected` so a regression on either duplicate check surfaces independently.

3. (Carry-over from prior review.) Future cleanup pass to rephrase literal `Sprint N` tokens in PR description and recent commit messages to keep the spirit of CLAUDE.local.md's planning-token rule.

4. Consider rephrasing the `BodyTooLarge` error message to match the binary-MiB constant ("5 MiB" instead of "5MB") for technical accuracy.

uzyn (commented):
# Re-review #2: Epic Markdown-Native Outbound (PR #206)

Verifying commit `75f4ccf html-email cleanup: quote-aware param split, dup-header test, 5 MiB normalization` against the three actionable items from my previous review.

## Verification of the previous review's findings

### 1. `;` inside quoted parameter values — Fixed ✓

`src/wire_assembly.rs:407-486` introduces `split_params_quote_aware` and rewires `extract_quoted_param` to use it. I traced the new helper against every case I had in mind:

- `application/pdf; filename="hello;world.pdf"` → segments `["application/pdf", " filename=\"hello;world.pdf\""]`. Returns `Some("hello;world.pdf")`. ✓
- `application/pdf; filename="a;b;c.pdf"; size=42` → segments `["application/pdf", " filename=\"a;b;c.pdf\"", " size=42"]`. Both `filename` and `size` resolve correctly. ✓
- `application/pdf; filename="my\"file.pdf"` → backslash-escape branch keeps the inner `"` from closing the quoted string; returns `Some("my\"file.pdf")` via `decode_quoted_value`. ✓ (Existing `\"` regression behavior preserved.)
- `application/pdf; filename="trailing\"` → `in_quote` is still true at end-of-input; the helper pushes the unterminated remainder as one segment, which `decode_quoted_value` then rejects with `None`. ✓ (Unterminated-escape behavior preserved.)
- `application/pdf; filename="k=v.pdf"` → inner `=` stays inside the quoted segment; `split_once('=')` on the segment finds the outer `=`. Returns `Some("k=v.pdf")`. ✓

The implementation is structurally correct: single linear walk, explicit `in_quote` / `escape_next` state, byte-safe slice indexing via `char_indices()`. The escape-next branch deliberately fires only when `in_quote` (so backslashes outside quotes are inert per RFC 2045).

Test coverage on the new helper (`src/wire_assembly.rs:1380-1465`):

- `extract_quoted_param_handles_semicolon_inside_quoted_value` — basic case
- `extract_quoted_param_handles_multiple_semicolons_inside_quote_with_trailing_param` — multiple `;` inside one quoted value plus a subsequent unquoted parameter
- `extract_quoted_param_handles_equals_inside_quoted_value` — guards the `=`-inside-quote case
- `attachment_filename_with_semicolon_round_trips_through_assembler` — end-to-end through `assemble_wire_message`, including a negative assertion that the daemon does not silently substitute the `"attachment"` fallback filename

### 2. `duplicate_html_body_length_header_rejected` — Fixed ✓

`src/send_protocol.rs:1872-1890` mirrors the existing `duplicate_text_only_header_rejected` line-for-line (ditto frame structure, ditto assertion shape, ditto failure mode). The codec already enforced this branch (`src/send_protocol.rs:580-583`); the test now pins it.

### 3. 5 MiB normalisation — Fixed ✓

I grepped every 5MB / 5 MiB occurrence on the head:

- `src/markdown_render.rs`: canonical `Display` message + matching test assertion + section header — all `5 MiB`.
- `src/wire_assembly.rs`: doc comment + `body_too_large_propagates_canonical_message_through_assembler` assertion — `5 MiB`.
- `src/send_protocol.rs`: `BodyTooLarge` enum doc — `5 MiB`.
- `book/markdown-email.md`: chapter narrative now spells out `MAX_MARKDOWN_BODY_BYTES = 5 * 1024 * 1024` so a reader sees why the unit is binary; receiver-side caps (Gmail 25 MB, Outlook 20 MB, M365 150 MB) are intentionally left in MB matching how those services document themselves.
- `book/release-notes.md` + `agents/common/references/mcp-tools.md`: `5 MiB`.

The "decimal vs. binary" inconsistency I flagged is gone, and the decision to leave external receiver caps as MB is sound.

## New issues introduced by this commit

None. The change is local to `extract_quoted_param`, the new `split_params_quote_aware` helper has no other call sites, and the MiB rename is a pure string change with no behavioral risk.

All three actionable items from the previous re-review are correctly implemented with thorough test coverage. No new findings, no carry-overs other than the prior nice-to-have about `Sprint N` literals in the PR description (inevitable for an epic integration PR; not actionable here).

```
Markdown-native outbound: render Markdown to HTML on the daemon, ship multipart/alternative

Hard break to `aimx send --body` semantics: the body is now interpreted
as CommonMark + GFM Markdown by default. The daemon renders to HTML with
an inlined per-element stylesheet (no <style> blocks, no remote
resources) and emits a multipart/alternative with the Markdown source as
text/plain and the rendered HTML as text/html. Two clap-level mutually
exclusive escape hatches: --text-only ships the body verbatim as
single-part text/plain, --html-body ships operator-supplied HTML
verbatim alongside --body as the text fallback. MCP email_send /
email_reply mirror the surface with text_only and html_body parameters
and a server-side mutual-exclusion check using the codec's canonical
wording.

Sent records remain single-file .md (the recipient's HTML view is
recoverable by re-running the deterministic renderer on the stored
Markdown). New `outbound_format` audit field stamps "markdown" / "text"
/ "html" so an operator browsing sent/ knows what the recipient saw.
Renderer enforces a 5 MiB body cap before comrak runs, with a dedicated
ERR BODY_TOO_LARGE wire code. Daemon-side multipart parser uses
RFC 2045 §5.1 quote-aware tokenisation so attachment filenames
containing `"` or `;` round-trip cleanly.

DKIM verification preserved across all three send paths (covered by
end-to-end integration tests). Renderer determinism pinned via comrak
=0.52.0 and checked-in fixture HTML; bump path is `AIMX_BLESS_EXPECTED=1
cargo test`. Version bumped 0.1.0 → 0.2.0; release notes call out the
hard break and migration guidance.
```

PR #208, opened 2026-05-07
[Sprint 2] Wire shape + default send path

uzyn:
## Sprint Goal
Daemon assembles `multipart/alternative` for the default Markdown path, with nested `multipart/mixed` when attachments are present. Signature integrates into the Markdown source so it renders in the HTML part. DKIM verification survives the new structure.

### S2-1: Split `compose_message` — client-side packaging vs. daemon-side wire assembly
- [x] `src/send.rs::compose_message` renamed to `compose_request`. The no-attachment path drops the `Content-Type: text/plain` header so the daemon owns the wire Content-Type.
- [x] New daemon-side `assemble_wire_message()` in `src/wire_assembly.rs`. Calls `render_markdown_to_email_html` and assembles `multipart/alternative` (or nested `multipart/mixed` + `multipart/alternative`) before DKIM signing.
- [x] Multipart text part: `Content-Type: text/plain; charset=utf-8`. ASCII-only bodies use `7bit`; non-ASCII bodies use `quoted-printable`. Same rule applies to the HTML part. Documented in the module header.
- [x] Boundary string format `aimx-<uuid_simple>`; per-message UUID, never hardcoded.
- [x] Unit test: bare Markdown send → wire bytes contain `multipart/alternative` with both parts in order (text first, HTML second per RFC 1341).
- [x] Existing send tests migrated; the `append_signature` tests inside `send_handler` are dropped (signature handling moved to `wire_assembly`); coverage replaced by the new `wire_assembly` test suite.
- [x] **Carry-forward:** the scoped `#[allow(dead_code)]` on `mod markdown_render;` in `src/main.rs` is removed.

### S2-2: Nested `multipart/mixed` + `multipart/alternative` when attachments are present
- [x] `assemble_wire_message` accepts attachments alongside headers + Markdown body and emits the nested `multipart/mixed` shape when attachments are present (extracted from a client-built `multipart/mixed` request body).
- [x] Outer and inner boundaries are independent UUIDs (regression test asserts they are not byte-identical).
- [x] Attachments are base64-encoded with 76-char line wrapping.
- [x] Unit test: 1 attachment → outer mixed, inner alternative, exactly one attachment part with `Content-Disposition: attachment; filename="..."`.
- [x] Unit test: 3 attachments → all three appear as siblings of the inner alternative block.
- [x] Unit test: assembled bytes parse correctly with `mail-parser` (text part present, HTML part present, attachment count matches).
- [x] `cargo test`, `cargo clippy`, `cargo fmt` clean.

### S2-3: Signature appended to Markdown source before rendering
- [x] `assemble_wire_message` reads the per-mailbox `signature` from the resolved `MailboxConfig` (via `Config::effective_signature`) and appends it to the Markdown body — separated by a blank line — before calling `render_markdown_to_email_html`.
- [x] Empty / unset signature appends nothing; the Markdown body is passed through verbatim.
- [x] Unit test: signature `-- [aimx](https://aimx.email)` → rendered HTML part contains `<a href=\"https://aimx.email\">aimx</a>`, text part contains the literal Markdown `-- [aimx](https://aimx.email)`.
- [x] Unit test: empty signature → no extra trailing whitespace beyond the user-supplied body.
- [x] Existing signature-related tests in the send-handler suite continue to pass under the new render path. The previously-multipart `signature_appended_in_multipart_text_part` test is rewritten to assert against the new wire shape (multipart/mixed wrapping multipart/alternative + attachment).
- [x] `cargo test`, `cargo clippy`, `cargo fmt` clean.

### S2-4: DKIM round-trip + bare-Markdown end-to-end test
- [x] Pre-existing `serve::tests::uds_end_to_end_signed_delivery` already runs a default Markdown send through the in-process listener and crypto-verifies the signature with `mail-auth`'s `MessageAuthenticator` against an in-memory TXT cache. It now exercises the new `multipart/alternative` wire shape and continues to pass — DKIM canonicalisation is body-shape agnostic.
- [x] New integration test `send_uds_end_to_end_dkim_verifies_with_attachment`: Markdown + 1 attachment via `aimx send` → captured wire is `multipart/mixed` wrapping `multipart/alternative` with the attachment as a sibling, DKIM body-hash verifies via the same relaxed-canonicalisation helper used by the legacy plain-text test.
- [x] New integration test `send_uds_end_to_end_emits_multipart_alternative_for_markdown_body`: end-to-end via `assert_cmd` + spawned `aimx serve`. `aimx send --body \"# Hello\\n\\nWorld\"` → captured wire is `multipart/alternative`, text part contains `# Hello`, HTML part contains `<h1>Hello</h1>` with the `>World</p>` rendering.
- [x] Sent record exists at `sent/<mailbox>/<stem>.md` with the Markdown source verbatim in the body. No `.html` sibling appears.
- [x] `cargo test`, `cargo clippy`, `cargo fmt` clean.

## Technical Decisions

- **Wire format kept additive.** The `AIMX/1 SEND` codec (`SendRequest { body }`) is unchanged this sprint. The daemon distinguishes \"bare Markdown\" from \"Markdown + attachments\" by inspecting the original `Content-Type` header in the request: no `multipart/...` → entire body is Markdown source; `multipart/mixed` → first part is Markdown, remaining parts are attachments. This avoids paying the codec extension cost twice (Sprint 3 will add `Text-Only` / `Html-Body-Length` headers + an optional second body section for the escape hatches).
- **Daemon-side reassembly, not pass-through.** When attachments are present the client still pre-packages `multipart/mixed`, but the daemon strips the original wrapper, renders the Markdown text part, and emits a fresh `multipart/mixed` wrapping a fresh `multipart/alternative` + attachments. This is what FR-B2 mandates and is also what makes the per-mailbox signature flow into the Markdown render path.
- **Transfer-encoding choice is automatic.** `7bit` for pure-ASCII text/HTML, `quoted-printable` otherwise. Verified by unit tests on both shapes. No `8bit` (mainstream MTAs reject it without `8BITMIME`); no `base64` for text parts (would obscure the Markdown source on text-only clients).
- **Renderer dependency stays daemon-side.** `wire_assembly` imports `markdown_render` (which imports `comrak` + `lol_html`). `src/send.rs` does not import either. The CLI client startup path doesn't initialise comrak's `OnceLock` — the renderer is only touched on the daemon's `handle_send` path.

## Deferred Items
None for this sprint. CLI escape hatches (`--text-only`, `--html-body`) and the `AIMX/1 SEND` codec extension land in Sprint 3 per the plan.

## Review Focus Areas
- `src/wire_assembly.rs` — the multipart parser/assembler. The hand-rolled `split_multipart_body` and `parse_attachment_part` are deliberately small (no full RFC 5322 parser); they only need to handle the well-formed shape `compose_request` produces. Worth a careful read.
- `src/send_handler.rs` — the new `assemble_wire_message` call replaces the old `append_signature` step. Note: `Malformed` is the right error code when assembly fails (the request body is structurally invalid), but the `BodyTooLarge` case (renderer cap exceeded) also surfaces as `Malformed` — Sprint 3 may want a dedicated code.
- The legacy `append_signature` helpers (and their tests) inside `send_handler` are deleted. Coverage migrates to `wire_assembly` tests.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,13 +21,6 @@ mod mailbox;
 mod mailbox_handler;
 mod mailbox_list_handler;
 mod mailbox_locks;
-// `markdown_render` lands the renderer used by the daemon's outbound
-// path. The daemon-side wire-up (`send_handler` calling
-// `render_markdown_to_email_html`) lands in a follow-up commit; until
-// then the public surface is consumed only by the module's own tests.
-// `#[allow(dead_code)]` here is scoped to keep `cargo clippy -D
-// warnings` clean without silencing genuinely-dead items elsewhere.
-#[allow(dead_code)]
 mod markdown_render;
 mod mcp;
 mod mx;
@@ -46,6 +39,7 @@ mod term;
 mod transport;
 mod trust;
 mod uds_authz;
+mod wire_assembly;
 // `release` lands the release-metadata + ReleaseOps used by
 // `aimx upgrade`. Per-item `#[allow(dead_code)]` attributes inside the
 // module scope the lint narrowly so any future genuinely-unused item
--- a/src/mcp.rs
+++ b/src/mcp.rs
@@ -666,7 +666,7 @@ fn build_send_args(params: EmailSendParams, from_address: &str) -> SendArgs {
 /// `From:` out of the composed body itself and resolves the sender
 /// mailbox against its in-memory Config.
 fn submit_via_daemon(args: &SendArgs) -> Result<String, String> {
-    let composed = send::compose_message(args).map_err(|e| e.to_string())?;
+    let composed = send::compose_request(args).map_err(|e| e.to_string())?;
     let request = SendRequest {
         body: composed.message,
     };
--- a/src/send.rs
+++ b/src/send.rs
@@ -84,7 +84,20 @@ fn write_common_headers(
     Ok(())
 }
 
-pub fn compose_message(args: &SendArgs) -> Result<ComposeResult, Box<dyn std::error::Error>> {
+/// Package the CLI arguments into the bytes the client ships over UDS:
+/// the RFC 5322 header block plus the Markdown body (no MIME wrapping for
+/// the default no-attachment path). The daemon's `wire_assembly` module
+/// renders the Markdown source and emits the final `multipart/alternative`
+/// (or nested `multipart/mixed` when attachments are present) before DKIM
+/// signing.
+///
+/// When attachments are present the client still pre-packages them as
+/// `multipart/mixed` with the Markdown source as the first `text/plain`
+/// part — that's the most compact way to ship binary attachment bytes
+/// over the existing UDS frame without extending the codec. The daemon
+/// unpacks that structure, renders the Markdown, and re-emits the final
+/// nested `multipart/mixed` + `multipart/alternative` wire shape.
+pub fn compose_request(args: &SendArgs) -> Result<ComposeResult, Box<dyn std::error::Error>> {
     validate_attachments(&args.attachments)?;
 
     let sanitized_from = sanitize_header_value(&args.from);
@@ -94,9 +107,10 @@ pub fn compose_message(args: &SendArgs) -> Result<ComposeResult, Box<dyn std::er
     let normalized_body = normalize_crlf(&args.body);
 
     if args.attachments.is_empty() {
+        // No attachments: ship headers + Markdown body verbatim. The
+        // daemon decides the wire Content-Type after rendering.
         let mut msg = String::new();
         write_common_headers(&mut msg, args, &date, &message_id)?;
-        msg.push_str("Content-Type: text/plain; charset=utf-8\r\n");
         msg.push_str("\r\n");
         msg.push_str(&normalized_body);
         msg.push_str("\r\n");
@@ -385,7 +399,7 @@ pub fn render_outcome<O: io::Write, E: io::Write>(
 /// `From:` header out of the body itself and resolves the mailbox
 /// against its in-memory Config.
 pub fn build_request(args: &SendArgs) -> Result<SendRequest, String> {
-    let composed = compose_message(args).map_err(|e| e.to_string())?;
+    let composed = compose_request(args).map_err(|e| e.to_string())?;
     Ok(SendRequest {
         body: composed.message,
     })
@@ -468,7 +482,7 @@ mod tests {
     #[test]
     fn compose_has_required_headers() {
         let args = test_args();
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("From: agent@example.com\r\n"));
@@ -482,7 +496,7 @@ mod tests {
     #[test]
     fn compose_has_body() {
         let args = test_args();
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("\r\n\r\nHello, world!\r\n"));
@@ -491,7 +505,7 @@ mod tests {
     #[test]
     fn message_id_format() {
         let args = test_args();
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         let mid_line = text.lines().find(|l| l.starts_with("Message-ID:")).unwrap();
@@ -505,7 +519,7 @@ mod tests {
         let mut args = test_args();
         args.reply_to = Some("<original123@example.com>".to_string());
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("In-Reply-To: <original123@example.com>\r\n"));
@@ -516,7 +530,7 @@ mod tests {
         let mut args = test_args();
         args.reply_to = Some("<original123@example.com>".to_string());
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("References: <original123@example.com>\r\n"));
@@ -528,7 +542,7 @@ mod tests {
         args.reply_to = Some("<reply@example.com>".to_string());
         args.references = Some("<first@example.com> <second@example.com>".to_string());
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("In-Reply-To: <reply@example.com>\r\n"));
@@ -540,7 +554,7 @@ mod tests {
         let mut args = test_args();
         args.reply_to = Some("original123@example.com".to_string());
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("In-Reply-To: <original123@example.com>\r\n"));
@@ -550,7 +564,7 @@ mod tests {
     #[test]
     fn no_reply_to_omits_threading_headers() {
         let args = test_args();
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(!text.contains("In-Reply-To:"));
@@ -597,7 +611,7 @@ mod tests {
             attachments: vec![file_path.to_string_lossy().to_string()],
         };
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("multipart/mixed"));
@@ -626,7 +640,7 @@ mod tests {
             ],
         };
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("filename=\"doc.pdf\""));
@@ -658,7 +672,7 @@ mod tests {
             ],
         };
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("application/pdf"));
@@ -678,7 +692,7 @@ mod tests {
             attachments: vec!["/nonexistent/file.txt".to_string()],
         };
 
-        let result = compose_message(&args);
+        let result = compose_request(&args);
         assert!(result.is_err());
         assert!(result.unwrap_err().to_string().contains("not found"));
     }
@@ -686,7 +700,7 @@ mod tests {
     #[test]
     fn compose_returns_message_id() {
         let args = test_args();
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
         let mid_line = text
             .lines()
@@ -714,7 +728,7 @@ mod tests {
             attachments: vec![file_path.to_string_lossy().to_string()],
         };
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("In-Reply-To: <orig@example.com>"));
@@ -739,7 +753,7 @@ mod tests {
             attachments: vec![file_path.to_string_lossy().to_string()],
         };
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(text.contains("filename=\"file\\\"name.txt\""));
@@ -771,7 +785,7 @@ mod tests {
             attachments: vec![],
         };
 
-        let result = compose_message(&args);
+        let result = compose_request(&args);
         assert!(result.is_err());
         let err = result.unwrap_err().to_string();
         assert!(
@@ -792,7 +806,7 @@ mod tests {
             attachments: vec![],
         };
 
-        let result = compose_message(&args);
+        let result = compose_request(&args);
         assert!(result.is_err());
         let err = result.unwrap_err().to_string();
         assert!(
@@ -813,7 +827,7 @@ mod tests {
             attachments: vec![],
         };
 
-        let result = compose_message(&args);
+        let result = compose_request(&args);
         assert!(result.is_err());
         let err = result.unwrap_err().to_string();
         assert!(
@@ -834,7 +848,7 @@ mod tests {
             attachments: vec![],
         };
 
-        let result = compose_message(&args);
+        let result = compose_request(&args);
         assert!(result.is_err());
         let err = result.unwrap_err().to_string();
         assert!(
@@ -848,7 +862,7 @@ mod tests {
         let mut args = test_args();
         args.reply_to = Some("<orig@example.com>\r\nBcc: victim@evil.com".to_string());
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
         for line in text.split("\r\n") {
             assert!(
@@ -871,7 +885,7 @@ mod tests {
         args.reply_to = Some("<orig@example.com>".to_string());
         args.references = Some("<first@example.com>\r\nBcc: victim@evil.com".to_string());
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
         for line in text.split("\r\n") {
             assert!(
@@ -893,7 +907,7 @@ mod tests {
             attachments: vec![],
         };
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
         assert!(text.contains("From: sender@example.com\r\n"));
         assert!(text.contains("To: recipient@example.com\r\n"));
@@ -948,7 +962,7 @@ mod tests {
             attachments: vec![],
         };
 
-        let result = compose_message(&args).unwrap();
+        let result = compose_request(&args).unwrap();
         let text = String::from_utf8(result.message).unwrap();
 
         assert!(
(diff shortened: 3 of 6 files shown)


uzyn (commented):
## Sprint Goal Assessment

The sprint goal — "daemon assembles `multipart/alternative` by default, nested `multipart/mixed` when attachments are present, signature flows through the Markdown source, and DKIM still verifies" — is fully achieved. The new `src/wire_assembly.rs` module owns the wire shape decision, the daemon's `handle_send` calls it before signing, and the pre-existing `serve::tests::uds_end_to_end_signed_delivery` (which uses `mail-auth` to crypto-verify the signature with an in-memory TXT cache) continues to pass against the new shape, confirming DKIM canonicalisation is unaffected.

**[S2-1] Split `compose_message` — client-side packaging vs. daemon-side wire assembly**
- [x] `compose_message` → `compose_request`, no-attachment path drops `Content-Type` — verified in `src/send.rs:100-121`.
- [x] New daemon-side `assemble_wire_message` in `src/wire_assembly.rs` calls `render_markdown_to_email_html` and emits `multipart/alternative` before signing — verified in `src/send_handler.rs:258-269`.
- [x] Text/HTML parts use `7bit` for ASCII, `quoted-printable` for non-ASCII — `pick_transfer_encoding` (`src/wire_assembly.rs:592-598`), unit tests `seven_bit_encoding_for_ascii_only` and `quoted_printable_encodes_non_ascii_text`.
- [x] Boundary format `aimx-<uuid_simple>`, per-message UUID — `make_boundary()` (line 682-684).
- [x] Unit test asserts default Markdown send produces `multipart/alternative` with text first, HTML second — `default_path_emits_multipart_alternative`.
- [x] Existing send tests migrated; legacy `append_signature` helpers deleted from `send_handler` — confirmed via diff.
- [x] **Carry-forward verified:** the `#[allow(dead_code)]` on `mod markdown_render;` is removed in `src/main.rs:24`.
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check` all pass locally.

**[S2-2] Nested `multipart/mixed` + `multipart/alternative` when attachments present**
- [x] `assemble_wire_message` emits nested mixed-around-alternative when attachments are present — `build_multipart_body` (line 498-537).
- [x] Outer and inner boundaries are independent UUIDs — `outer_and_inner_boundaries_differ` test asserts they differ byte-wise.
- [x] Attachments base64-encoded with 76-char wrap — `base64_encode_wrapped(&att.data, 76)` (line 531).
- [x] Unit test for one attachment — `multipart_request_extracts_text_and_attachments`.
- [x] Unit test for three attachments as siblings — `three_attachments_appear_as_siblings` (asserts 5 outer-boundary occurrences = 1 alt + 3 attachments + 1 close).
- [x] `mail-parser` round-trip test — `mail_parser_roundtrip_with_attachment` confirms text + html + 1 attachment recovered.

**[S2-3] Signature appended to Markdown source before rendering**
- [x] `assemble_wire_message` appends `signature` to Markdown body separated by a blank line before calling the renderer — `append_signature_to_markdown` (line 689-700).
- [x] Empty signature is a no-op — `empty_signature_appends_nothing` test.
- [x] Link-bearing signature renders as clickable HTML and stays Markdown in text — `signature_with_link_renders_clickable_in_html_part` and `signature_is_appended_to_markdown_before_render`.
- [x] Existing signature-related send-handler tests rewritten for the new path and pass — `default_signature_appended_when_config_omits_it`, `custom_signature_overrides_default`, `empty_signature_disables_appending`, `signature_renders_in_text_and_html_parts_with_attachment` all green.

**[S2-4] DKIM round-trip + bare-Markdown end-to-end test**
- [x] Pre-existing `serve::tests::uds_end_to_end_signed_delivery` (mail-auth full-crypto verifier with in-memory TXT cache) passes against the new wire shape.
- [x] New integration test `send_uds_end_to_end_dkim_verifies_with_attachment` asserts nested mixed+alternative + DKIM body-hash verification with 1 attachment.
- [x] New integration test `send_uds_end_to_end_emits_multipart_alternative_for_markdown_body` asserts wire shape, rendered HTML (`<h1>Hello</h1>`), and sent-record body verbatim with no `.html` sibling.
- [x] All three tests pass locally.

## Test Coverage

Coverage is genuinely strong: 14 new `wire_assembly` unit tests, 4 daemon-level signature tests in `send_handler`, 2 new end-to-end integration tests, and the legacy mail-auth crypto-verifier test still passes. Gaps worth tracking (none rise to blocker):

- **`extract_quoted_param` unquoted-value branch is untested.** The `else` arm in `src/wire_assembly.rs:349-357` reads until `;` or whitespace. No test exercises an unquoted boundary or filename. Real production inputs always come from `compose_request`, which always quotes, so this is a non-blocker — but if a future refactor lets unquoted params reach the assembler the regression won't fire.
- **`BodyTooLarge` end-to-end path has no test.** The renderer's `MarkdownRenderError::BodyTooLarge` is converted via `AssembleError::Render(...)` and surfaces as `ERR MALFORMED` with the canonical message in the reason string — but no test feeds a >5MB body through `assemble_wire_message` to confirm the error message reaches the wire intact. Worth adding a 5MB+1-byte unit test in `wire_assembly` so a Sprint 3 dedicated error code change doesn't silently lose the message.

## Potential Bugs

**Non-blocker — `extract_quoted_param` matches `name=` inside `filename=`.** In `src/wire_assembly.rs:341-358`, `extract_quoted_param(s, "name")` finds the substring `name=` inside `filename=` because the lookup is plain `find`, not a delimiter-aware search. Reproduced standalone:
```
input:  application/pdf; filename="doc.pdf"
result: Some("doc.pdf")   // returned by extract_quoted_param(s, "name")
```
The only caller affected is `extract_name_param` in `parse_attachment_part` (line 301), which is the **fallback** when `Content-Disposition: filename=` is absent. In practice the client always emits both `Content-Disposition: attachment; filename="..."` AND `Content-Type: ...; name="..."`, so the bug returns the same value the right code path would have returned by coincidence. It's not exploitable today, but it's a latent foot-gun: if a future refactor calls `extract_quoted_param(ct, "name")` on a header with `filename=` and no `name=`, it will silently return the filename instead of `None`. Worth a delimiter-aware fix (e.g., require the matched position to be at the start or preceded by `;` / whitespace) — but **non-blocker** for Sprint 2 because the observable behaviour is correct on every wire shape `compose_request` produces.

**Non-blocker (acknowledged) — `BodyTooLarge` surfaces as `ERR MALFORMED`.** The implementer flagged this in the PR description. The reason field carries the canonical PRD-mandated error string verbatim, so operators see the actionable guidance. A dedicated code is reasonable Sprint 3 work but not a Sprint 2 blocker.

No other bugs identified. Edge cases I traced through that hold up: empty body, `\n\n` separator (LF-only), display-name `From:`, headers with continuation lines, headers with original `Content-Type` to be stripped, multipart bodies with no parts (falls back to Markdown), boundary occurrences inside attachment payloads (impossible since attachments are base64-encoded by the client), comrak's `OnceLock` initialization on the daemon path.

## Security Issues

- **No new attack surface.** The renderer remains `unsafe_=false` (raw HTML escaped/stripped). The new multipart parser only consumes bytes from a `compose_request`-built UDS frame, which itself enforces CRLF-injection guards on `From`/`To`/`Subject`/`Reply-To`/`References`. The daemon-side `strip_outgoing_content_headers` correctly only matches at line starts (continuation-aware), so a `Subject:` value cannot smuggle a fake `Content-Type:` header.
- **DKIM key handling unchanged.** The key is still loaded once, held in `Arc<RsaPrivateKey>`, never logged, and only consumed by the existing signer.
- **No dependency additions.** No new `Cargo.toml` lines.

## Code Quality

- `src/wire_assembly.rs` is a single 1000-line module; the hand-rolled multipart parser is cleanly factored (`split_headers_body`, `split_multipart_body`, `parse_raw_part`, `parse_attachment_part`). The minimal-parser approach is justified by the fact that input shape is fully under `compose_request`'s control.
- `assemble_wire_message` is the right level of abstraction: the daemon caller just hands it the request body + signature and gets back signing-ready bytes.
- Module header documents the wire-shape decision rule and the transfer-encoding choice — the kind of context a future reader needs.

## Alignment with PRD

Everything in this sprint's PRD scope (FR-A1, A3, A5, B1, B2, B4, B5, C1, F1, F2) is delivered. The two intentionally-deferred PRD items are explicitly Sprint 3 (CLI `--text-only` / `--html-body`, AIMX/1 SEND codec extension) and Sprint 4 (`OutboundFrontmatter::outbound_format`, MCP parameter additions, `book/` updates). Their absence is **not** a scope gap on this PR.

CLAUDE.local.md ban on planning-doc cross-references: **clean.** The PR diff introduces zero new `Sprint N`, `S2-N`, or `FR-XN` tokens outside `docs/`. The pre-existing references in unrelated files (`src/mailbox.rs`, `tests/integration.rs`, `src/mcp.rs`, etc.) are from the user-mailbox track and predate this PR.

- **Overall verdict: Ready to merge.**
- **Blockers:** none.
- **Non-blockers (worth landing in a follow-up, not blocking):**
  - Make `extract_quoted_param` delimiter-aware so `name=` doesn't substring-match inside `filename=` (`src/wire_assembly.rs:341-358`).
  - Add a `wire_assembly` unit test that pumps a >5MB body through `assemble_wire_message` and asserts the canonical `BodyTooLarge` error message reaches the response reason verbatim.
- **Nice-to-haves:**
  - Sprint 3 should add a dedicated `ERR BODY_TOO_LARGE` code so the `MALFORMED` overload goes away (already noted by the implementer).
  - Optional: a unit test covering `extract_quoted_param` against an unquoted boundary value, since the `else` branch is currently uncovered by any direct test.

```
wire_assembly: render Markdown bodies into multipart/alternative on the daemon

Splits the send pipeline so the CLI client (`aimx send`) ships only headers
plus an unprocessed Markdown body over UDS, and the daemon renders + assembles
the wire shape before DKIM signing. Default path emits
`multipart/alternative`; nested `multipart/mixed` wraps it when attachments
are present. Per-mailbox signature is appended to the Markdown source before
rendering so a `[link](url)` signature renders as a clickable `<a>` in HTML
and stays Markdown in the text part.

- New `src/wire_assembly.rs` with `assemble_wire_message`, signature appender,
  multipart parser/assembler, transfer-encoding picker (7bit for ASCII,
  quoted-printable otherwise), independent UUID boundaries.
- `src/send.rs::compose_message` renamed to `compose_request`. No-attachment
  path no longer wraps in MIME — daemon owns the wire Content-Type.
- `src/send_handler.rs` calls `assemble_wire_message` instead of the legacy
  `append_signature`. Legacy signature helpers and their tests are removed;
  coverage migrates to `wire_assembly` tests.
- `mod markdown_render;` no longer carries `#[allow(dead_code)]` (Sprint 1
  carry-forward cleared).
- 14 new `wire_assembly` unit tests, 2 new end-to-end integration tests
  (`send_uds_end_to_end_emits_multipart_alternative_for_markdown_body`,
  `send_uds_end_to_end_dkim_verifies_with_attachment`). Pre-existing
  `serve::tests::uds_end_to_end_signed_delivery` (mail-auth crypto verifier)
  continues to pass against the new wire shape.
```

uzyn:
Addressed both review non-blockers in `8484dab`.

## What was fixed

1. **`extract_quoted_param` substring-match bug** (`src/wire_assembly.rs:347-378`) — rewrote the parser to be delimiter-aware. It now skips the header value's primary value (everything before the first `;`) and walks `;`-separated parameter segments, comparing the trimmed parameter name (case-insensitive) instead of substring-matching the raw header. Asking for `name` against `application/pdf; filename="doc.pdf"` now correctly returns `None` instead of `"doc.pdf"`. The previous behaviour was masked in production because every `compose_request`-built attachment carried both `Content-Disposition: filename=` AND `Content-Type: name=`, so `extract_name_param` returned the right value by coincidence.

2. **`BodyTooLarge` propagation through `assemble_wire_message`** — added a unit test that pumps a `MAX_MARKDOWN_BODY_BYTES + 1` (>5MB) Markdown body through `assemble_wire_message`, asserts the error matches `AssembleError::Render(MarkdownRenderError::BodyTooLarge)`, and asserts the `Display` output carries the canonical `"markdown body exceeds 5MB; use --html-body for pre-rendered large documents or --attachment for sending the document as a file"` string verbatim. This locks the message guarantee at the daemon-side wire-assembly call site so the upcoming Sprint 3 dedicated `ERR BODY_TOO_LARGE` codec extension can swap the wire code without losing the actionable hint.

## Nice-to-haves also addressed

- **Unit test for the unquoted-value branch of `extract_quoted_param`** — covered by `extract_quoted_param_handles_unquoted_values` (asserts `charset=utf-8` and `format=flowed` resolve correctly without surrounding quotes, including the trailing-`;` bound case).
- **Case-insensitive parameter name lookup test** — `extract_quoted_param_is_case_insensitive_on_param_name`.
- **End-to-end sanity test for the original bug shape** — `extract_name_from_content_type_with_only_filename_returns_none` calls `extract_name_param` directly so a future regression at the call site fires.

## Summary of new tests (6 total)

- `extract_quoted_param_does_not_match_name_inside_filename` — the regression itself.
- `extract_quoted_param_finds_explicit_name_alongside_filename` — both params present, each lookup returns its own value.
- `extract_quoted_param_handles_unquoted_values` — unquoted-value branch coverage.
- `extract_quoted_param_is_case_insensitive_on_param_name` — `FileName=` matches a `filename` lookup.
- `extract_name_from_content_type_with_only_filename_returns_none` — call-site sanity.
- `body_too_large_propagates_canonical_message_through_assembler` — the >5MB propagation guard.

## Intentionally left as-is

- **Dedicated `ERR BODY_TOO_LARGE` wire code** — explicitly deferred to Sprint 3 per the review and the original PR description. The new propagation test exists precisely to make that codec swap a no-risk change.
- **Behavioural change for malformed inputs without a primary value** (e.g. a bare `filename="foo"` with no preceding `attachment;`) — the new parser returns `None` for those where the old parser would have matched. This is correct: in MIME, `Content-Type` and `Content-Disposition` values always start with a primary value (type/subtype, or disposition-type), and no caller in this repo passes a primary-less value. Treating malformed input as "no match" is the safer default and keeps the substring-match foot-gun closed.

## Verification

- `cargo test --bin aimx wire_assembly` — 20/20 pass (14 existing + 6 new).
- `cargo test` — 1094 unit + 109 integration + 1 base64 = 1204 pass, 0 fail (pre-existing root-only tests skipped as before).
- `rg` grep for planning-doc cross-references in the diff — clean.

uzyn (commented):
## Re-review: cycle 1 feedback resolved — ready to merge

Verified the fixes in `8484dab`. Both non-blockers from the cycle 1 review are properly addressed; no regressions introduced; all 1094 unit tests + the wire_assembly suite (20/20) pass; clippy and fmt clean.

## Resolved from cycle 1

**1. `extract_quoted_param` substring-match (`src/wire_assembly.rs:341-377`) — FIXED.**
The rewrite is correctly delimiter-aware:
- Skips the header value's primary value via `s.split_once(';')?.1` (returns `None` when no params at all — correct).
- Walks `;`-separated segments, splits each on `=`, and compares the trimmed name with `eq_ignore_ascii_case` against the requested name.
- Handles both quoted (`"..."`) and unquoted (token-until-`;`-or-whitespace) values.

Traced the regression case manually: `extract_quoted_param("application/pdf; filename=\"doc.pdf\"", "name")` now returns `None` instead of `Some("doc.pdf")`. Test `extract_quoted_param_does_not_match_name_inside_filename` locks the behaviour, and `extract_name_from_content_type_with_only_filename_returns_none` provides end-to-end sanity at the call site.

**2. `BodyTooLarge` propagation through `assemble_wire_message` — FIXED.**
The new test `body_too_large_propagates_canonical_message_through_assembler` does exactly what the cycle 1 review asked for:
- Pumps `MAX_MARKDOWN_BODY_BYTES + 1` bytes through `assemble_wire_message`.
- Asserts the variant matches `AssembleError::Render(MarkdownRenderError::BodyTooLarge)`.
- Asserts `err.to_string()` returns the canonical `"markdown body exceeds 5MB; use --html-body for pre-rendered large documents or --attachment for sending the document as a file"` message **verbatim**.

Confirmed via inspection that `AssembleError::Display` for the `Render` arm delegates to `MarkdownRenderError`'s `Display` (`src/wire_assembly.rs:66`), so the message survives the wrap.

## Nice-to-haves also addressed

- `extract_quoted_param_handles_unquoted_values` covers the previously-uncovered `else` branch (both bare `charset=utf-8` and the trailing-`;` case `charset=utf-8; format=flowed`).
- `extract_quoted_param_is_case_insensitive_on_param_name` locks case-insensitivity.

## New observation (non-blocker — for the backlog, not for this PR)

The new parser uses `split(';')` on the after-primary substring. RFC 2045 quoted-string parameter values can in principle contain `;`, e.g. `filename="foo;bar.pdf"`. The previous parser handled that correctly (find `filename=`, then read between the quotes); the new parser splits on the inner `;`, sees an unmatched `"` in the first segment, and falls through to `None`. Worst-case observable impact: the attachment falls back to filename `"attachment"` in `parse_attachment_part`. No crash, no data loss, no body corruption.

This is bounded today because `escape_filename` in `src/send.rs:33-38` doesn't strip `;` — but neither MCP attachments nor the CLI `--attachment` flag prevent a user from passing `foo;bar.pdf`. A delimiter-aware splitter that respects quoted strings (or simply scanning `;` at top-level only) would close this. Not blocking — the cycle 1 substring foot-gun was strictly worse — but worth tracking. The implementer's "Intentionally left as-is" note in their response covers the conservative-on-malformed-input direction; this is the specific case worth a future visit.

## Verification ran

- `cargo test --bin aimx wire_assembly` → 20/20 pass (14 pre-existing + 6 new).
- `cargo test --bin aimx` → 1094 pass, 0 fail, 8 ignored (root-only).
- `cargo clippy --all-targets -- -D warnings` → clean.

## Verdict: Ready to merge

No blockers. No remaining cycle 1 issues. The minor `;`-in-quoted-filename edge case noted above is a non-blocker observation only — does not need to block this PR.

```
wire_assembly: render Markdown bodies into multipart/alternative on the daemon

Splits the send pipeline so the CLI client (`aimx send`) ships only headers
plus an unprocessed Markdown body over UDS, and the daemon renders + assembles
the wire shape before DKIM signing. Default path emits
`multipart/alternative`; nested `multipart/mixed` wraps it when attachments
are present. Per-mailbox signature is appended to the Markdown source before
rendering so a `[link](url)` signature renders as a clickable `<a>` in HTML
and stays Markdown in the text part.

- New `src/wire_assembly.rs` with `assemble_wire_message`, signature appender,
  multipart parser/assembler (delimiter-aware param parser that does not
  substring-match `name=` inside `filename=`), transfer-encoding picker
  (7bit for ASCII, quoted-printable otherwise), independent UUID boundaries.
- `src/send.rs::compose_message` renamed to `compose_request`. No-attachment
  path no longer wraps in MIME — daemon owns the wire Content-Type.
- `src/send_handler.rs` calls `assemble_wire_message` instead of the legacy
  `append_signature`. Legacy signature helpers and their tests are removed;
  coverage migrates to `wire_assembly` tests.
- `mod markdown_render;` no longer carries `#[allow(dead_code)]` (Sprint 1
  carry-forward cleared).
- `BodyTooLarge` propagates through `AssembleError::Render` with the
  canonical actionable message verbatim, so the wire `ERR MALFORMED <reason>`
  reaches operators with the actionable hint intact.
- 20 wire_assembly unit tests, 4 send-handler signature tests, 2 end-to-end
  integration tests (default-Markdown shape + nested mixed/alternative DKIM
  body-hash). Pre-existing `serve::tests::uds_end_to_end_signed_delivery`
  (mail-auth full-crypto verifier) continues to pass against the new shape.
```

PR #210, opened 2026-05-07
[Sprint 4] MCP + sent storage + docs + release notes

uzyn:
## Sprint Goal
MCP `email_send` and `email_reply` mirror the CLI surface (`text_only` / `html_body`); `OutboundFrontmatter` gains `outbound_format`; sent storage stays single-file `.md` for all three paths; book / agent primer / release-notes ship the hard-break advisory; `aimx --version` bumps.

### [S4-1] `email_send` and `email_reply` MCP tools gain `text_only` and `html_body` parameters
- [x] Both tools accept optional `text_only: bool` and `html_body: string` parameters with the same docstring guidance.
- [x] Server-side mutual-exclusion check fires before the UDS is opened. Error wording matches the codec's canonical message (`AIMX/1 SEND: --text-only and --html-body are mutually exclusive`) so operators see one consistent string regardless of which layer tripped.
- [x] `EmailSendParams` / `EmailReplyParams` schema property order pinned by snapshot tests (existing pattern in `mcp::schema_order_tests`).
- [x] Default-Markdown render path preserved when both new params absent — backward compat verified via deserialization tests.
- [x] New unit tests for `validate_text_only_html_body_exclusion`, `build_send_args` forwarding, and the new params' deserialization.
- [x] New integration tests: `mcp_email_send_text_only_emits_single_part_text_plain`, `mcp_email_send_html_body_uses_supplied_html_verbatim`, `mcp_email_send_text_only_plus_html_body_rejected`, `mcp_email_reply_text_only_plus_html_body_rejected`.

### [S4-2] `OutboundFrontmatter::outbound_format` field + sent-storage tests
- [x] `OutboundFrontmatter::outbound_format: String` added in `src/frontmatter.rs`. Default value `"markdown"` on writes from the new code paths.
- [x] `format_outbound_frontmatter` writes the field in the Outbound block, immediately after `outbound`. Field ordering verified by `outbound_format_serializes_immediately_after_outbound`.
- [x] Reader: parsing an outbound frontmatter that lacks `outbound_format` treats it as `"text"` (backward compat, verified by `outbound_format_missing_field_defaults_to_text`).
- [x] `persist_sent_file` accepts the format and stamps it; the daemon derives the value from the SEND request's `text_only` / `html_body` fields.
- [x] Sent-storage integration tests for all three paths now also assert `outbound_format = "..."`.
- [x] Unit-level invariant: no `.html` sibling appears for any of the three paths (`sent_record_never_writes_html_sibling`).

### [S4-3] `book/cli.md` rewrite + new `book/markdown-email.md` chapter
- [x] `book/cli.md` `aimx send` section rewritten to lead with the Markdown default; `--text-only` and `--html-body` documented as escape hatches; `--html-body \"\$(cat ...)\"` shell pattern shown in an example; mutual-exclusion behavior noted; sent-storage pointer added.
- [x] New `book/markdown-email.md` chapter covers: rendering pipeline, supported Markdown features (table), inlined stylesheet's element scope (table), 5MB body cap with canonical error, both escape hatches with worked examples, sent-storage `outbound_format` table, deterministic re-render guarantee, and a worked example.
- [x] `book/SUMMARY.md` registers the new chapter.
- [x] `book/mailboxes.md` "Sending email" section refreshed with Markdown / text-only / html-body examples.
- [x] No planning-doc cross-references in any modified file (verified by the `rg` regex check from `CLAUDE.local.md`).
- [x] `mdbook build website/` succeeds with the new chapter rendered.

### [S4-4] Agent primer + `references/mcp-tools.md` updates + release notes hard-break
- [x] `agents/common/aimx-primer.md` email-tools section rewritten: `email_send` / `email_reply` now lead with Markdown default; new `text_only` / `html_body` parameters documented; "Send an email" workflow shows all three send shapes.
- [x] `agents/common/references/mcp-tools.md` `email_send` and `email_reply` rows updated: new parameter table rows, new docstring guidance, three new examples (default Markdown, plain-text OTP, custom HTML), and the new error wording (mutual exclusion + 5MB cap).
- [x] `book/release-notes.md` carries a hard-break advisory at the very top (`> Hard break (Unreleased): aimx send --body is now interpreted as Markdown by default...`) plus a full release block describing the new behavior, migration story, and rationale.
- [x] `aimx --version` bumped from `0.1.0` to `0.2.0` in `Cargo.toml` so operators on older builds know to re-read the changelog.
- [x] No planning-doc cross-references introduced anywhere outside `docs/`.
- [x] `cargo build` succeeds and the rebuilt agent skill bundles ship the new primer text (`aimx agents setup claude-code --print` shows the new lead-in).

## Technical Decisions

- **MCP submit path now uses `send::build_request` instead of building a `SendRequest` inline.** The previous MCP path hardcoded `text_only=false` / `html_body=None`. Switching to the shared `build_request` helper means MCP and CLI submit byte-identical SEND frames for the same `SendArgs`, removes a hardcoded bug class, and applies the same CRLF normalization to `html_body` automatically.
- **Server-side mutual-exclusion check on the MCP surface, not just the codec.** The codec already rejects both-set frames, but firing the check server-side in the MCP tool body avoids opening the UDS at all and gives the agent a clean tool error rather than a codec error percolating back through the rmcp framing. The wording is intentionally identical so operators see one canonical string regardless of layer.
- **`outbound_format = "text"` as the read-side default for missing field.** Pre-feature sent records were always single-part `text/plain` — defaulting the missing field to `"text"` keeps the audit trail honest about what those legacy records actually shipped. Defaulting to `"markdown"` would have lied about the wire shape.
- **Version bump to `0.2.0` (minor) rather than `0.1.1` (patch).** Markdown is a hard break per the PRD; pre-1.0 semver allows minor bumps to signal breaking changes. The version is the operator's signal to re-read the changelog.
- **Trimmed primer aggressively to stay under the 500-line bundling target.** Initial draft of the "Send an email" workflow was verbose; pulled the `attachments` example and the operator-side `aimx mailboxes create` aside out to keep the primer's size within the existing CI guard. The full content survives in `references/mcp-tools.md`.

## Deferred Items

None — the entire sprint shipped.

## Review Focus Areas

- **`src/mcp.rs` `validate_text_only_html_body_exclusion` + `submit_via_daemon` rewrite.** The new helper is the single point of truth for the MCP-side check; verify the wording matches the codec exactly and the call sites in both `email_send` and `email_reply` invoke it before any IO.
- **`src/frontmatter.rs` `outbound_format` field.** Default-on-read = `"text"`, default-on-write (via the sample fixture) = `"markdown"`. The asymmetry is deliberate — pre-feature records ship `"text"`, new code paths ship a concrete value picked at the call site.
- **`src/send_handler.rs` `outbound_format` derivation.** Three-branch match on `req.text_only` / `req.html_body.is_some()`. The codec rejects both-set frames so the match is exhaustive without a tie-breaker.
- **`book/markdown-email.md`.** The big new docs surface for this PRD. Worth a careful read for accuracy on the rendering pipeline description, the styled-element table (vs. FR-C2), and the sent-storage section's `outbound_format` table.
- **`book/release-notes.md` hard-break advisory.** First line is the operator's first signal; the wording is load-bearing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,7 +19,7 @@ dependencies = [
 
 [[package]]
 name = "aimx"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "assert_cmd",
  "base64",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aimx"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2024"
 rust-version = "1.88.0"
 authors = ["U-Zyn Chua <chua@uzyn.com>"]
--- a/agents/common/aimx-primer.md
+++ b/agents/common/aimx-primer.md
@@ -103,15 +103,23 @@ surface.
   metadata, sorted descending by filename (newest first). `folder` is
   `"inbox"` (default) or `"sent"`. `limit` defaults to 50, max 200; values
   above 200 are silently clamped. Returns a JSON array — agents
-  filter client-side (e.g. `read == false` for triage). aimx never
+  filter client-side (e.g. `read == false` for triage). AIMX never
   scans on your behalf; pass `offset` to page deeper.
 - `email_read(mailbox, id, folder?)`: return the full Markdown file
   (frontmatter + body) for one email.
-- `email_send(from_mailbox, to, subject, body, attachments?)`: compose,
-  DKIM-sign, and deliver an email. `from_mailbox` must be a mailbox you
-  own.
-- `email_reply(mailbox, id, body)`: reply to an existing email. aimx sets
-  `In-Reply-To`, `References`, and `Re:` subject automatically.
+- `email_send(from_mailbox, to, subject, body, attachments?, text_only?, html_body?)`:
+  compose, DKIM-sign, and deliver an email. **`body` is interpreted as
+  Markdown** (CommonMark + GFM extensions). AIMX renders it to HTML and
+  sends as `multipart/alternative` — recipients on rich-capable clients
+  see styled HTML; recipients on text-only clients see the Markdown
+  source. Set `text_only: true` for plain-text-only sends (e.g. OTPs,
+  transactional one-liners). Set `html_body` to override the
+  auto-rendered HTML with your own template. `text_only` and `html_body`
+  are mutually exclusive. `from_mailbox` must be a mailbox you own.
+- `email_reply(mailbox, id, body, text_only?, html_body?)`: reply to an
+  existing email. `body` is Markdown by default with the same `text_only` /
+  `html_body` escape hatches as `email_send`. AIMX sets `In-Reply-To`,
+  `References`, and `Re:` subject automatically.
 - `email_mark_read(mailbox, id)`: mark a single inbox email as read.
 - `email_mark_unread(mailbox, id)`: mark a single inbox email as unread.
 
@@ -247,22 +255,30 @@ tools (or call `email_read` if you prefer the daemon-mediated path), then
 
 ### 2. Send an email
 
+`body` is **Markdown by default** — AIMX renders it to HTML and sends
+as `multipart/alternative`, so recipients on rich clients see styled
+HTML and text-only clients see the Markdown source. Use Markdown for
+headings, tables, links, code blocks, and lists.
+
 ```
 email_send(
   from_mailbox: "agent",
   to: "alice@example.com",
   subject: "Weekly report",
-  body: "Here is the summary..."
+  body: "# Weekly report\n\n- Highlight one\n- Highlight two\n\nDetails: [dashboard](https://example.com/dash)."
 )
 ```
 
+For plain-text sends (OTPs, transactional one-liners) pass
+`text_only: true` to ship `body` verbatim as `text/plain`. For custom
+branded HTML layouts, pass `html_body` (used verbatim as the HTML part)
+and keep `body` as the plain-text fallback. `text_only` and `html_body`
+are mutually exclusive.
+
 The mailbox must exist and you must own it. Provision new mailboxes
-via `mailbox_create(name)` (owned by your uid) or, when the operator
-needs to create one owned by a different user, via
-`sudo aimx mailboxes create <name> --owner <user>` on the host.
-`from_mailbox` is the local part only (e.g. `"agent"`, not
-`"agent@example.com"`). aimx DKIM-signs the message and delivers it via
-direct SMTP to the recipient's MX server.
+via `mailbox_create(name)` (owned by your uid). `from_mailbox` is the
+local part only (e.g. `"agent"`, not `"agent@example.com"`). AIMX
+DKIM-signs the message and delivers it via direct SMTP.
 
 ### 3. Reply to a message
 
--- a/agents/common/references/mcp-tools.md
+++ b/agents/common/references/mcp-tools.md
@@ -251,6 +251,21 @@ Compose, DKIM-sign, and deliver an email. The message is submitted to
 `aimx serve` via UDS, which signs it with the server's DKIM key and delivers
 directly to the recipient's MX server.
 
+`body` is interpreted as **Markdown by default** (CommonMark + GFM
+extensions: tables, strikethrough, autolinks, task lists, footnotes).
+AIMX renders the Markdown to HTML with an inlined stylesheet and ships
+a `multipart/alternative` — recipients on rich-capable clients see
+styled HTML; recipients on text-only clients see the Markdown source.
+
+Two escape hatches cover the edge cases:
+
+- `text_only: true` — ship `body` verbatim as `text/plain`. No
+  rendering. Use for OTPs, transactional one-liners, and existing
+  scripts that must not change shape.
+- `html_body: "<custom html>"` — supply a custom HTML template that
+  AIMX uses verbatim as the `text/html` part. `body` is the
+  `text/plain` fallback. Mutually exclusive with `text_only`.
+
 **Parameters:**
 | Name           | Type     | Required | Description |
 |---------------|----------|----------|-------------|
@@ -258,9 +273,11 @@ directly to the recipient's MX server.
 | `to`           | string   | yes      | Recipient email address |
 | `subject`      | string   | yes      | Email subject |
 | `reply_to`     | string   | no       | Message-ID of the email being replied to. Sets `In-Reply-To`. When `references` is omitted, `References` is built automatically from this value. Required to enable threading. Without `reply_to`, any `references` value is silently ignored and no threading headers are emitted |
-| `body`         | string   | yes      | Email body text |
+| `body`         | string   | yes      | Email body. Interpreted as Markdown by default (CommonMark + GFM). Used verbatim as `text/plain` when `text_only: true`; used as the `text/plain` fallback when `html_body` is set |
 | `attachments`  | string[] | no       | Absolute file paths to attach |
 | `references`   | string   | no       | Full `References` header chain (space-separated Message-IDs). **Only applied when `reply_to` is also set.** Supplied alone, it is silently ignored |
+| `text_only`    | boolean  | no       | When `true`, ship `body` verbatim as `text/plain` with no Markdown rendering and no HTML alternative part. Mutually exclusive with `html_body` |
+| `html_body`    | string   | no       | Custom HTML for the `text/html` part. Operator-supplied; bypasses the renderer. Mutually exclusive with `text_only` |
 
 For simple replies to a single sender, prefer `email_reply`. It reads the
 original email and fills in threading headers and the `Re:` subject
@@ -270,15 +287,37 @@ custom threading chain.
 
 **Returns:** Confirmation with Message-ID.
 
-**Example, plain text:**
+**Example, default Markdown:**
 ```
 email_send(
   from_mailbox: "agent",
   to: "alice@example.com",
   subject: "Status Update",
-  body: "All systems operational."
+  body: "# Status\n\n- All systems operational.\n- See [dashboard](https://example.com/dash) for details."
 )
 → "Sent. Message-ID: <abc123@domain.com>"
+```
+
+**Example, plain-text only (OTP):**
+```
+email_send(
+  from_mailbox: "agent",
+  to: "alice@example.com",
+  subject: "Verification code",
+  body: "Your code: 184293",
+  text_only: true
+)
+```
+
+**Example, custom HTML layout:**
+```
+email_send(
+  from_mailbox: "agent",
+  to: "bob@example.com",
+  subject: "Newsletter",
+  body: "Plain-text fallback for text-only clients.",
+  html_body: "<!DOCTYPE html><html><body><h1>Newsletter</h1>...</body></html>"
+)
 ```
 
 **Example, with attachments:**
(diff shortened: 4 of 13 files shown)


uzyn (commented):
## Sprint Goal Assessment

The sprint goal — mirror the Sprint 3 CLI escape hatches on the MCP surface, land `OutboundFrontmatter::outbound_format`, and ship the docs / release-notes sweep — is fully delivered. All four stories' acceptance criteria check out under independent verification: cargo test, cargo clippy --all-targets -- -D warnings, cargo fmt -- --check, and `mdbook build website/` are clean; the four new MCP integration tests, the three frontmatter unit tests, the per-branch `outbound_format` send-handler tests, and the "no .html sibling" invariant test all pass on a fresh checkout. Brand prose is consistent (capital "AIMX" in the new chapter, primer, and release notes; lowercase `aimx` only as command/identifier). No new banned planning-doc tokens are introduced anywhere outside `docs/` (verified by diffing only the added lines against the regex). The release binary bundles the new primer correctly (`aimx agents setup claude-code --print` shows the new lead-in line).

**[S4-1] `email_send` and `email_reply` MCP tools gain `text_only` and `html_body` parameters**
- [x] Both tools accept `text_only: Option<bool>` and `html_body: Option<String>` with consistent docstring guidance.
- [x] Server-side mutual-exclusion check (`validate_text_only_html_body_exclusion`) fires before any UDS open. Wording matches the codec's canonical message byte-for-byte: `"AIMX/1 SEND: --text-only and --html-body are mutually exclusive"`.
- [x] Schema property order pinned by `schema_order_tests::email_send_params_property_order` and `email_reply_params_property_order`.
- [x] Default Markdown render path preserved when both new params absent — verified by `email_send_params_omitting_new_fields_parses` and `build_send_args_defaults_when_new_params_absent`.
- [x] Eleven new unit tests under `text_only_html_body_tests` cover the validator, the `build_send_args` forwarding paths, and `EmailSendParams` / `EmailReplyParams` deserialization.
- [x] Four new integration tests cover `text_only` end-to-end, `html_body` end-to-end, and the mutex check on both `email_send` and `email_reply`.

**[S4-2] `OutboundFrontmatter::outbound_format` field + sent-storage tests**
- [x] `outbound_format: String` lands in `src/frontmatter.rs` with the `#[serde(default = "default_outbound_format")]` annotation.
- [x] Field ordering — `outbound_format` immediately after `outbound` — verified by `outbound_format_serializes_immediately_after_outbound`, including the strict "nothing between them" check.
- [x] Read-side default of `"text"` for missing field verified by `outbound_format_missing_field_defaults_to_text` against an explicit legacy fixture.
- [x] `persist_sent_file` carries the format through and `handle_send` derives it from the three-branch match on `req.text_only` / `req.html_body.is_some()`.
- [x] Three integration assertions land on `send_uds_end_to_end_emits_multipart_alternative_for_markdown_body` (markdown), `..._text_only_emits_single_part_text_plain` (text), and `..._html_body_uses_supplied_html_verbatim` (html).
- [x] "No `.html` sibling" invariant tightened by `sent_record_never_writes_html_sibling` covering all three branches in one loop.

**[S4-3] `book/cli.md` rewrite + new `book/markdown-email.md` chapter**
- [x] `book/cli.md` `aimx send` section rewritten to lead with the Markdown default. Both escape hatches documented; `--html-body "$(cat ...)"` shell pattern shown; mutual-exclusion behaviour described; sent-storage pointer added.
- [x] New 189-line `book/markdown-email.md` chapter covers rendering pipeline, supported features (table), inlined stylesheet element scope (table), 5MB cap with the canonical error string, both escape hatches with worked examples, sent-storage `outbound_format` table, deterministic re-render guarantee, and a worked end-to-end example.
- [x] `book/SUMMARY.md` registers the new chapter immediately after `Mailboxes & Email`.
- [x] `book/mailboxes.md` "Sending email" examples refreshed for the three send shapes.
- [x] `mdbook build website/` succeeds with the new chapter rendered.

**[S4-4] Agent primer + `references/mcp-tools.md` updates + release notes hard-break**
- [x] `agents/common/aimx-primer.md` email-tools section rewritten; the "Send an email" workflow leads with Markdown and shows all three send shapes. Primer at 490 lines stays under the 500-line bundling target.
- [x] `agents/common/references/mcp-tools.md` `email_send` / `email_reply` rows updated with the new parameter rows, three new example blocks (default Markdown, plain-text OTP, custom HTML), and the new error wording (mutex + 5MB cap).
- [x] `book/release-notes.md` carries the hard-break advisory at the very top, plus a full "Unreleased" release block with What changed / Migration / Why now sections.
- [x] `Cargo.toml` version bumps `0.1.0` → `0.2.0`; `Cargo.lock` matches.
- [x] Release-build agent bundle (`aimx agents setup claude-code --print`) shows the new lead-in text — verified locally.

## Test Coverage

Test coverage is strong and proportionate. Each new code path has both unit-level coverage (pure plumbing, fast) and end-to-end coverage (MCP stdio → daemon UDS → file-drop transport → captured wire bytes). The four new MCP integration tests deliberately mirror the Sprint 3 CLI integration tests so a regression in either layer surfaces independently — the comment on `mcp_email_send_text_only_emits_single_part_text_plain` calls this out explicitly. The `sent_record_never_writes_html_sibling` test sweeps all three branches in one loop, which is the right shape for a structural invariant. The schema property-order tests are extended to pin the new fields' position in the schema (avoids LLM-prompt regressions when the rmcp framing reorders fields).

One minor weakness: the validator unit test `validate_allows_text_only_alone` accepts `text_only=true, html_body=Some("")` as valid, treating empty string as "not supplied". The codec, by contrast, would emit `Html-Body-Length: 0` for an empty `html_body` and reject the frame as the mutex case. See Potential Bugs below; the test pins behaviour that's slightly inconsistent with the codec, but the user-visible result is still the canonical error.

## Potential Bugs

### Non-blocker

1. **Empty `html_body=""` skips the MCP-side validator but trips the codec.**
   `validate_text_only_html_body_exclusion(true, Some(""))` returns `Ok(())` because the validator computes `html_body.map(|s| !s.is_empty()).unwrap_or(false)`. But `build_send_args` forwards `Some("")` unchanged into `SendArgs.html_body`, then `send::build_request` produces `SendRequest { text_only: true, html_body: Some(vec![]) }`, and the codec's encoder (which checks `html_body.is_some()`, not `!html_body.unwrap().is_empty()`) emits both `Text-Only: true` and `Html-Body-Length: 0`. The daemon's parser then rejects the frame because `text_only && html_body_length.is_some()`. The user still sees the canonical error message — but the MCP server's "fast-fail before opening UDS" claim is technically violated for this exact input shape (the UDS is opened, the frame is sent, the daemon parses it and refuses, the error percolates back). Fix: align the validator with the codec's "is-some" semantics (treat `Some("")` as supplied) or have `build_send_args` normalise `Some("")` to `None`. Both are small.

2. **Test comment in `mcp_email_reply_text_only_plus_html_body_rejected` is misleading.**
   The comment says the mutex check fires "before the `validate_email_id` / `lookup_mailbox_row` walk would have returned a different error." In reality, in `email_reply` the order is `validate_email_id` → `lookup_mailbox_row` → `validate_text_only_html_body_exclusion`. The id `"2099-12-31-235959-nope"` happens to pass `validate_email_id` (no `..`, `/`, `\`, `\0`), and `setup_test_env` provisions `alice` so `lookup_mailbox_row` succeeds — so the mutex check does fire and the test passes. But the comment misstates the order. Either reword the comment ("the id need not exist; the parent-message resolution comes after this check") or move the mutex check above `validate_email_id` / `lookup_mailbox_row` to actually match the comment. The `email_send` ordering is similarly post-lookup; if the design intent is "as cheap as possible," moving the mutex check to the very top of both methods is consistent and easy.

## Code Quality

### Non-blocker

3. **`book/markdown-email.md` calls comrak's pin a "minor version" — it's actually an exact-patch pin.**
   The chapter's Determinism section says: *"`comrak` is pinned to an exact minor version in `Cargo.toml`"*. The actual `Cargo.toml` has `comrak = "=0.52.0"` — an exact patch version. The wording rolls into the determinism guarantee, so making it precise matters: a minor-version pin would still allow `0.52.x` patch bumps (which could change rendered whitespace). Suggest: *"`comrak` is pinned to an exact version (`=0.52.0`)"* or *"pinned to an exact patch version"*.

4. **`book/markdown-email.md` `--html-body` example uses `<html>` as a placeholder.**
   *"AIMX uses `<html>` verbatim as the `text/html` part"* — `<html>` is also a real HTML tag, which makes the sentence ambiguous. Suggest a clearer placeholder like `<the supplied HTML>` or just *"AIMX uses your HTML verbatim"*.

## Alignment with PRD

The PR aligns with the PRD §6.5 (FR-E1 through E5) for the MCP surface, §6.6 (FR-F1 through F5) for sent storage, and §6.7 (FR-G1 through G5) for documentation. The deliberate read-side default of `"text"` for the missing `outbound_format` field matches PRD §8 verbatim ("when reading old sent files without the field, treat as `\"text\"`"). The 5MB body cap error string in the docs matches the `markdown_render::MarkdownRenderError::BodyTooLarge` Display impl byte-for-byte. The version bump to `0.2.0` (rather than `0.1.1`) is consistent with the PRD's "hard break" framing — pre-1.0 semver allows minor bumps to signal breaking changes, and the PRD §10 risks table calls this out as a mitigation.

**Overall verdict:** Ready to merge

The sprint closes out the html-email track cleanly. All four stories deliver against acceptance criteria; the new MCP surface mirrors the Sprint 3 CLI semantics down to the canonical error wording; the `outbound_format` audit field has the right read-on-missing-default for backward compat; docs and release notes ship the hard-break advisory at the right places. The two non-blocker findings are minor and don't gate this from landing on `epic/html-email`.

None.

**Non-blockers (track for a future cleanup pass):**

- `validate_text_only_html_body_exclusion` treats `Some("")` as not-supplied while the codec treats it as supplied — produces the right error wording but violates the "fast-fail before UDS" claim for this edge case.
- Test comment on `mcp_email_reply_text_only_plus_html_body_rejected` misstates the check order.
- `book/markdown-email.md` Determinism section calls the comrak pin a "minor version" — it's an exact patch pin.
- `book/markdown-email.md` `--html-body` example uses `<html>` as a placeholder, which collides with the real HTML tag.

**Nice-to-haves:**

- Consider moving the mutex check to the very top of `email_send` and `email_reply` (above `validate_email_id` / `lookup_mailbox_row`) to literally make the "fastest possible failure" claim true, and to keep the two tools symmetrical.

```
mcp + sent storage + docs: Markdown-native outbound is final piece

Mirrors the CLI's `--text-only` / `--html-body` escape hatches on the MCP
`email_send` and `email_reply` tools, lands the `outbound_format` audit
field on `OutboundFrontmatter`, and ships the full docs sweep (book/cli.md
rewrite, new book/markdown-email.md chapter, agent primer + mcp-tools.md
updates, hard-break release notes, version bump to 0.2.0) that explains
the new send semantics to operators and agents.

- email_send / email_reply accept optional text_only: bool and
  html_body: string. Server-side mutual-exclusion check fires before any
  UDS open with the canonical wording the codec uses. Tool docstrings now
  lead with "body is interpreted as Markdown".
- OutboundFrontmatter gains outbound_format immediately after outbound,
  with values "markdown" (default render path), "text" (--text-only),
  "html" (--html-body). Pre-feature sent records default to "text" on
  read for backward compat.
- New book/markdown-email.md chapter documents the rendering pipeline,
  supported Markdown features, the inlined stylesheet's element scope,
  the deterministic re-render guarantee, the 5MB body cap, and the
  outbound_format audit field. Registered in book/SUMMARY.md.
- book/release-notes.md carries a hard-break advisory at the top plus a
  full release block describing the new behavior, migration story, and
  rationale. aimx --version bumps from 0.1.0 to 0.2.0.

20 new unit / integration tests across the new code paths; all 1133 unit
tests + 116 integration tests + 43 verifier tests pass; cargo clippy
--all-targets -- -D warnings clean; cargo fmt -- --check clean; mdbook
build clean.
```

PR #221, opened 2026-05-11
setup drop-through: drop bash -i to fix Step 6 hang

uzyn:
## Summary

- Fixes `aimx setup` hanging at Step 6 with the terminal frozen after `Dropping through to aimx agents setup as <user>...`. Regression introduced four days ago by b9c1fa1, which added `bash -i` to the drop-through. Interactive bash here is several pgrps below the foreground (`sudo -> aimx setup -> runuser -> bash`); its `tcsetpgrp(0, shell_pgrp)` raises SIGTTOU and bash either self-stops or spin-retries.
- Switches the drop-through from `bash -ic` to `bash -lc` (no job control) and explicitly prepends `$HOME/.npm-global/bin` and `$HOME/.local/bin` to PATH inside the shell command body. That replaces the `~/.bashrc`-sourced PATH edits the prior fix was reaching for, without the deadlock.
- Renames the regression test to `drop_through_uses_login_bash_with_path_augmentation` and adds positive assertions on the new shape (`-lc`, `PATH=`, `.npm-global/bin`) plus a negative assertion that `-ic` cannot return.

## Test plan

- [x] `cargo test --bin aimx drop_through` — 2 passed (renamed regression + companion banner-reprint guard).
- [x] Full `cargo test` — 1144 passed locally; the 3 `uds_authz` failures are macOS-specific nsswitch flake (UID 4294967294 resolves to `nobody` on macOS), confirmed to also fail on the unmodified branch. AIMX is Linux-only; CI on Ubuntu is unaffected.
- [x] Live shape verification on Ubuntu 24.04 (Docker): `runuser -l ubuntu -s /bin/bash -- -lc 'PATH=$HOME/.npm-global/bin:... exec ...'` exits 0 and resolves `claude` from the canonical `~/.npm-global/bin/`.
- [ ] End-to-end manual: rebuild on the host that previously hung, run `sudo aimx setup`, walk to Step 6. Expect the agents-setup TUI to render (or `No AI agents detected on this user's filesystem.` if none installed) instead of a frozen terminal.

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -2775,24 +2775,30 @@ pub(crate) fn build_agents_setup_argv(
 ///
 /// 1. If `$SUDO_USER` is set and non-empty → re-exec the resolved aimx
 ///    binary as the invoking user via
-///    `runuser -l $SUDO_USER -s /bin/bash -- -ic "<shell-quoted argv>"`.
+///    `runuser -l $SUDO_USER -s /bin/bash -- -lc "PATH=...:$HOME/.npm-global/bin:... exec <argv>"`.
 ///    Control returns here once the child TUI exits so the closing
 ///    message can print after `aimx agents setup` finishes.
 /// 2. If `$SUDO_USER` is unset → print the guidance message (naming
 ///    `--dangerously-allow-root` as the root-login option) and return
 ///    [`AgentsSetupOutcome::Skipped`].
 ///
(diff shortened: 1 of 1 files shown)


uzyn:
Too kludgy. Considering dropping direct MCP user setup with a message instructing user to run as regular user instead.

PR #222, opened 2026-05-11
Markdown-native outbound: signature on every path + directive MCP guidance

uzyn:
## Summary
- The Markdown-Native Outbound epic moved the signature-append step into the default Markdown branch only, so `--text-only` and `--html-body` silently dropped the configured footer — contradicting `book/configuration.md`'s long-standing promise that "every outbound email gets a signature appended." A real `text_only` send via MCP arrived without the `Sent from AIMX.` trailer.
- `assemble_wire_message` in `src/wire_assembly.rs` now appends the signature ahead of the per-branch wire shaping, so the body carried into all three branches already includes it. `--html-body` still passes the operator-supplied HTML verbatim — the signature only lands in the text/plain fallback on that branch — and the default Markdown path renders the signature into both parts as before.
- Helper renamed `append_signature_to_markdown` → `append_signature_to_body` and re-doc'd; module + `send_handler.rs` comments updated. `book/markdown-email.md` (steps, `--text-only` / `--html-body` callouts, sent-storage table) brought in line with the new universal behavior.

## Test plan
- [x] `cargo test wire_assembly` — 39 passed, including the rewritten escape-hatch cases that now assert the signature IS present and (on `--html-body`) does NOT leak into the HTML part.
- [x] `cargo test send_handler` — 41 passed, including the flipped `text_only_send_emits_single_part_text_plain` / `html_body_send_uses_supplied_html_verbatim` and the new `custom_signature_appended_on_text_only_path` regression.
- [x] `cargo test --test integration` — 116 passed (8 ignored, root-only); flipped the text-only and html-body end-to-end assertions to match.
- [ ] After deploy: send a plain MCP `email_send` and confirm the recipient sees the `Sent from AIMX.` + `https://aimx.email` footer (reproducer for the original report).
- [ ] After deploy: `aimx send --text-only --body "hi" --to <addr>` carries the signature; `aimx send --html-body "<p>x</p>" --body fallback --to <addr>` carries it in the text/plain fallback but leaves the operator's HTML verbatim.

### DIFF
--- a/agents/common/aimx-primer.md
+++ b/agents/common/aimx-primer.md
@@ -123,9 +123,9 @@ surface.
 - `email_read(mailbox, id, folder?)`: return the full Markdown file
   (frontmatter + body) for one email.
 - `email_send(from_mailbox, to, subject, body, attachments?, text_only?, html_body?)`:
-  compose, DKIM-sign, and deliver. `body` is Markdown by default;
-  `text_only` / `html_body` are mutually-exclusive escape hatches (see
-  "Send an email" below). `from_mailbox` must be a mailbox you own.
+  compose, DKIM-sign, deliver. `body` is Markdown → `multipart/alternative` by default
+  — **do NOT pass `text_only: true`** for formatted content (see "Send an email"
+  below). `from_mailbox` must be a mailbox you own.
 - `email_reply(mailbox, id, body, text_only?, html_body?)`: reply with the
   same Markdown semantics; AIMX sets `In-Reply-To`, `References`, and `Re:`
   automatically.
@@ -278,10 +278,10 @@ email_send(
 )
 ```
 
-For OTPs / transactional one-liners pass `text_only: true` to ship
-`body` verbatim as `text/plain`. For branded HTML layouts pass
-`html_body` verbatim as the HTML part with `body` as the plain-text
-fallback. The two flags are mutually exclusive.
+**Do NOT pass `text_only: true`** for prose, briefs, or any body with Markdown
+syntax — it ships raw markers (`#` / `**bold**` / `-`) as `text/plain`. Reserve
+it for unformatted one-liners (OTPs, verification codes). For branded HTML,
+pass `html_body` as the HTML part with `body` as the plain-text fallback (mutually exclusive with `text_only`).
 
 The mailbox must exist and you must own it. Provision via
 `mailbox_create(name)`, or `sudo aimx mailboxes create <name> --owner <user>` for cross-user mailboxes.
--- a/agents/common/references/mcp-tools.md
+++ b/agents/common/references/mcp-tools.md
@@ -261,9 +261,13 @@ styled HTML; recipients on text-only clients see the Markdown source.
 
 Two escape hatches cover the edge cases:
 
-- `text_only: true` — ship `body` verbatim as `text/plain`. No
-  rendering. Use for OTPs, transactional one-liners, and existing
-  scripts that must not change shape.
+- `text_only: true` — **DO NOT use for prose, briefs, summaries,
+  reports, or any body containing Markdown syntax.** That ships the
+  raw Markdown source as `text/plain` and the recipient sees `#` /
+  `**bold**` / `-` markers instead of rendered HTML. Use only for
+  short transactional one-liners with no formatting (OTPs,
+  verification codes, single-line confirmations) and existing scripts
+  that must not change shape.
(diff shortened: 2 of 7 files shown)


uzyn (commented):
## Changes Overview

This PR restores the documented contract that "every outbound email gets a signature appended." The Markdown-Native Outbound epic had narrowed the signature-append site to the default Markdown branch only, causing `--text-only` and `--html-body` (and their MCP equivalents) to silently drop the signature — visible in the wild when an MCP `email_send` with `text_only: true` arrived without the `Sent from AIMX.` trailer. The fix moves `append_signature_to_body()` ahead of the per-branch wire shaping in `assemble_wire_message`, so all three branches ship a body that carries the signature in the `text/plain` region (operator-supplied HTML on `--html-body` stays verbatim). Second commit tightens MCP tool / book guidance to be directive ("DO NOT pass `text_only: true` for prose / formatted bodies") rather than permissive.

## Scope Alignment

Scope fully met. The fix is minimal (one `let body_text = append_signature_to_body(...)` insertion at the right point + a helper rename) and the surrounding tests, integration assertions, and book docs are all flipped consistently. No unrelated refactors.

## Potential Bugs

None. I walked the full data flow:

- `wire_assembly.rs:141` — signature appended once on `body_text` post-extraction, pre-branch.
- `wire_assembly.rs:152-156` — `text_only_top_cte` is picked from `body_text` *after* the append, so the top-level `Content-Transfer-Encoding` in the header block matches the bytes `build_text_only_body` emits.
- `wire_assembly.rs:158-177` — all three branches consume the same `body_text`. Default Markdown renders `body_text` to HTML (signature lands in both parts because it was in the source). `--html-body` uses `body_text` as the text/plain part and the operator's HTML verbatim. `--text-only` ships `body_text` verbatim.
- DKIM signing in `send_handler.rs:291-304` happens *after* `assemble_wire_message`, so the body hash covers the signature. Integration tests `send_uds_end_to_end_text_only_emits_single_part_text_plain` and `send_uds_end_to_end_html_body_uses_supplied_html_verbatim` re-verify this with the real DKIM body-hash routine.
- Persisted sent record in `persist_sent_file` (`send_handler.rs:705`) is the signed wire bytes, so the signature rides into the on-disk `.md` automatically. New test `custom_signature_appended_on_text_only_path` pins both wire and persisted body.
- `append_signature_to_body` with empty `signature` returns the body verbatim, preserving the `signature = ""` operator escape hatch on every path.

## Security Issues

None. No new untrusted-input surfaces; the `--html-body` path still passes operator-supplied HTML verbatim and the renderer is not invoked on that branch (asserted by the `style="..."` absence check at `send_handler.rs:1802`).

## Test Coverage

Coverage is thorough — wire-assembly unit tests, daemon-level `handle_send` tests, and end-to-end integration tests all flipped to assert the new universal behavior, with a new regression `custom_signature_appended_on_text_only_path` covering the exact bug shape. The text-only-with-attachments test additionally asserts the signature lands *before* the first attachment boundary, which is the right invariant.

One minor coverage observation (see Nice-to-haves).

## Code Quality

Helper rename `append_signature_to_markdown` → `append_signature_to_body` is the right move; the new doc string accurately describes the broadened contract. Module-level docs and the `assemble_wire_message` docstring are rewritten and now correctly enumerate the three branches.

- **Blockers**: none
- **Non-blockers**: none
- **Nice-to-haves**:
  - `wire_assembly.rs:1842` `html_body_with_attachment_wraps_alternative_in_multipart_mixed` still passes `""` as the signature, so the html-body + attachments matrix cell is silent about signature placement. Adding a sig-placement assertion (signature in the inner text/plain part, *not* in the supplied HTML, all before the first attachment boundary) would close the test matrix the rest of the PR otherwise covers explicitly.

```
Append outbound signature on every send path

The Markdown-Native Outbound epic had narrowed the signature-append step
to the default Markdown branch, so `--text-only` and `--html-body`
silently dropped the configured footer and contradicted
`book/configuration.md`'s "every outbound email gets a signature
appended" promise. Move the append ahead of the per-branch wire shaping
in `assemble_wire_message` so all three branches carry the signature in
the `text/plain` region (the operator-supplied HTML on `--html-body`
stays verbatim — operators wanting the signature in the HTML view put it
in their template). Tests, integration assertions, and book docs flipped
to match; new regression covers the bug's exact shape. MCP tool / book
guidance now directive against reflex `text_only=true` for formatted
bodies.
```

PR #257, opened 2026-06-16
cargo(deps): upgrade sparkv 0.1 → 0.2.0

uzyn:
## Summary

Upgrade the `sparkv` dependency from `0.1` (locked `0.1.1`) to `0.2.0`. sparkv is the per-mailbox Message-Id dedup cache that stops `on_receive` hooks from re-firing on legitimate SMTP retries. 0.2.0 has three breaking API changes, so a handful of call sites in `src/ingest.rs` adapt. Dedup behavior stays byte-for-byte identical. The verifier crate (`services/verifier/`) does NOT use sparkv and is untouched.

## Requirements

- [x] `Cargo.toml` pins `sparkv = "0.2"` and `Cargo.lock` resolves `sparkv` to `0.2.0` (was `0.1.1`).
- [x] `src/ingest.rs`: dedup existence check `cache.get(&message_id).is_some()` → `cache.contains_key(&message_id)` (0.2.0 `get` returns owned `Option<V>` requiring `V: Clone`; `contains_key` is the allocation-free existence test).
- [x] `src/ingest.rs`: insert value passed by ownership — truncation branch produces an owned `String` (`meta.id.clone()` / `String::new()`) passed by value to `set_with_ttl`.
- [x] `src/ingest.rs` disk-rebuild path: `set_with_ttl(&msg_id, "loaded", …)` → owned `"loaded".to_string()`.
- [x] `src/ingest.rs`: `config.max_item_size = DEDUP_MAX_VALUE_SIZE;` → `config.max_item_size = Some(DEDUP_MAX_VALUE_SIZE);` (field is now `Option<usize>`).
- [x] `src/ingest.rs`: comment referencing `Err(MaxItemsExceeded)` updated to `Err(CapacityExceeded)` (renamed 0.2.0 variant). `ItemSizeExceeded` references unchanged.
- [x] `src/mailbox_locks.rs`: doc comment trimmed from "`SparKV` mutates on every read/write" to reflect that only writes take `&mut self` in 0.2.0; the `Mutex<Option<sparkv::SparKV>>` type annotation needs no change (`SparKV<V = String>` defaults to `String`).

## Out of scope

- The verifier crate (`services/verifier/`) — no sparkv dependency, not touched.
- Any change to dedup semantics, the 24h TTL window, item caps (`DEDUP_MAX_ITEMS` / `DEDUP_MAX_VALUE_SIZE`), or on-disk storage format. Behavior is identical.
- Adopting other new 0.2.0 surface (`get_item`, `delete`, custom `ValueSize` impls) — not needed.

## Technical approach

0.2.0 breaking changes and their mapping:

| 0.2.0 change | Call-site impact |
|---|---|
| `set`/`set_with_ttl` take value by ownership (`V`, default `String`) instead of `&str` | owned-`String` insert values |
| `Config::max_item_size` is now `Option<usize>` (was `usize`) | `Some(...)` wrap |
| `get(&self) -> Option<V> where V: Clone` (0.1 read no longer matches) | switch existence check to `contains_key` |
| Error variant `MaxItemsExceeded` renamed `CapacityExceeded`; `ItemSizeExceeded` unchanged | comment-only update |
| `SparKV<V = String>` now generic with `String` default | no type-annotation change needed |

Verified 0.2.0 signatures: `SparKV<V = String>`, `contains_key(&self, key: &str) -> bool`, `get(&self, key: &str) -> Option<V> where V: Clone`, `set_with_ttl(&mut self, key: &str, value: V, ttl: Duration) -> Result<(), Error> where V: ValueSize`, `Config { max_items: usize, max_item_size: Option<usize>, ... }`. Keys remain `&str`, so existing `&String` call args coerce fine. Error handling only debug-formats `{err:?}` and never matches variants, so the rename is comment-only.

## Test plan

- sparkv is not referenced directly by any test; dedup behavior is exercised transitively through inbound ingest tests in `src/ingest.rs` (`with_fresh_message_id` helper, the `ingest_skips_duplicate_message_id` / `ingest_allows_distinct_message_ids` / `ingest_dedup_cache_loads_from_disk` / `ingest_dedup_cache_excludes_expired` tests) and `tests/integration.rs`.
- Full gate run and green: `cargo build` OK, `cargo clippy -- -D warnings` clean, `cargo fmt --check` clean, `cargo test` all pass (1151 lib/unit + 120 integration + others; 0 failed).
- `grep -rn sparkv src/` confirms every call site migrated — no remaining `&str`-value `set_with_ttl`, no bare-`usize` `max_item_size`.

## Notes

Existing dedup test coverage (same Message-Id ingested twice → second skipped, no second `.md`, hook not re-fired; plus the disk-rebuild and TTL-expiry paths) is comprehensive and serves as the regression guard. Per the plan's guidance to not over-build tests for a mechanical version bump, no new test was added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2742,9 +2742,9 @@ dependencies = [
 
 [[package]]
 name = "sparkv"
-version = "0.1.1"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e3eef4e6fb0a66200df893d1292b96cb92e4f8a43864fd938ba2f96cd345c40"
+checksum = "98a10d99f717743724a8bbf127db24ba4281104dde9534ebcae81c2c8e06219e"
 
 [[package]]
 name = "spin"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ flate2 = { version = "1", default-features = false, features = ["rust_backend"] 
 comrak = { version = "=0.52.0", default-features = false }
 lol_html = "3"
 tempfile = "3"
-sparkv = "0.1"
+sparkv = "0.2"
 
 [dev-dependencies]
 assert_cmd = "2"
--- a/src/ingest.rs
+++ b/src/ingest.rs
@@ -369,7 +369,7 @@ pub fn ingest_email(
             .lock()
             .unwrap_or_else(|p| p.into_inner());
         let cache = cache_guard.get_or_insert_with(|| build_dedup_cache_from_disk(&inbox_dir));
-        if cache.get(&message_id).is_some() {
+        if cache.contains_key(&message_id) {
             tracing::info!(
                 target: "aimx::ingest",
                 "duplicate message_id, skipping rcpt={rcpt} mailbox={mailbox} message_id={message_id}",
@@ -506,13 +506,13 @@ pub fn ingest_email(
             // check is belt-and-braces against the cache's `set_with_ttl`
             // returning `Err(ItemSizeExceeded)` and silently dropping the
             // dedup record.
-            let value: &str = if meta.id.len() <= DEDUP_MAX_VALUE_SIZE {
-                meta.id.as_str()
+            let value: String = if meta.id.len() <= DEDUP_MAX_VALUE_SIZE {
+                meta.id.clone()
             } else {
-                ""
+                String::new()
             };
             if let Err(e) = cache.set_with_ttl(&meta.message_id, value, DEDUP_TTL) {
-                // Failure mode at scale: SparKV returns `Err(MaxItemsExceeded)`
+                // Failure mode at scale: SparKV returns `Err(CapacityExceeded)`
                 // when this mailbox accumulates more than `DEDUP_MAX_ITEMS`
                 // (100k) live entries inside the 24h window — roughly 1.2
                 // sustained inbound messages/second per mailbox, which is
@@ -659,7 +659,11 @@ fn build_dedup_cache_from_disk(inbox_dir: &Path) -> sparkv::SparKV {
         }
         let remaining = total - elapsed;
         let remaining_secs = remaining.num_seconds().max(1) as u64;
-        let _ = kv.set_with_ttl(&msg_id, "loaded", Duration::from_secs(remaining_secs));
+        let _ = kv.set_with_ttl(
+            &msg_id,
+            "loaded".to_string(),
+            Duration::from_secs(remaining_secs),
+        );
     }
     kv
 }
@@ -757,7 +761,7 @@ fn parse_toml_string_value(after_key: &str) -> Option<String> {
 fn make_dedup_sparkv() -> sparkv::SparKV {
     let mut config = sparkv::Config::new();
     config.max_items = DEDUP_MAX_ITEMS;
-    config.max_item_size = DEDUP_MAX_VALUE_SIZE;
+    config.max_item_size = Some(DEDUP_MAX_VALUE_SIZE);
     config.max_ttl = DEDUP_TTL;
     config.default_ttl = DEDUP_TTL;
     config.auto_clear_expired = true;
--- a/src/mailbox_locks.rs
+++ b/src/mailbox_locks.rs
@@ -83,7 +83,7 @@ pub struct MailboxState {
     /// Message-Id dedup cache. `None` until the first ingest into this
     /// mailbox after daemon startup; populated lazily from disk on
     /// first use. Wrapped in `std::sync::Mutex` because `SparKV`
-    /// mutates on every read/write.
+    /// writes take `&mut self`.
     pub seen_message_ids: Mutex<Option<sparkv::SparKV>>,
 }
 


uzyn (commented):
## Standalone Review — PR #257: `cargo(deps): upgrade sparkv 0.1 → 0.2.0`

### Changes Overview

Bumps the `sparkv` dependency from `0.1` (locked `0.1.1`) to `0.2.0` and adapts the four call sites in `src/ingest.rs` to 0.2.0's three breaking API changes, plus a one-line doc-comment correction in `src/mailbox_locks.rs`. sparkv is the per-mailbox Message-Id dedup cache that prevents `on_receive` hooks from re-firing on legitimate SMTP retries. The diff is exactly 4 files (`Cargo.lock`, `Cargo.toml`, `src/ingest.rs`, `src/mailbox_locks.rs`), +15/-11 lines.

### Scope Alignment

Fully met, nothing extra. Every item in the PR's Requirements checklist is present in the committed code, and nothing out of scope leaked in:

- `Cargo.toml` pins `sparkv = "0.2"`; `Cargo.lock` resolves `0.2.0` (checksum `98a10d9…`). The lock diff touches **only** the sparkv block — no other dependency version churned.
- `services/verifier/` is untouched and has no sparkv dependency (confirmed: no `sparkv` in `services/verifier/Cargo.toml` or sources).
- No `.PLAN.md` in the diff. No stray call sites left on the old API — `grep` for sparkv shows exactly the five touched sites (`mailbox_locks.rs:87` type annotation, `ingest.rs` build/config/insert paths), all migrated.

### API Migration Correctness

I extracted sparkv 0.2.0 from the registry cache and verified each call-site change against the actual source — not just the PR's claims:

- **Existence check** `get(&id).is_some()` → `contains_key(&id)`. In 0.2.0 `get(&self, key) -> Option<V> where V: Clone` returns an owned (cloning) value; `contains_key(&self, key) -> bool` is the allocation-free test. Both delegate to the expiry-aware `get_item`, so the dedup decision is semantically identical (still returns `false` for expired-but-unswept entries). Correct.
- **Insert by ownership.** `set_with_ttl(&mut self, key: &str, value: V, ttl)` now takes `V` by value. The truncation branch produces an owned `String` (`meta.id.clone()` / `String::new()`) and the disk-rebuild path passes `"loaded".to_string()`. Correct.
- **`config.max_item_size = Some(DEDUP_MAX_VALUE_SIZE)`.** The field is now `Option<usize>`. Correct.
- **Comment fix `MaxItemsExceeded` → `CapacityExceeded`.** Confirmed: the 0.2.0 error enum is `{ CapacityExceeded, ItemSizeExceeded, TTLTooLong }`; `ensure_capacity()` returns `CapacityExceeded` on `len() >= max_items`. `ItemSizeExceeded` is correctly left unchanged (it is the variant for oversized values). Comment-only, accurate.
- **Type annotation unchanged.** `pub struct SparKV<V = String>` — so `Mutex<Option<sparkv::SparKV>>` resolves to `SparKV<String>`. No change needed; correct.

### Behavior Preservation (dedup)

Byte-for-byte preserved:

- **Value-size bound.** `String::value_size()` returns `self.len()` (byte length), identical to the old `&str` impl. The truncation branch still bounds the inserted value at `DEDUP_MAX_VALUE_SIZE` (256): when `meta.id.len() > 256` it inserts an empty string (size 0, well under cap); otherwise the id verbatim. The `Some(256)` size check enforces the same threshold the old bare-`usize` config did.
- **TTL window** (`DEDUP_TTL` 24h), **item cap** (`DEDUP_MAX_ITEMS` 100k), and **fail-open error handling** (the insert error is only debug-formatted via `{err:?}` and never pattern-matched on a variant) are all untouched — the variant rename is invisible to runtime behavior.

### Test Coverage

No new test was added, which is the right call for a mechanical API-shape migration with no behavioral delta — the value the cache stores and the dedup decision are unchanged. The existing regression guards exercise the migrated paths transitively and all pass: `ingest_skips_duplicate_message_id` (covers `contains_key` + `set_with_ttl` insert), `ingest_dedup_cache_loads_from_disk` (covers the `"loaded".to_string()` rebuild path), and `ingest_dedup_cache_excludes_expired` (covers TTL expiry). 81/81 ingest tests green.

### Verification Performed

- `cargo build` — clean against sparkv 0.2.0
- ingest + dedup test suite — 81 passed, 0 failed

### Summary

- **Overall verdict: Ready to merge.**
- Blockers: none.
- Non-blockers: none.
- Nice-to-haves: none.

This is a clean, correct, fully-scoped dependency bump. Every API mapping was independently verified against the 0.2.0 source, dedup behavior is provably unchanged, and the build/clippy/test gate is green.

```
cargo(deps): upgrade sparkv 0.1 → 0.2.0

Migrate the per-mailbox Message-Id dedup cache in src/ingest.rs to the
0.2.0 API: existence check via contains_key, set_with_ttl values passed
by ownership (owned String), Config::max_item_size wrapped in Some(...),
and the MaxItemsExceeded→CapacityExceeded comment fix. Dedup behavior
(24h TTL, 100k item cap, 256-byte value bound, fail-open insert) is
unchanged. services/verifier/ untouched.
```

